PR #54, opened 2019-01-22
removed the inserter:false prop on the accordion item

colinduwe:


richtabor:
Hey @colinduwe, thanks again for the feedback. We figured out a better way to add new accordion items, so I'm going to close this one up.

screenflow

_Originally posted by @richtabor in https://github.com/thatplugincompany/coblocks/issues/26#issuecomment-456889470_

PR #70, opened 2019-01-23
Template inserter selection fixes, where innerBlocks are not selected when a block is added

richtabor:
Makes sure that when we insert an InnerBlocks block, the parent block gets selected instead of its child getting the selection right away.


richtabor:
This issue was just archived on [Trello](https://trello.com/c/4kh0NAtf) [ _internal_ ] and can be closed. 💪

> 🤖 This is an automated message from our zap.

PR #98, opened 2019-01-24
Fix the icon color within the transforms UI

richtabor:
I think it's better for our custom block icon UI to be removed in places like this, so that they fit into the UI better.

screenshot 2019-01-24 11 37 53


richtabor:
I ended up reverting this as it makes it harder to tell which blocks are ours.

PR #139, opened 2019-01-25
Replace blockcoblocks and blocktemplate text with coblocks

richtabor:


richtabor:
Everything seems to still be working properly

richtabor:
I just wanted to make sure that everything worked before merging this PR. If it does,you can go for it @phpbits. 👍👍

PR #313, opened 2019-02-19
Introduce the new Buttons block

richtabor:
Add the new side-by-side innerBlocks based Buttons block (for use in the upcoming Hero block)


richtabor:
@phpbits This is merging into `master` which I think is right. We'll then pull from master, into the `hero-block` branch, which will bring this official Buttons block into that branch. I'm not sure how you've been keeping the Buttons block in Hero updated. 

Then when Hero is done, I think we have to merge it into the `layout` branch (as it's originally from `layout`). Does that sound right?

PR #371, opened 2019-02-27
Button fullwidth

### LINKED ISSUE #364: Add button styles to the core Button block

phpbits:
Closes #364

### DIFF
--- a/src/extensions/button-controls/styles/style.scss
+++ b/src/extensions/button-controls/styles/style.scss
@@ -1,6 +1,6 @@
 body{
 	.wp-block-button {
-		&.btn-is-full{
+		&.w-100{
 			.wp-block-button__link{
 				width: 100%;
 			}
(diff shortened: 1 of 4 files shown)


src/extensions/button-controls/styles/style.scss
```diff
@@ -1,6 +1,6 @@
 body{
 	.wp-block-button {
-		&.btn-is-full{
+		&.w-100{
```
richtabor: Let's pull this out of the `.wp-block-button` so that we can use it elsewhere where we need to apply 100% width as well.

phpbits: Nice call! Let me check if that will on all instances 👍

phpbits: @richtabor Changed :) Let me know your thoughts. Thanks!

phpbits:
@richtabor Let me know if I can merge this now to layout block branch. Thanks!

richtabor:
Yup, looks good!

Issue #151, opened 2019-01-27
Allow the backgroundImagePanel to have video backgrounds


richtabor:
Im not sure how this would work exactly, but it’d be really neat if we added the ability for blocks that use background images to also support background videos.

richtabor:
@phpbits Do you want to explore how this would work perhaps?

phpbits:
@richtabor We will be needing extra js file for this one ( fitvids, I think ). Or we can use fixed video with 100% width and height of the container. What do you think?

richtabor:
We can try it without JS but we may need to load on the frontend like we do with the Map block.

richtabor:
We could maybe do it like this Coverr does: https://coverr.co

Or perhaps using this method: https://www.w3schools.com/howto/howto_css_fullscreen_video.asp

phpbits:
@richtabor I've created `video-background` branch and applied video background option to hero block. Let me know your thoughts. Thanks!

PR #384, opened 2019-03-15
Add support for video backgrounds in the BackgroundImagePanel component

richtabor:


src/components/background/classes.js
```diff
@@ -20,7 +20,7 @@ function BackgroundClasses( attributes ) {
 		{ 'has-background': attributes.backgroundColor || attributes.customBackgroundColor },
 		{ [ backgroundClass ]: backgroundClass },
 		{ 'has-parallax': attributes.backgroundImg && attributes.hasParallax },
-		{ 'has-background-image': attributes.backgroundImg },
+		{ [ `has-background-${ attributes.backgroundType }` ] : attributes.backgroundImg && attributes.backgroundType },
```
richtabor: @phpbits Will this cause deprecation issues for folks running older versions of CoBlocks, before `backgroundType` existed?

src/blocks/hero/components/edit.js
```diff
@@ -162,6 +163,11 @@ class Edit extends Component {
 					className={ classes }
 				>
 					<div className={ innerClasses } style={ innerStyles } >
+						{ backgroundType == 'video' ? 
+							<div className="coblocks-video-background">
+								<video playsinline="" autoplay="" muted="" loop="" src={ backgroundImg } ></video>
```
richtabor: @phpbits What do you think about making `autoplay`, `muted`, and `loop` ToggleControl options within the Background Settings panel (only if the backgroundType is 'video').

richtabor: On second though, just muted and loop - not autoplay (as it should autoplay).

src/components/background/classes.js
phpbits: @richtabor I've actually changed this to `has-background-${ attributes.backgroundType }` 👍

richtabor:
Closes #151

PR #440, opened 2019-03-26
Modify existing blocks to support video backgrounds

### LINKED ISSUE #422: Existing blocks using the BackgroundPanel need to be updated to receive video
Currently only the Hero block receives video properly. We need to update the following blocks to receive the same functionality (and ensure backwards compatibility with CoBlocks 1.9.3 - ). 

- [x] Row
- [x] Column
- [x] Media Card
- [x] Features
- [x] Hero

Current (Row block):
Screenshot 2019-03-23 14 07 47

@phpbits Is this something we should build a reusable component within BackgroundImagePanel, instead of having the same exact markup/options used in each block?

richtabor:
Closes #422


richtabor:
This only thing left to do to merge this branch is resolve #441

PR #477, opened 2019-04-05
Fix Google Map rendering issue : https://wordpress.org/support/topic/…

phpbits:
Closes : https://wordpress.org/support/topic/map-block-doesnt-work-after-latest-update/


richtabor:
Yup, does the trick! Good find!

Issue #400, opened 2019-03-20
ISBAT use the core Block Manager if its available


richtabor:
We need to figure out a way to check for the core Block Manager and turn ours off if it's present.

I don't want to remove it entirely (yet), as some folks are likely not using Gutenberg 5.3 and may not update to WordPress 5.2 (where the Block Manager be included in core). 

@phpbits Is there also a way to perhaps migrate the selected blocks from our manager to core's?

phpbits:
@richtabor Created deprecation on `remove/block-manager` branch 👍

phpbits:
Forgot to tag, here's the commit : https://github.com/coblocks/coblocks/commit/0e39575bfa514ba07814140c22ad0bf38d4ee5c2

phpbits:
For future reference : https://github.com/WordPress/gutenberg/issues/14139#issuecomment-480261866

phpbits:
@richtabor Added the conditional deprecation so we won't have any issues with core WordPress Gutenberg version vs plugin repo 💯

PR #480, opened 2019-04-05
Deprecate the CoBlocks Block Manager in place of core

richtabor:
This PR deprecates our custom block manager in place of core's manager, which will land in [WordPress 5.2](https://make.wordpress.org/core/2019/02/19/wordpress-5-2-schedule-and-scope/) due April 30th, 2019. #400 

This deprecation will migrate disabled blocks in our block manager to the core manager, so there will be no change in users' currently disabled blocks. 

In the next iteration, @phpbits will be adding a check for the core Block Manager, which will only run the deprecation if the core block manager is not present (so we can push out this before 5.2, and not remove the management feature for folks not on 5.2).


richtabor:
Closes #400

richtabor:
ScreenFlow

PR #484, opened 2019-04-08
Check if posts exists before trying to get the meta

andreilupu:
Hey awesome people!

Just wanted to let you know that in some edge cases like the 404.php templates there is a PHP warning since there is no `global $post` set. Here's a screenshot with the Twentynineteen theme:

ID

A simple "check and quit early if" can avoid that, and there is no side effect since these pages won't need the custom style anyway.

I love the plugin, I hope I can contribute more soon.


richtabor:
Thanks @andreilupu!

richtabor:
This PR will close #523

PR #485, opened 2019-04-08
Adding PHPCS to encourage standards compliance

jrtashjian:
This pull requests adds PHP_CodeSniffer functionality to the project through composer and resolves a number of warnings and errors surfaced by it.

## Running the Code Sniffer
1. Install [Composer](https://getcomposer.org/) if you haven't already.
2. Run `composer install` to install the sniffer and needed standards.
3. Run `composer phpcs` to check the project according to the rules defined in `phpcs.xml`.

This addition utilizes the coding standards defined by the [WordPress Coding Standards](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards) project. We're also using the coding standards defined by the [PHPCompatibility](https://github.com/PHPCompatibility/PHPCompatibilityWP) project to ensure the PHP code is compatible with the stated requirement (current PHP 5.2.4+).

This pull request fixes all issues surfaced while ignoring some specific rules for reasons of preference or where they don't actually apply to the context. You can find these ignored rules by looking for inline comments with `// phpcs:ignore` followed by the rule we're ignoring.

The purpose of this is to help everyone follow the same coding standards as defined in the [guidelines for contributing](https://github.com/coblocks/coblocks/blob/master/CONTRIBUTING.md) document.

PR #506, opened 2019-04-18
Coblocks should validate the activate_plugin nonce before redirecting to the welcome screen

EvanHerman:
In preparation of the PHPCS implementation (#485) this PR checks the security nonce before redirecting the user to the welcome screen. This ensures that Coblocks was the activated plugin and the nonce security check passes.

Where possible we should avoid ignoring PHPCS warnings/errors (ie: `// phpcs:ignore`)

### Other minor tweaks such as:
- Single quotes instead of double
- Spacing tweaks
- Returning early under certain conditions (eg: Nonce security check does not pass, `$activate_multi` is true, the plugin was activated via `WP_CLI` etc.)

### DIFF
--- a/includes/admin/class-coblocks-getting-started-page.php
+++ b/includes/admin/class-coblocks-getting-started-page.php
@@ -173,7 +173,7 @@ class CoBlocks_Getting_Started_Page {
 
 			WP_CLI::log(
 				WP_CLI::colorize(
-					'%b' . sprintf( '🎉 %s %s', __( 'Get started with CoBlocks here:', 'coblocks' ), admin_url( 'admin.php?page=coblocks-getting-started' ) ) . '%n'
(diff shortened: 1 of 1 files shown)


EvanHerman:
@jrtashjian It looks like `@@textdomain` is swapped out for the appropriate text domain in the `Gulp.js` file https://github.com/godaddy/coblocks/blob/master/gulpfile.js#L256

I ran this by @richtabor yesterday before I issued this PR, but I'm sure he can chime in with any additional info I may have missed (as this code base is also fairly new to me :)).

jrtashjian:
@EvanHerman Dove into the `@@textdomain` a bit and I completely missed how coblock's build process is for releases, not just compiling js, etc. `@@textdomain` is correct so I will be reverting back to that in my #485. // @richtabor

PR #509, opened 2019-04-19
Fix overlap of resize handle and color picker

### LINKED ISSUE #501: ISBAT resize the Shape block without UI conflicts
Gutenberg 5.5 tweaked ResizableBox handles, which now conflict with the color selector in the Shape block (at the bottom left of the block). This new line that shows when resizing the divider should not conflict with the color palette selector. 

ScreenFlow-3

EvanHerman:
Resolves #501 

### Example
Example

### DIFF
--- a/src/components/inline-color-picker/styles/editor.scss
+++ b/src/components/inline-color-picker/styles/editor.scss
@@ -1,10 +1,10 @@
 .components-coblocks-inline-color-picker {
-	opacity: 0.99;
-	z-index: 999;
 	bottom: 10px;
 	left: 25px;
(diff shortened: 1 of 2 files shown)


src/components/inline-color-picker/styles/editor.scss
```diff
@@ -1,4 +1,6 @@
 .components-coblocks-inline-color-picker {
+	opacity: 0.99;
```
EvanHerman: Interesting z-index and opacity quirk:
https://philipwalton.com/articles/what-no-one-told-you-about-z-index/

PR #510, opened 2019-04-19
Refactor the Google Map block to make it easier

### LINKED ISSUE #495: ISBAT have a simplified Google Map experience if I don't specify an API key
We could provide a subset of the experience if a person doesn't want to register for Google Maps (iframe the map directly).

jonathanbardo:
Resolve #495


richtabor:
New user flow, when simply adding an address:

ScreenFlow

richtabor:
New user flow, when adding an advanced map (using a Google Maps API key): 

ScreenFlow

src/blocks/map/components/inspector.js
```diff
@@ -197,7 +197,10 @@ class Inspector extends Component {
 							/>
 						</PanelBody>
 					}
-					<PanelBody title={ __( 'Google Maps API Key' ) } initialOpen={ false }>
+					<PanelBody
+						title={ __( 'Google Maps API Key' ) }
+						initialOpen={ address ? false : true }
```
jonathanbardo: @richtabor I think we can simplify here `initialOpen={ !address }`

src/blocks/map/components/controls.js
```diff
@@ -26,8 +26,8 @@ function Controls( { attributes, setAttributes, apiKey } ) {
 
 	const toolbarControls = [
 		{
-			icon: 'edit',
-			title: __( 'Edit' ),
+			icon: icons.mapEdit,
```
jonathanbardo: @richtabor Nice!

src/blocks/map/components/inspector.js
richtabor: Good catch 👌

PR #519, opened 2019-05-11
Introduce a form block

EvanHerman:
Introduce a new form block into Coblocks.

- When first adding the form block, the form is rendered within the block content area. The following fields are defaults and visible: **Name**, **Email Address**, and **Message**.
- The fields can be rearranged or individual form fields can be removed as needed.

### Current State
Demo


src/blocks/form/components/edit.js
```diff
@@ -0,0 +1,403 @@
...
+			height,
+			id,
+			url,
+			width,
+			caption,
+		} = attributes;
+
+		return (
+			<Fragment>
+				<InspectorControls>
```
richtabor: We should pull the InspectorControls into it's own inspector.js file. Do you see foresee an issue with that @EvanHerman?

EvanHerman: @richtabor No that shouldn't be an issue 👍

PR #525, opened 2019-05-13
Media card resize handle fix

sbalakr2:
- Resize handle would go below the card
- Adjusted top positioning


richtabor:
Looks good!

PR #532, opened 2019-05-14
Add grunt + new deploy process

richtabor:


EvanHerman:
When doing a tag commit the `v` needs to be included in the version number for the `deploy` job the kick off.

eg: use `v2.0.0` instead of `2.0.0`

PR #543, opened 2019-05-20
Add missing responsive margin/padding units to Hero and Column blocks

jesstech:
The Column and Hero blocks have non-functional input fields in the Block Inspector when setting "advanced" margin and padding values for a responsive device type.  This is due to missing props on the component.  

This PR adds these missing props to bring feature parity with other block components that utilize the advanced margin/padding controls.


richtabor:
@jrtashjian could you review this? thanks!

PR #544, opened 2019-05-20
Add percentage based background-size to supplement existing options

jesstech:
Currently, block backgrounds support either `auto`, `contain`, or `cover` background sizes.  I have a template which needs to explicitly set a percentage-based value, so I've added an inspector panel control and utilized an inline style when not specifying any of the keywords.

Looking for feedback on this one, as I'm not too familiar with the way WordPress diffs block markup.  I updated `Row`'s 'save' and 'edit' components to match the way other components spread BackgroundStyles as it looked to me like an oversight, but I suspect that doing this may break existing blocks.

### DIFF
--- a/src/components/background/panel.js
+++ b/src/components/background/panel.js
@@ -109,6 +109,7 @@ class BackgroundPanel extends Component {
 			{ value: 'auto', label: __( 'Auto' ) },
 			{ value: 'cover', label: __( 'Cover' ) },
 			{ value: 'contain', label: __( 'Contain' ) },
+			{ value: 'percent', label: __( 'Custom' ) },
 		];
 
 		const overlayStyleOptions = [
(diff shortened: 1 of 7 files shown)


src/components/background/panel.js
```diff
@@ -108,6 +109,7 @@ class BackgroundPanel extends Component {
 			{ value: 'auto', label: __( 'Auto' ) },
 			{ value: 'cover', label: __( 'Cover' ) },
 			{ value: 'contain', label: __( 'Contain' ) },
+			{ value: 'percent', label: __( 'Custom' ) },
 		];
 
 		const overlayStyleOptions = [
```
paranoia1906: This constant is not in use. Should be removed after you merge from master.

AnthonyLedesma:
Hi @jesstech,

Are you still interested in making this change? Thank you for taking time to contribute to CoBlocks.

richtabor:
Hey team, I'm closing this as it's stale, not a high priority, and there's quite a bit of conflicting files currently. If you'd like to jump back in @jesstech you're more than welcome to. :)

PR #545, opened 2019-05-20
Update map block placeholder string

### LINKED ISSUE #534: ISBAT understand that I can search for locations using the Map block
As a customer, ISBAT understand that I can also search for locations in the map block, instead of just adding specific addresses.
 
Acceptance Criteria: 
When you first add the CoBlocks Map block, the placeholder text for the address input field should be: "Search for a place or address".

EvanHerman:
Resolves #534 

Example
Example


richtabor:
Looks good!

PR #547, opened 2019-05-22
Add @wordpress/postcss-themes support

richtabor:
This PR adds theme support for colored UI elements within CoBlocks components/blocks (_as seen below_). Now, when I set a new admin color scheme within WordPress, the relevant UI within CoBlocks properly adjusts. 

To modify the build process we also needed to eject from `create-guten-block`. 

ScreenFlow


jrtashjian:
@richtabor This looks good to me. The only change I made was extracting the PostCSS configuration into it's own file. That way we're only configuring it in a single location for both webpack configs.

PR #556, opened 2019-05-26
Introduce a honeypot spam check field

EvanHerman:
Introduce a hidden honeypot field to help mitigate the number of spam submissions come through.

- Introduce a hidden `coblocks-verify-email` field.
- If `coblocks-verify-email` field contains a value when the form is submitted, `wp_mail()` does not fire.
- Add support for Google reCaptcha v3
- When reCaptcha v3 site/secret keys are added to the form block, the form is protected with reCaptcha on the front of site.
- `coblocks-verify-email` honeypot also persists when reCaptcha is used, as a secondary catch, and a fallback for when reCaptcha is not used.

### Front of Site
Example

___________________________________________________________________________

### To Do:
- [x] Ensure reCaptcha fields work/initialize when multiple forms are rendered on a single page.
- [x] Server side reCaptcha token validation


fjarrett:
@EvanHerman honeypot fields shouldn’t reveal they are honeypots. For example, you could make it `coblocks-verify-email` with the point being to trick bots into actually filling it out.

EvanHerman:
@fjarrett Good point, updating that now.

includes/class-coblocks-form.php
```diff
@@ -351,13 +352,32 @@ public function process_form_submission( $atts ) {
 
 		}
 
+		$spam_check = filter_input( INPUT_POST, 'coblocks-verify-email', FILTER_SANITIZE_STRING );
```
fjarrett: @EvanHerman We should probably comment in the server side code here what is going on, so developers can clearly see this is our honeypot strategy and not a legit field.

richtabor:
Tweaked the UX flow a bit. What do you think @EvanHerman? 

ScreenFlow

EvanHerman:
@richtabor Updates look good to me 👍  :shipit:

PR #559, opened 2019-05-29
Fix transform/style issues with Gutenberg 5.8

richtabor:


richtabor:
Closes #560

richtabor:
> Looks good to me!
> 
> My only concern was calls to `wp.data.select( 'core/editor' )` being changed to `wp.data.select( 'core/block-editor' )` but it looks like they both contain the same properties/functions and won't break anything here.

Yea that was something I was concerned about too. Core uses both, I just implemented whichever one used there in the same context in our stuff.

PR #563, opened 2019-05-29
Add image reordering to gallery blocks

richtabor:
This PR adds the ability to organize images within the Stacked and Masonry blocks using the block content area, similar to how the future core Gallery block will do so. 

ScreenFlow


src/components/block-gallery/gallery-image.js
```diff
@@ -175,14 +181,37 @@ class GalleryImage extends Component {
...
-							className="components-coblocks-remove-gallery-item-button-button"
-							label={ __( 'Remove Image' ) }
-						/>
-					</div>
+					<Fragment>
+						{ supportsMoving &&
+							<div className="components-coblocks-gallery-item__move-menu">
+								<IconButton
+									icon={ verticalMoving ? "arrow-up" : "arrow-left" }
+									onClick={ isFirstItem ? undefined : onMoveBackward }
```
jrtashjian: Instead of using an if-else ternary here we can utilize the `&&` operator and remove the `undefined` return value.

```suggestion
									onClick={ isFirstItem && onMoveBackward }
```

If `isFirstItem` is true and `onMoveBackward` is defined, `onMoveBackward` will be returned here. If `isFirstItem` is false then false will be returned.

src/components/block-gallery/gallery-image.js
```diff
@@ -175,14 +181,37 @@ class GalleryImage extends Component {
...
+									className="coblocks-gallery-item__button"
+									label={ __( 'Move Image Backward' ) }
+									aria-disabled={ isFirstItem }
+									disabled={ ! isSelected }
+								/>
+								<IconButton
+									icon={ verticalMoving ? "arrow-down" : "arrow-right" }
+									onClick={ isLastItem ? undefined : onMoveForward }
```
jrtashjian: Same here: Instead of using an if-else ternary here we can utilize the `&&` operator and remove the `undefined` return value.

```suggestion
									onClick={ isLastItem && onMoveForward }
```

richtabor: I used `onClick={ ! isLastItem && onMoveForward }`

src/components/block-gallery/gallery-image.js
richtabor: I used `onClick={ ! isFirstItem && onMoveBackward }`

jrtashjian: Good catch! :)

PR #564, opened 2019-05-30
Add error notice to Map block

### LINKED ISSUE #551: ISB informed when I don't add a address to the Map block
As a user, I should be informed when I incorrectly fill out the Map block (by not adding an address), applying an empty address to generate a Map. 

**Current:** 

no-missing-address-feedback

**AC:**
1. When an address is empty and I press "Apply", I should be presented with an error message below the form informing me that I need an address.

jrtashjian:
If an attempt to "Apply" the address when none was given, the user is shown an error notice. We prevent multiple error notices from showing each time "Apply" is clicked. All notices are cleared when an address is given.

PR is set to merge into the `issue-495` branch.
Closes #551


fjarrett:
@jrtashjian Why not make the Apply button disabled when the text field is empty?

jrtashjian:
@fjarrett Hmm, that might actually be a better alternative. I can change that pretty quickly.

richtabor:
@jrtashjian I like that alternative!

PR #566, opened 2019-05-30
Add/social circular style

richtabor:
This PR adds a new circular style to the Social block, along with a new "Circle Size" setting for when the circular style is applied. This also touches up the style previews so that each icon is appropriately centered within the core Styles selector. 

**Recording:** 

ScreenFlow


src/blocks/social/index.php
```diff
@@ -93,13 +93,16 @@ function coblocks_render_social_block( $attributes ) {
 	// Attributes.
 	$text_align    = is_array( $attributes ) && isset( $attributes['textAlign'] ) ? "style=text-align:{$attributes['textAlign']}" : false;
 	$border_radius = is_array( $attributes ) && isset( $attributes['borderRadius'] ) ? "border-radius: {$attributes['borderRadius']}px;" : false;
+	$has_padding   = is_array( $attributes ) && isset( $attributes['padding'] ) ? 'has-padding' : false;
```
EvanHerman: @rtabor Technically, when these checks (`is_array( $attributes ) && isset( $attributes['padding'] )`) don't return true, shouldn't these be empty strings instead of boolean values?

src/blocks/social/components/edit.js
```diff
@@ -85,11 +88,12 @@ class Edit extends Component {
 			borderRadius: borderRadius ? borderRadius + 'px' : undefined,
 			backgroundColor: ! hasColors && ! isMaskStyle ? backgroundColor.color : undefined,
 			color: ! hasColors ? this.getTextColor( isMaskStyle ) : undefined,
+			padding: isCircularStyle && padding ? padding + 'px' : undefined,
```
jrtashjian: We can actually drop the ternary operator here and just use the `&&` operator for the same effect. The difference is we're returning `false` instead of `undefined` if `isCircularStyle` is false.

```suggestion
			padding: isCircularStyle && padding + 'px',
```

src/blocks/social/components/edit.js
```diff
@@ -85,11 +88,12 @@ class Edit extends Component {
...
 		};
 
 		const iconStyles = {
-			height: isMaskStyle && iconSize ? iconSize + 'px' : undefined,
-			width: isMaskStyle && iconSize ? iconSize + 'px' : undefined,
+			height: isMaskStyle || isCircularStyle && iconSize ? iconSize + 'px' : undefined,
```
jrtashjian: Same as above: We can actually drop the ternary operator here and just use the `&&` operator for the same effect. The difference is we're returning `false` instead of `undefined` if `isCircularStyle` is false.

```suggestion
			height: (isMaskStyle || isCircularStyle) && iconSize + 'px',
```

src/blocks/social/components/edit.js
```diff
@@ -85,11 +88,12 @@ class Edit extends Component {
...
+			width: isMaskStyle || isCircularStyle && iconSize ? iconSize + 'px' : undefined,
```
jrtashjian: Same as above: We can actually drop the ternary operator here and just use the `&&` operator for the same effect. The difference is we're returning `false` instead of `undefined` if `isCircularStyle` is false.

```suggestion
			width: (isMaskStyle || isCircularStyle) && iconSize + 'px',
```

src/blocks/social/index.php
richtabor: Good catch! 🎣

PR #578, opened 2019-05-30
Rename the "Social block" to "Share block"

richtabor:
In the near future we'll create a block to let folks add their social links/profiles to a page. We need to change the name of our current block to prevent confusion between the two blocks. 

The proposed name for the future block is the "Social Profiles" block, so our two blocks would be the "Share" block and the "Social Profiles" block. 

Thoughts @fjarrett, @EvanHerman, @jrtashjian?


fjarrett:
@richtabor I was going to suggest this very thing several weeks ago. I could not agree more! 💯

richtabor:
Do you think "Profiles" is a better block name, or "Social Profiles"?

fjarrett:
@richtabor I prefer “Social Profiles”

richtabor:
I moved the new Share block within the `src/blocks/share` directory. The block's slug will remain `social` (so that the existing blocks on folks' websites will not be affected).

richtabor:
Just a note, the block's wrapper class `.wp-block-coblocks-social` should stay the same, as we'll reuse most (if not all) of this block's styling for the upcoming **Social Profiles** block. This way we don't have to repeat ourselves, as these blocks will be similarly styled.

PR #579, opened 2019-05-31
WIP: Reordering social icons within block

### LINKED ISSUE #466: ISBAT re-order social icons in the Share block
The [Contact Widgets plugin](https://wordpress.org/plugins/contact-widgets/) has an interesting UI where users can drag to organize the activated icons how they see fit. It'd be interesting to apply that same technique to our Social block.

ScreenFlow

I think it'd be interesting if we could make the icons the draggable components.

jrtashjian:
Work in progress. A bit of cleanup is necessary as each shareable place is hardcoded as a boolean variable.

Closes #466


jrtashjian:
Closing this for now. We can come back to it another time.

PR #580, opened 2019-05-31
Add optional links to individual gallery images

richtabor:
This PR enables users to add individual links to gallery images within our gallery blocks (a common support request from [WordPress.org](https://wordpress.org/support/plugin/block-gallery/)). 

There's still a few loose ends: 
- [x] Add a "Custom" value to the blocks' Link Settings PanelBody, which then reveals the imageLink UI
- [x] The new link UI within the Stacked block is not working 100%
- [x] The URLInput empty upon transforms, even if the current image's custom link is applied
- [x] Carousel does not currently support links, do we remove the UI on Carousel - or work in support?

**Preview:** 

ScreenFlow


richtabor:
The Custom Link UI only displays if "Custom URL" is selected via Link Settings: 
ScreenFlow

richtabor:
This only works with the Stacked and Masonry blocks, as the carousel does not support links at all (currently).

src/components/block-gallery/gallery-link-settings.js
```diff
@@ -0,0 +1,93 @@
...
+						{ linkTo !== 'none' &&
+							<Fragment>
+								<ToggleControl
+									label={ __( 'Open in New Tab' ) }
+									onChange={ this.setNewTab }
+									checked={ target === '_blank' }
+								/>
+								<TextControl
+									label={ __( 'Link Rel' ) }
+									value={ rel || '' }
```
jrtashjian: Since `rel` is an attribute, you could probably just set a default for it when you define the attributes in `attributes.js`. It's not absolutely necessary and you can decide if you want to do this. Just thought it was worth mentioning.

```
rel: {
    type: 'string',
    default: '',
}
```

jrtashjian: With a default this conditional wouldn't be necessary.

richtabor:
Looks ready to go!

richtabor:
@jrtashjian Should we wait to merge this before the Jest tests?

jrtashjian:
@richtabor we will need to implement integration tests to test this out, which are going to come at a later time. Let's go ahead and merge this one.

benjaminluoma:
Is there a plan to add this to carousel blocks as well?

richtabor:
> Is there a plan to add this to carousel blocks as well?

Not immediately, though that's something I do want to look into. If I recall correctly, the big issue was figuring out how to properly navigate the carousel within the editor to add links - as currently, if you let off of the carousel (for example, to click into a URL field for an image), it resets the carousel instead of keeping it in place. If you have any ideas, I'm all ears :)

PR #582, opened 2019-06-07
Update gallery block icons

richtabor:
This PR updates the gallery block icons (Masonry, Carousel, Stacked) to be more representative of actual image galleries, pulling from the core image/gallery block iconography style. 

**Before:** 
before

**After:**
after


richtabor:
@EvanHerman Do you mind looking at the CircleCI tests that are failing?

Issue #581, opened 2019-06-07
ISBAT understand what "Stacked" and "Masonry" blocks are


richtabor:
Current it is not very clear that the "Stacked" and "Masonry" blocks are in fact gallery blocks. I should be able to clearly understand that they are types of galleries. 

I have two suggestions: 

**1.** Change the block names within the Block Inserter to add "Gallery" (Screenshot A). 

**2.** Move our gallery blocks (Masonry, Carousel and Stacked) into a new "Galleries" category within the block inserter. (Screenshot B). When a user searches for a gallery, the gallery blocks will be clearly labeled in the Gallery category.

**Screenshot A:**
<img width="619" alt="Screen Shot 2019-06-07 at 11 51 13 AM" src="https://user-images.githubusercontent.com/1813435/59127024-838ef680-8934-11e9-8a17-3b20493362d6.png">

**Screenshot B:**
![ScreenFlow](https://user-images.githubusercontent.com/1813435/59125650-f5654100-8930-11e9-8ee1-72bbec12fd74.gif)

richtabor:
The labels in Option 1 (changing block names) are quite long and can be skewed in other areas of the editor UI: 
<img width="712" alt="Screen Shot 2019-06-07 at 3 01 07 PM" src="https://user-images.githubusercontent.com/1813435/59127247-1891ef80-8935-11e9-8b55-14adefb1b7f7.png">

gmays:
Both could work, I don't have a strong opinion either way.  Some thoughts/questions:

Is there a reason that only CoBlocks has the CoBlocks logo and not Gallery Blocks?  Could customers accidentally assume the Gallery Blocks are part of core and uninstall CoBlocks after assuming they're not using any other blocks?

Do we think this is scalable?  There will likely be other reasons to clarify names for future blocks.  Does this set a precedence for creating a new group every time?

How big a deal is it that the Block names are long?  I assume at some point in the long future of Gutenberg this will become the norm as we get more nuanced blocks and this will be solved a more elegant way (maybe we solve it).

richtabor:
> Is there a reason that only CoBlocks has the CoBlocks logo and not Gallery Blocks? Could customers accidentally assume the Gallery Blocks are part of core and uninstall CoBlocks after assuming they're not using any other blocks?

We could have the CoBlocks logo on both categories. I think that's fine. 

> Do we think this is scalable? There will likely be other reasons to clarify names for future blocks. Does this set a precedence for creating a new group every time?

[...]

PR #584, opened 2019-06-10
Add new category for gallery blocks

richtabor:
This is an experimental PR for #581 that adds a new "Gallery Blocks" category to the block inserter. 

**Screenshot:**
ScreenFlow


fjarrett:
@richtabor Personally, I prefer this category approach over #592

EvanHerman:
@richtabor @fjarrett I'm also a fan of this approach over the other. I wonder if we should re-name the category to something like `CoBlocks Gallery Blocks` so the two categories `CoBlocks` and `CoBlocks Gallery Blocks` live next to each other.

EvanHerman:
Also, just another thought...isn't `CoBlocks Gallery Blocks` a little redundant since everything in that block inserter window are blocks.

richtabor:
> Also, just another thought...isn't `Gallery Blocks` a little redundant since everything in that block inserter window are blocks.

I agree, the pattern is sort of disjointed in core. Perhaps ‘CoBlocks Galleries’ would work better?

richtabor:
Here's another pass, with the CoBlocks categories displaying right after each other. I like this much better. Thoughts @fjarrett, @jrtashjian, @EvanHerman, @gmays? 

ScreenFlow

fjarrett:
:100:

PR #585, opened 2019-06-10
Experiment: Use proper plugin basename

richtabor:
This PR moves `class-coblocks.php` to `includes/class-coblocks.php` to utilize a proper plugin basename.

### DIFF
--- a/.dev/tests/phpunit/bootstrap.php
+++ b/.dev/tests/phpunit/bootstrap.php
@@ -9,7 +9,7 @@ require_once $_tests_dir . '/includes/functions.php';
 
 function _manually_load_plugin() {
 
-	require dirname( dirname( dirname( dirname( __FILE__ ) ) ) ) . '/class-coblocks.php';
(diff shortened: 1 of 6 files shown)


richtabor:
Yea I don't like that it deactivates too.

fjarrett:
@jrtashjian we should make it back compat

jrtashjian:
This might take a little looking into unless you have seen this accomplished before.

My first thought is to add a new `class-coblocks.php` file to the plugin root which just calls the new `coblocks.php` file but with the same plugin header info. This doesn't do what we'd like because WordPress lists the plugin twice in the plugins list. Having the header only in `coblocks.php` will cause the same deactivation issue.

Maybe there is a way we can have both files with the plugin header in each. After updating we would need to deactivate the plugin with the old basename and activate the one with the new basename. Finally we need to hide the plugin with the old basename from the plugins table.

That is the best solution I can think of right now. The `class-coblocks.php` file will have to live in the plugin dir until we can determine users are on the version this is merged into.

PR #590, opened 2019-06-13
Add Menu block

jrtashjian:
## Description

New block to add a menu to an eatery’s website so that potential customers know what kind of food/beverages are served.

## Features

- Starter content that helps the user know how to proceed with editing the block.
- Display the menu item's name.
- Display the menu item's description.
- Display the menu item's cost.
- Optionally display the menu item's image, toggled as a group or individually.
- Three styles available, which so far I've named: "Grid", "Justified Grid", and "List".
- The ability to click a button which inserts another menu block with placeholder data. The style of the new block should replicate styles selected from the previous block on creation.
- Automatically insert a new placeholder menu item block to make editing easier.

## Screenshots

**Style One (Grid)**

is-style-one with images

**Style Two (Justified Grid)**

is-style-two with images

**Style Three (List)**

is-style-three with images


src/blocks/menu-item/edit.js
```diff
@@ -0,0 +1,192 @@
...
+	// Add a placeholder if there are none. Remove trailing placholders if there are more than one.
+	if ( placeholders.length === 0 ) {
+		const newMenuItem = wp.blocks.createBlock( 'coblocks/menu-item', {} );
+		dispatch( 'core/editor' ).insertBlocks(
+			newMenuItem,
+			menuItems.length,
+			menuClientId,
+			false
+		);
+	} else if ( placeholders.length > 1 ) {
```
jonathanbardo: @jrtashjian Normally when I see an else statement I think there might be a better way. What about a method that tells you explicitely what you are about to do with only 1 if that returns early. Maybe something like `dispatchCoreInsertBlock` or something.

src/blocks/menu-item/edit.js
```diff
@@ -0,0 +1,193 @@
...
+		);
+	}
+
+	renderImage() {
+		const { attributes } = this.props;
+		return (
+			<Fragment>
+				{ this.renderToolbarEditButton() }
+				<figure>
+					<img src={ attributes.itemImage } alt={ '' } />
```
jonathanbardo: @jrtashjian Could we fill the alt from the image caption here?

jrtashjian:
Some touchups:

- [x] Put the /menu-item/ within the src/blocks/menu directory. [ea48fdd]
- [x] Separate svg icon(s) into a icons.js file (for each block). [1897b17]
- [x] Use SVG, Path and G components for our svg icons. [1897b17]
- [ ] Let's break out the Inspector controls into a separate inspector.js file (to follow the same conventions as our other blocks.
- [x] Do we need the "item" preface for the "itemImage" attributes and such? If not, let’s make our attributes simpler. [8336225]
- [x] Add a “Price” ToggleControl to the Menu Item block (same as the Image). [af380fd]
- [ ] List view menu items should have less margin bottom than grid view.

jrtashjian:
Note to self: we need to make sure we're reading attributes from the markup, not the block comment.

richtabor:
Just a few more tweaks left: 

- [ ] List style placeholder icon is not centered (_Screenshot 1_)
- [ ] Price needs adjusted for the List style (_Screenshot 2_)
- [ ] Add UI to remove a selected image from a Menu Item block (_Screenshot 2_)

**Screenshot 1:**
menu-placeholders

**Screenshot 2:**
Screen Shot 2019-06-16 at 8 40 50 PM

jrtashjian:
> * Explore adding double-click functionality to pull up the Media Library when double-clicking on an image (similar in functionality to how the core Image block provides editing when double-clicking on the image).

After diving into this a bit I would suggest we wait until this functionality is merged into core (right now it's included in the Gutenberg plugin). We will have to think about how this interaction would work in the list view with the smaller image. There is a lot of info in the MediaPlaceholder after double-clicking which works well for large images but not smaller ones.

image

I'm not seeing an easy way to just load up the media library when we double click without also re-implementing a lot of stuff that the MediaPlaceholder component already handles. I think that the "Edit Image" toolbar button (shown when an image has been uploaded), the ability to drag and drop a new image to replace the current one, and (after implemented) the ability to remove the image by clicking the "x" icon will work well for now.

src/blocks/menu-item/edit.js
jrtashjian: Good point. I came up with a new solution that avoids the else statement. Decided to just always remove all placeholder blocks categorized as trailing when this runs. We now only check the length to add a placeholder block.

```
dispatch( 'core/editor' ).removeBlocks(
    placeholders
        .filter( ( item, index ) => item.clientId !== childClientId && index !== 0 )
        .map( item => item.clientId ),
    false
);
```

src/blocks/menu-item/edit.js
jrtashjian: Added.

jrtashjian:
@richtabor I think we're ready to start merging this in! All tweaks above should be resolved/addressed. My only question is if we're wanting to keep the justified style out for now or add it. I think we can go either way but I also don't think it's a hugely different style to require it's inclusion.

I'm going to remove the "Do Not Merge" tag from this PR. I'm satisfied where this is at where it could be merged in at any time. Feel free to merge it after your final review.

richtabor:
Good stuff @jrtashjian! There are just a few extra bits that we should do that I noticed while testing:

- [ ] Add `align-items: center` to `.wp-block-coblocks-menu.is-style-list .wp-block-coblocks-menu-item` to vertically center the text with the image on the left for the List style
- [ ] Set the image height/width to `100px` instead of `130px`
- [ ] Remove the margin bottom applied to the description text in List style
- [ ] The padding on the heading block within the Menu block is acting up (_screenshot below_). 

Screen Shot 2019-06-20 at 6 24 06 PM

jrtashjian:
Commit b2a59bf will resolves #610.

EvanHerman:
@richtabor @jrtashjian Looks great - and works really well on my end! Only thing I think would take this to the next level is ability to add custom item attributes/icons, or filter existing attribute icons with custom ones. This would be a nice enhancement for a future release.

This PR is good to go! :shipit:

PR #592, opened 2019-06-13
Add "Gallery" to gallery block names

richtabor:
This PR is the alternate experiment of the #581 issue, appending "Gallery" to the gallery block names, to make them much more recognizable - instead of having a separate Gallery Blocks category within the block inserter (which is this #584 PR).

Screen Shot 2019-06-13 at 3 19 16 PM


richtabor:
Closing this experiment as we're likely going with #584.

PR #598, opened 2019-06-17
Fix Row Block selected state

richtabor:
This PR fixes an issue where the selected state of the Row block was misaligned due to recent updates in Gutenberg. 

**Before:** 
before

**After:**
Screen Shot 2019-06-16 at 8 31 38 PM


src/blocks/row/styles/editor.scss
```diff
@@ -83,11 +84,12 @@
...
-				margin-top: -12px;
+				margin-top: 0;
+				// margin-top: -12px;
 
-				&::before {
-					top: -1px;
-				}
+				// &::before {
+				// 	top: -1px;
+				// }
```
jrtashjian: Let's remove this code instead of leaving it in commented.

src/blocks/row/styles/editor.scss
```diff
@@ -83,11 +84,12 @@
 
 			> .editor-block-list__block-edit {
 				margin-bottom: 0;
-				margin-top: -12px;
+				margin-top: 0;
+				// margin-top: -12px;
```
jrtashjian: Let's remove this code instead of leaving it in commented.

src/blocks/row/styles/editor.scss
```diff
@@ -54,8 +54,9 @@
 
 		// Adjust the individual column block.
 		> [data-type="coblocks/column"] {
-			display: flex;
-			flex-direction: column;
+			margin-left: 0;
+			// display: flex;
+			// flex-direction: column;
```
jrtashjian: Let's remove this code instead of leaving it in commented.

richtabor:
@EvanHerman Mind taking over these requested changes?

PR #602, opened 2019-06-24
WIP: Edit Gist component to handle when invalid URL is provided

paranoia1906:
Pull to handle https://github.com/godaddy/coblocks/issues/601.


paranoia1906:
I left some extra spacing and line breaks in the original commit. The subsequent commits were to address the extra lines and formatting.

src/blocks/gist/components/gist.js
```diff
@@ -97,11 +106,13 @@ export default class Gist extends Component {
 				gistContent: gist.div
 			});
 		}
-
 		let gistScript = document.createElement('script');
 		gistScript.type = 'text/javascript';
 		gistScript.src = this._tranformedURL(gistCallback);
-		document.head.appendChild(gistScript);
+		gistScript.onerror = function (err) {
+		  _handleScriptError();
```
jrtashjian: I think this needs to be `this._handleScriptError()` otherwise the method is not being called.

jrtashjian:
It looks like the `master` branch has an update which resolves a PHP syntax error causing WordPress to throw an error when the admin is loaded. Pull in the latest commits from master to resolve it.

src/blocks/gist/components/gist.js
jrtashjian: Disregard. Because we're inside a callback function `this` will reference the script element rather than the Gist class. I see that `const { _handleScriptError } = this;` is the workaround.

richtabor:
@paranoia1906 Could you post a screen-recording of what's changed specifically?

paranoia1906:
Gist Block

Scenario: I enter into the Gist Block a URL that does not resolve to a valid Gist.

Before: Spinning animation persists.

GistOriginal

After: Error loading gist error message is displayed.

GistModified

paranoia1906:
This forked branch had grown unruly and must be stopped. Closing pull request to resubmit on a new branch.

PR #604, opened 2019-06-25
Introduce filter to nest Getting Started page in a parent menu item

### LINKED ISSUE #603: Add option to move CoBlocks dashboard menu item to "Settings" or "Tools"
**Is your feature request related to a problem? Please describe.**
As of now, CoBlocks adds a top-level menu item in the WP dashboard. Apart from some basic onboarding info there is no real value there that would justify such an exposed position. As I understood from [this issue](https://github.com/godaddy/coblocks/issues/265), there might be more settings coming in the future. In any case and in order to unclutter the WP dash, I would love to see an option to move the CoBlocks menu item under the "Settings" or the  "Tools" menus.

UI: Add a checkbox to move the CoBlocks Settings Page.

You could consider not adding a top-level menu item in the first place and rather adding a submenu item under "Settings" or "Tools" right away.

EvanHerman:
Resolves #603 

Introduce a new filter, `coblocks_getting_started_submenu_parent_slug `, which allows the CoBlocks Getting Started page to be nested in a parent menu item, if the user chooses.

#### Example Usage
```php
/**
 * Nest the CoBlocks Getting Started page inside the Tools admin menu
 *
 *
 */
function nest_coblocks_getting_started_menu_item( $parent_slug ) {

   return 'tools.php';

}
add_filter( 'coblocks_getting_started_submenu_parent_slug', 'nest_coblocks_getting_started_menu_item' );
```

### DIFF
--- a/includes/admin/class-coblocks-getting-started-page.php
+++ b/includes/admin/class-coblocks-getting-started-page.php
@@ -37,7 +37,7 @@ class CoBlocks_Getting_Started_Page {
 		 *
 		 * @var string
 		 */
-		$submenu_parent_slug = (string) apply_filters( 'coblocks_getting_submenu_slug', '' );
+		$submenu_parent_slug = (string) apply_filters( 'coblocks_getting_submenu_parent_slug', '' );
 
 		if ( '' !== $submenu_parent_slug ) {
 
--- a/includes/admin/class-coblocks-getting-started-page.php
+++ b/includes/admin/class-coblocks-getting-started-page.php
@@ -37,7 +37,7 @@ class CoBlocks_Getting_Started_Page {
 		 *
 		 * @var string
 		 */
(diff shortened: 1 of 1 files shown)


includes/admin/class-coblocks-getting-started-page.php
```diff
@@ -32,6 +32,28 @@ public function screen_page() {
...
 
+		/**
+		 * Allow users to nest the CoBlocks menu page
+		 *
+		 * @var string
+		 */
+		$submenu_parent_slug = (string) apply_filters( 'coblocks_getting_submenu_parent_slug', '' );
```
richtabor: I think this would be better suited as `coblocks_getting_started_submenu_parent_slug`.

EvanHerman: @richtabor Whoops - good catch! 😅

PR #605, opened 2019-06-26
Double clicking the rendered map brings up the location editor

### LINKED ISSUE #594: ISBAT double-click a rendered map within the Map block to edit a location
Following a design pattern set in the core Image Block, I should be able to double-click on a rendered map within the editor to change the saved address/location that the map is currently displaying.  

While in the editor, double-clicking on a rendered map should bring up the edit location view for me to type in a new location/address to display.

EvanHerman:
Resolves #594 

Double clicking on the map inside of the map block displays the location editor

### DIFF
--- a/src/blocks/map/components/edit.js
+++ b/src/blocks/map/components/edit.js
@@ -257,7 +257,9 @@ class Edit extends Component {
 						} }
 					>
 						{ !! this.state.apiKey ?
-							<GoogleMapApiRender address={ address } /> :
+							<div style={ { width: '100%', height, position: 'absolute' } } onDoubleClick={ () => setAttributes( { pinned: ! pinned } ) }>
+								<GoogleMapApiRender address={ address } />
+							</div> :
 							GoogleMapIframeRender
 						}
 					</ResizableBox>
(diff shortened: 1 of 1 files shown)


richtabor:
Just tested it out, easy map looks/works great. What are you thinking for the advanced map? If we need to loose click control of the map controls, I think that's fine (as they aren't saving actual attributes or anything and are more for end users).

EvanHerman:
@richtabor I just did some testing and it looks like the advanced map is going to be a bit more challenging. The double clicks are hijacked by the Google maps API (for zoom functionality), so I can't track double clicks, and the react-google-maps component (https://github.com/tomchentw/react-google-maps) doesn't support `onDoubleclick`. We may need to think of a solution for the advanced block.

EvanHerman:
@richtabor The advanced map also does not render for me on the front of site, but it does in the editor.

richtabor:
> @richtabor The advanced map also does not render for me on the front of site, but it does in the editor.

I'm not getting that, could you screen-record what's happening?

richtabor:
Yea we may need to add something on top of the block that can be double-clicked. 🤔

EvanHerman:
@richtabor Sorry about that

> > @richtabor The advanced map also does not render for me on the front of site, but it does in the editor.
> 
> I'm not getting that, could you screen-record what's happening?

@richtabor Sorry about that - I didn't have my scripts in the `/dist` directory. Just had to rebuild the files then the map loaded properly.

PR #626, opened 2019-06-29
Remove custom admin footer text

### LINKED ISSUE #625: Task: Remove the custom admin footer text
**Is this related to a problem? Please describe.**
It's not necessary to have a link to leave a review for CoBlocks at the footer of the edit.php page. 

Remove the `CoBlocks_Admin_Footer` class (therefore removing the admin footer text). We could then also remove the `COBLOCKS_SHOP_URL` constant, as this is the only place it's used.

**Current**
Screen Shot 2019-06-28 at 6 47 18 PM

domclone:
This should resolve #625 

1. Deleted `class-coblocks-admin-footer.php` as it's not necessary to have a link to leave a review for CoBlocks at the footer of the `edit.php` page.

2. Deleted the `COBLOCKS_SHOP_URL` constant, as the `class-coblocks-admin-footer.php` file is the only place it was referenced.


richtabor:
Thanks for jumping in @domclone. While the removals are great, the plugin is still trying to load the `class-coblocks-admin-footer.php` file [here](https://github.com/godaddy/coblocks/blob/master/class-coblocks.php#L145), which is causing a fatal error (screenshot below). 

It's a small miss/quick fix, though if missed its quite a bit deal. Please double-check the proposed PR locally before submitting it next time. Thanks!

Screen Shot 2019-07-01 at 10 51 06 AM

richtabor:
Closing this as the issue was recently resolved and merged with #632

PR #630, opened 2019-07-01
Removed Icon Keyword From Share and Features Blocks

paranoia1906:
The blocks already appear for their own names and I could not think of any other keywords to use that would not seem excessive. What are your thoughts?


paranoia1906:
This pull closes https://github.com/godaddy/coblocks/issues/620

Issue #619, opened 2019-06-28
ISNBAT edit the HTML of the Form field blocks


richtabor:
**Describe the bug**
It's currently possible to edit the HTML of the child blocks within the Form block. This functionality does not work as the block is a derivative of InnerBlocks, and will cause the block to break if edited. 

**To Reproduce**

1. Add a Form block
2. Select a field child block
3. Click on the vertical ... to open the panel
4. See "Edit as HTML" listed

**Screenshots**
<img width="973" alt="Screen Shot 2019-06-28 at 12 13 57 PM" src="https://user-images.githubusercontent.com/1813435/60356478-17396d00-999f-11e9-8e1f-244f4a7fe424.png">

**Expected behavior**
The option edit the HTML of a Form field block is not present, using the following:
```
supports: {
	html: false,
},
```

**Isolating the problem:**
Mark completed items with an [x].

-   [x] This bug happens with no other plugins activated
-   [x] This bug happens with a default WordPress theme active
-   [x] This bug happens without the Gutenberg plugin active
-   [x] I can reproduce this bug consistently using the steps above

**WordPress Environment**

paranoia1906:
https://github.com/godaddy/coblocks/pull/631 closes this issue.

Issue #618, opened 2019-06-28
ISNBAT save the Form fields blocks as reusable blocks


richtabor:
**Describe the bug**
It's currently possible to save form fields as reusable blocks. This functionality is not necessary ,or useful, in this context - so we should opt out of it at the block level.

**To Reproduce**

1. Add a Form block
2. Select a field child block
3. Click on the vertical ... to open the panel
4. See "Add to Reusable blocks" listed

**Screenshots**
<img width="973" alt="Screen Shot 2019-06-28 at 12 13 57 PM" src="https://user-images.githubusercontent.com/1813435/60356159-40a5c900-999e-11e9-9426-4839055656ff.png">

**Expected behavior**
The option to set a field as a reusable block is not present, using the following: 

```
supports: {
	reusable: false,
},
```

**Isolating the problem:**
Mark completed items with an [x].

-   [x] This bug happens with no other plugins activated
-   [x] This bug happens with a default WordPress theme active
-   [x] This bug happens without the Gutenberg plugin active
-   [x] I can reproduce this bug consistently using the steps above

**WordPress Environment**
1. What version of WordPress are you using? 5.2.2
2. What version of the Gutenberg plugin are you using? (If any)

paranoia1906:
https://github.com/godaddy/coblocks/pull/631 closes this issue.

paranoia1906:
Hey @richtabor, Looks like we have merged the fix for this issue. We should be safe to close this issue as resolved.

PR #631, opened 2019-07-02
Disable reusable and editable HTML support for Form > Field blocks

paranoia1906:
Handles https://github.com/godaddy/coblocks/issues/618 & https://github.com/godaddy/coblocks/issues/619
Added options to `supports` object to disable `html` and `reusable` properties.

### DIFF
--- a/src/blocks/form/edit.js
+++ b/src/blocks/form/edit.js
@@ -85,6 +85,8 @@ export const childBlocks = [
 			icon: icons.name,
 			supports: {
 				inserter: false,
+				html: false,
+				reusable: false,
 			},
 			edit: props => (
 				<CoBlocksFieldName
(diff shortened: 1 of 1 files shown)


src/blocks/form/edit.js
```diff
@@ -85,6 +85,8 @@ export const childBlocks = [
 			icon: icons.name,
 			supports: {
 				inserter: false,
+				html: false,
```
richtabor: We could move these to the `FieldDefaults` const up above (along with `inserter: false`) to clean this up, instead of repeating it each time. The entire `supports` could be part of `FieldDefaults`.

paranoia1906: FieldDefaults now has the correct supports. Removed supports from individual children blocks. 
```javascript
supports: {
		reusable: false,
		html: false,
		inserter: false,
	},
```

paranoia1906:
Also linted in this commit.

richtabor:
Closes #618 and closes #619

PR #632, opened 2019-07-02
Remove Admin Footer

paranoia1906:
A few notes about this pull request. 
- The constant `COBLOCKS_SHOP_URL` is used in another location. [class-coblocks-url-generator.php](https://github.com/paranoia1906/coblocks/blob/ed9e6884268ed4df53bd9cdb5993859dd154c333/includes/admin/class-coblocks-url-generator.php#L47): 
- I went ahead and replaced the constant with a string.
- Deleted the `class-coblocks-admin-footer.php` file.
- Removed the reference to this file from `class-coblocks.php`.
- Removed declaration of `COBLOCKS_SHOP_URL` from `class-coblocks.php`.
- Removed reference from `coblocks.pot` for `class-coblocks-admin-footer.php`.


includes/admin/class-coblocks-url-generator.php
```diff
@@ -44,7 +44,7 @@ public function get_store_url( $path = '', $params = array() ) {
 
 		$params = array_merge( $params, $id );
 
-		$url = trailingslashit( COBLOCKS_SHOP_URL . $path ) . '?' . http_build_query( $params, '', '&#038;' );
+		$url = trailingslashit( 'https://coblocks.com/' . $path ) . '?' . http_build_query( $params, '', '&#038;' );
```
EvanHerman: @paranoia1906 I'd probably use [add_query_arg()](https://developer.wordpress.org/reference/functions/add_query_arg/) here.

paranoia1906: Using the built-in `add_query_arg()` makes sense to me. Commit is updated to include this change. Here an example of output URL. 

New: 
```
https://coblocks.com/?utm_medium=coblocks-lite&utm_source=plugins-page&utm_campaign=plugins-row&utm_content=help-and-faqs
```
Old: 
```
https://coblocks.com/?utm_medium=coblocks-lite&utm_source=plugins-page&utm_campaign=plugins-row&utm_content=help-and-faqs
```

richtabor:
Looks good!

Issue #621, opened 2019-06-28
Task: Change Share block PanelBody title to `Icon Settings`


richtabor:
**Is this enhancement related to a problem? Please describe.**
Now that we've condensed the Icon and Share settings into one panel, we should properly title it `Icon Settings` (from Share Settings). 

We're also planning to implement a separate Social Profiles block, which will reuse much of this UI. Changing the title will ensure parity between the two blocks from a UI perspective.

**Describe the solution you'd like**
Change the existing PanelBody title `Share Settings` to `Icon Settings`.

**Additional context**
<img width="1097" alt="Screen Shot 2019-06-28 at 5 51 03 PM" src="https://user-images.githubusercontent.com/1813435/60373195-57174900-99cd-11e9-9391-dd2a56fd9d27.png">

zelol:
Can I work on this?

zelol:
"The guidelines for contributing" link on the github home page does not work.

richtabor:
> "The guidelines for contributing" link on the github home page does not work.

Thanks for the catch. We've actually fixed that this am. [Here's the correct link](https://github.com/godaddy/coblocks/blob/master/.github/CONTRIBUTING.md) to the guidelines. And certainly, you're welcome to it!

paranoia1906:
Pull https://github.com/godaddy/coblocks/pull/612 closes this issue. Pull also handles https://github.com/godaddy/coblocks/issues/607 and partially handles https://github.com/godaddy/coblocks/issues/504 as the block is re-factored.

zelol:
Okay!

paranoia1906:
https://github.com/godaddy/coblocks/pull/639 to close this issue.

richtabor:
Closed by #639

PR #639, opened 2019-07-02
Update Panel Title to 'Icon Settings'. #621

paranoia1906:
-Swap out title from 'Social Settings' to 'Icon Settings'.


paranoia1906:
Maybe I am overthinking the coblocks.pot file. Those are rebuilt automatically on each release yes? Maybe it's useless to make those changes?

richtabor:
> Maybe I am overthinking the coblocks.pot file. Those are rebuilt automatically on each release yes? Maybe it's useless to make those changes?

Correct, that's really only necessary when building releases.

PR #642, opened 2019-07-03
Deleted removed files from SVN during WordPress.org deployment

EvanHerman:
Resolves the issue where removed files here on Github are not removed from the WordPress.org SVG repo during the deployment.

### DIFF
--- a/.dev/bin/deploy-plugin.sh
+++ b/.dev/bin/deploy-plugin.sh
@@ -9,7 +9,7 @@ cp -a build/coblocks/* $HOME/coblocks/trunk/
 
 # Copy the WordPress.org assets over
 rm -rf $HOME/coblocks/assets/*
-cp -a wordpress-org/* $HOME/coblocks/assets/
+cp -a .wordpress-org/* $HOME/coblocks/assets/
 
 # Deploy Coblocks to WordPress.org
 cd $HOME/coblocks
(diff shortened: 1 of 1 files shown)


.dev/bin/deploy-plugin.sh
```diff
@@ -9,11 +9,13 @@ cp -a build/coblocks/* $HOME/coblocks/trunk/
 
 # Copy the WordPress.org assets over
 rm -rf $HOME/coblocks/assets/*
-cp -a .wordpress-org/* $HOME/coblocks/assets/
```
richtabor: Is the removal of the `.` intentional?

EvanHerman: @richtabor Yeah, that was intentional. It was unintentionally included. Even tho it still worked, it wasn't correct. It really should have been `./wordpress-org/* $HOME/coblocks/assets/` but since were in the `project` directory in CirlceCI we can remove the `./`, similar to how we're doing it here https://github.com/godaddy/coblocks/blob/master/.dev/bin/deploy-plugin.sh#L8

EvanHerman: @richtabor Shoot!

EvanHerman: @richtabor You are correct. That should not have been removed. I just remembered that we issued a previous PR and moved everything into a `.wordpress-org` directory from a `wp-org-assets` directory - so that `.` is actually needed there. My mistake.

Issue #641, opened 2019-07-03
ISBAT apply a default background color to the Features block in TwentyNineteen


richtabor:
**Describe the bug**
Applying a default color palette color selection to the Features block does not properly display while using the TwentyNineteen theme. 

This only occurs with TwentyNineteen, and only within the editor. Custom colors work fine (as they are inline-styles).

**To Reproduce**

1. Install/Activate the TwentyNineteen theme
2. Add a Features block
3. Open the block's Settings Sidebar
4. Set the Background Color to any of the default color picker options
5. See error...

**Screenshots**
![2019-features](https://user-images.githubusercontent.com/1813435/60601460-27c35c00-9d80-11e9-9d4d-8a5ac0fbcf8a.gif)

**Expected behavior**
The background color class works properly, regardless of the active theme

**Isolating the problem:**
Mark completed items with an [x].

-   [x] This bug happens with no other plugins activated
-   [x] This bug happens with a default WordPress theme active
-   [x] This bug happens **without** the Gutenberg plugin active
-   [x] I can reproduce this bug consistently using the steps above

**WordPress Version**
WordPress 5.2, TwentyNineteen 1.4

**Gutenberg Version**
Happens with, and without, Gutenberg 6.0 active.

jrtashjian:
I'm experiencing this issue in multiple themes.

**Twenty Seventeen**
Notice that the text color is not correctly set but the background color is.

![twentyseventeen](https://user-images.githubusercontent.com/375788/60733379-afdd6900-9f1a-11e9-8daa-f5823d804d04.png)

**CoBlocks Theme**

![coblocks](https://user-images.githubusercontent.com/375788/60733472-ee732380-9f1a-11e9-94cb-c822c6c8444a.png)

**Twenty Fifteen**
Background and text colors are not set.

![twentyfifteen](https://user-images.githubusercontent.com/375788/60733523-1b273b00-9f1b-11e9-8609-8faba05e5148.png)

jrtashjian:
The background and text color selection works flawlessly in the Underscores base theme with no editor styles. Added the new theme with the `wp scaffold underscores` WP-CLI command.

My opinion is whenever we need to make style updates for a specific theme, we should separate these changes/overrides from the base styles that work with the editor without a theme editor stylesheet. This would also require more compiled css files and loading styles for only the active theme.

It will also add complexity in which themes we decide to support officially or we only provide style fixes for GoDaddy, ThemeBeans, and WordPress Core themes.

PR #646, opened 2019-07-04
Features block background color fix

nicolad:
Fixes: https://github.com/godaddy/coblocks/issues/641

## How has this been tested?
Peek 2019-07-04 23-55


nicolad:
@richtabor let me know if this is ok

jrtashjian:
@nicolad although it looks like this fixes the issue present, there looks to be a `BackgroundStyles` function being called before your fix. Look into applying the fix to the `src/components/background/styles.js` file where that function is defined and the background color could be defined.

jrtashjian:
Also in #641 I found that depending on the theme active the background or text color might not be applied in the editor.

src/components/background/styles.js
```diff
@@ -6,17 +6,16 @@ const { getColorClassName } = wp.blockEditor;
...
-	const backgroundClass = getColorClassName( 'background-color', attributes.backgroundColor );
+const BackgroundStyles = ( attributes, backgroundColor ) => {
+	const backgroundClass = attributes && getColorClassName( 'background-color', attributes.backgroundColor );
 
 	const styles = {
-		backgroundImage: attributes.backgroundImg && attributes.backgroundType == 'image' ? `url(${ attributes.backgroundImg })` : undefined,
-		backgroundColor: backgroundClass ? undefined : attributes.customBackgroundColor,
-		backgroundPosition:  attributes.focalPoint && !  attributes.hasParallax ? `${  attributes.focalPoint.x * 100 }% ${  attributes.focalPoint.y * 100 }%` : undefined,
+		backgroundImage: attributes.backgroundImg && attributes.backgroundType === 'image' ? `url(${ attributes.backgroundImg })` : undefined,
+		backgroundColor: backgroundClass ? backgroundColor && backgroundColor.color : attributes.customBackgroundColor,
```
nicolad: Optional chaining might be used here to avoid short-circuit evaluation: https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining

PR #647, opened 2019-07-04
Hero block background fix

nicolad:
Fixes: https://github.com/godaddy/coblocks/issues/636
## How has this been tested?
Peek 2019-07-05 00-06


nicolad:
@richtabor let me know if this is suffice

jrtashjian:
@nicolad just like in #646: look into applying the fix to the `src/components/background/styles.js` file where that function is defined and the background color could be defined.

nicolad:
> @nicolad just like in #646: look into applying the fix to the `src/components/background/styles.js` file where that function is defined and the background color could be defined.

@jrtashjian has moved, thx.

PR #658, opened 2019-07-08
Minor dependency updates

glendaviesnz:
Some of the current package dependencies have listed vulnerabilities. I have checked and none of them have any impact on CoBlocks at all, but it might be good to update the package.json to cover them off:

* shelljs (https://snyk.io/vuln/npm:shelljs:20140723) as far as I can see is not currently used anywhere in CoBlocks so could be removed

* Lodash (https://snyk.io/vuln/SNYK-JS-LODASH-450202) the affected method is not currently used by CoBlocks. A patch for this vulnerability is due out in the next few days, but currently the package.json is locked to the vulnerable version, so suggest adding ^ to version number so patched version is picked up


EvanHerman:
LGTM 👍- thanks @glendaviesnz !

PR #664, opened 2019-07-10
Add Social Profiles block

paranoia1906:
- Link social profile urls from your website.
- Transforms to and from `share` block.


richtabor:
Good stuff, let's tweak the new icon colors to the following hex values: 

Instagram: #ef0175
Houzz: #4ebc16
Yelp: #d32323
Youtube: #ff0000

src/blocks/social-profiles/block.json
```diff
@@ -0,0 +1,80 @@
+{
```
richtabor: Is the block.json on this block is registered server-side? I noticed that core Gutenberg does not have block.json files for their server-side blocks (Latest Posts/Comments).

paranoia1906: You are correct that core blocks which are registered server side do **not** include block.json. In this case we were only using them to import the name of the block. This has been replaced with a string and the block.json has been removed in my latest commits.

richtabor:
Nice work. Let's try setting an opacity of 0.5 if there there are no social links added.

Screen Shot 2019-07-10 at 12 11 15 PM

richtabor:
Also, let's explore merging the two blocks' editor styles and then their frontend styles - as they're essentially the same.

richtabor:
Here's the updated flow for adding new icons which does not require access to the Settings Sidebar. 👌

social-profiles

richtabor:
> Nice work. Let's try setting an opacity of 0.5 if there there are no social links added.

We don't need this now that the flow is tweaked.

src/blocks/social-profiles/index.php
```diff
@@ -207,7 +207,7 @@ function coblocks_render_social_profiles_block($attributes)
 		esc_attr($text_align),
 		$markup
 	);
-
+	echo esc_attr($block_content);
```
jrtashjian: This bit is breaking the block. Need to return a string, not echo it. I'm guessing this was for debugging purposes.

src/blocks/share/index.js
```diff
@@ -15,19 +14,15 @@ const { __, _x } = wp.i18n;
...
 
 const title = __( 'Share' );
 
 const icon = icons.social;
 
-const keywords = [
-	__( 'social' ),
-	__( 'coblocks' ),
-];
+const keywords = [ __( 'social' ), __( 'coblocks' ) ];
```
jrtashjian: @richtabor We should rethink the keywords here. We are limited to just three but we don't need to add 'social' because it's the block name, and we don't need to add 'coblocks' because it's within that group. Both of these string locations are handled by search.

src/blocks/social-profiles/inspector.js
```diff
@@ -0,0 +1,217 @@
+/**
+ * External dependencies
+ */
+import includes from 'lodash/includes';
+import escape from 'lodash/escape';
```
jrtashjian: I believe you can also do this:

 `import { includes, escape } from 'lodash';`

src/blocks/social-profiles/inspector.js
```diff
@@ -0,0 +1,217 @@
...
+								value={ twitter }
+								onChange={ value => setAttributes( { twitter: escape( value ) } ) }
+							/>
+							<TextControl
+								label="Instagram"
+								value={ instagram }
+								onChange={ value => setAttributes( { instagram: escape( value ) } ) }
+							/>
+							<TextControl
+								label="Pintrest"
```
jrtashjian: Typo. Should be "Pinterest"

src/blocks/social-profiles/edit.js
```diff
@@ -0,0 +1,493 @@
...
+									) }
+									style={ buttonStyles }
+									onClick={ () => this.setState( { currentIcon: 'youtube' } ) }
+								>
+									<span
+										className="wp-block-coblocks-social__icon"
+										style={ iconStyles }
+									/>
+									<span className="wp-block-coblocks-social__text">
+										{ __( 'See us on Youtube' ) }
```
jrtashjian: Typo: Should be 'YouTube'

src/blocks/social-profiles/inspector.js
jrtashjian: Typo: Should be 'Pinterest'

src/blocks/share/index.php
```diff
@@ -306,9 +308,13 @@ function coblocks_register_share_block() {
 					'type'    => 'boolean',
 					'default' => false,
 				),
+				'socialProfileAttributes'		=> array(
```
jrtashjian: I would avoid adding a new attribute to a block for the purpose of different block. I understand this is for transforming back to the social-profile block. The method works and I don't think we absolutely have to change this.

My initial thought is to use localStorage to persist the urls during the transform so we can read them back. We can read them back if we stay in the editor. I think it's a valid expectation that if you leave the editor you've "finalized" the edit and wouldn't expect the urls to be remembered once transformed into the social block.

src/blocks/social-profiles/index.php
```diff
@@ -0,0 +1,257 @@
...
+				),
+				'yelp'                => array(
+					'type'    => 'string',
+					'default' => '',
+				),
+				'houzz'                 => array(
+					'type'    => 'string',
+					'default' => '',
+				),
+				'shareAttributes'		=> array(
```
jrtashjian: Same comment with the socialProfileAttributes on the share block.

src/blocks/social-profiles/edit.js
richtabor: And the "See us on" on this and other icons should be removed, and just print the icon label.

src/blocks/share/index.php
paranoia1906: Rewrote transforms and removed cross-block attributes in favor of localStorage use.

src/blocks/social-profiles/index.php
paranoia1906: Rewrote transforms and removed cross-block attributes in favor of localStorage use.

src/blocks/social-profiles/inspector.js
paranoia1906: This has been done.

src/blocks/share/index.js
paranoia1906: This is now `[ __( 'share' ), __( 'links' ), __( 'icons' ) ];`

PR #665, opened 2019-07-10
WIP: Moving column implementation to a block supports feature

jrtashjian:
This branch is exploring a way to apply specific features to different blocks, preventing re-implementation in each block.

Some benefits to this:
1. Standardization of the UI/UX for individual features.
2. Adding support to a block is as easy as adding [block supports](https://developer.wordpress.org/block-editor/developers/block-api/block-registration/#supports-optional) on block registration.
3. No need to add attributes for the feature in the individual block attribute definition. Although they can be overridden in this way.
4. Removes any duplication of features which cleans up the codebase.
5. Exposes the feature to any third-party blocks that want to use these features so long as CoBlocks is installed and active.

In this PR I've created a `coblockColumns` feature and applied it to the Features block. The new feature adds a new "Column Settings" Panel with a controls for the number of columns and the gutter size within the InspectorControls.

**New Panel on the Features block**

Screenshot from 2019-07-10 16-00-36

**Masonry Block controls are inconsistent**

Screenshot from 2019-07-10 16-18-25

The `coblockColumns` attribute is passed into our block so we can use it normally. We need to explore automatically appending related CSS classes to the block as well as centralizing the CSS styles.

The next pieces I think we can simplify this way are the dimensions control, typography controls, background, and grid controls.

### DIFF
--- a/src/blocks.js
+++ b/src/blocks.js
@@ -31,6 +31,9 @@ import './extensions/button-controls';
 // Formats
 import './formats/';
 
+// Hooks
+import './hooks/columns';
+
 // Sidebars
 import './sidebars/block-manager';
 import './sidebars/block-manager/deprecated';
--- a/src/blocks/features/components/controls.js
+++ b/src/blocks/features/components/controls.js
@@ -27,7 +27,6 @@ class Controls extends Component {
 
 		const {
 			contentAlign,
-			columns,
 		} = attributes;
 
 		return (
--- a/src/blocks/features/components/edit.js
+++ b/src/blocks/features/components/edit.js
@@ -62,9 +62,9 @@ class Edit extends Component {
 		const {
 			coblocks,
 			backgroundImg,
-			columns,
+			coblockColumns,
 			contentAlign,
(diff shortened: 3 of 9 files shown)


jrtashjian:
Latest changes migrate the column support in the Masonry Gallery away from the "Masonry Settings" panel. Column sizes from the old panel has been converted to number format (min 2 and max 4). Gutter was converted from a range control to the select control. The responsive controls have not been migrated over yet and it would be something to think about how to handle.

Here's a quick gif of the changes working:

columns-masonry-gallery

The integration was quick and dirty just to show the concept. I'm actually doing a conversion from column numbers to the existing string values ('med', 'lrg', 'xlrg') because there are conditionals and functionality that check for those string values still.

richtabor:
Yea the column size in the masonry block is as such because if we set a hard number, and you’re on mobile, the column count may not line up. For example, five columns can’t stay five columns on mobile.

jrtashjian:
Closing this for now. We can come back to it another time.

PR #666, opened 2019-07-11
WIP: Add Services block

jrtashjian:
## Description

New block used to highlight the services of a business with a call-to-action so that potential customers know what services are offered and how to get started.

## Features

- Display an optional image related to the service..
- Display the service's name.
- Display the service's description.
- Display an optional call-to-action button.
- Ability to display up to 3 items within a single block using a column layout.
- Three styles available around the style of the image with text alignment modifications applied: 16:9, square, and circle.
- Call-to-action buttons are able to be toggled at the parent level as well as on individual services.
- Heading level is able to be toggled at the parent level as well as on individual services.
- Text alignment is able to be toggled at the parent level as well as on individual services.

## Screenshots

**Initial State**

services-block-initial-state

**Call to Action Toggled**

services-block-call-to-action-toggle

**Style: Square**

services-block-style-square

**Style: Circle**

services-block-style-circle

## Remaining Cleanup Tasks

- [x] Pull inspector controls into its own file.
- [ ] ~Abstract shared function for updating inner block attributes.~
- [x] Clean up CSS styles.
- [x] Add CSS styles for mobile.
- [x] Add image focal point selection.
- [x] Resolve an issue preventing image replacement with drag-and-drop (initial drag-and-drop upload works).
- [x] Ensure block attributes that can be pulled from the HTML are pulled from the rendered HTML.
- [ ] ~Integrate Service microdata within the rendered HTML (https://schema.org/Service).~
- [x] Add svg icons for each available style.


src/blocks/services/styles/editor.scss
```diff
@@ -0,0 +1,59 @@
...
+			display: block;
+		}
+	}
+
+	.wp-block[data-type="coblocks/service"] {
+		flex: 1 100%;
+		max-width: 100%;
+
+		> .block-editor-block-list__block-edit {
+			margin-top: 1em;
```
richtabor: What are these margin and bottom parts applied for? They seem to be affecting the block selection border.

src/blocks/services/styles/editor.scss
```diff
@@ -0,0 +1,59 @@
...
+			margin-bottom: 1em;
+		}
+	}
+
+	@media all and (min-width: 514px) {
+		> .block-editor-block-list__block-edit {
+			margin-top: 0;
```
richtabor: Same here, what are these margin and bottom parts applied for?

src/blocks/services/styles/editor.scss
```diff
@@ -0,0 +1,59 @@
...
+			max-width: calc(100% / 2);
+		}
+
+		&[data-columns="3"] .wp-block[data-type="coblocks/service"] {
+			flex: 1 calc(100% / 3);
+			max-width: calc(100% / 3);
+		}
+	}
+
+	.wp-block-image .block-editor-rich-text,
```
richtabor: Is this necessary?

jrtashjian: Nope, this is left over from when I was trying to utilize the image block but eventually just used the image functionality from our Food and Drinks block. It'll be removed.

src/blocks/services/styles/editor.scss
jrtashjian: This was put in place to make the block look the same within the editor and the theme. I'm not satisfied with this either so I'm going to find a different solution.

PR #672, opened 2019-07-17
Add Carousel thumbnails/responsive height and Masonry lightbox

grabovacnem:
Enhanced carousel and masonry blocks


richtabor:
Looking good. Here are a few thoughts : 

**Masonry:**
1. Should we expandable image as an additional class? What happens if that class is changed by the user? Is the toggle still turned on? What happens if there are other classes already added to that advanced control. Could we just use a data attribute perhaps? Or add a class in the edit/save functions when the toggle is active. 

2. We should adjust the lightbox class names to use BEM style markup. 

3. Let's rename `coblocks-expand-image.js` to `coblocks-lightbox.js`.

4. Add a cursor hover effect to the images if lightbox is enabled.

**Carousel:**
1. When you toggle the responsive height on, the gallery within the editor should be set to the correct height and not be resizable. 

2. Thumbnails should have the margin top that is the same as the gutter, instead of left/right margin.

richtabor:
Closed in favor of #733

PR #673, opened 2019-07-17
WIP: Add Blogroll block

dzoniub:


richtabor:
Looking good, here are a few thoughts: 

1. I'm thinking that “Blogroll” may be a better block name. 

2. The category selector should not be available when in external URL mode, as it's not possible to set. 

3. The number of items control should have a lower max. I'm thinking 20 max. 

4. As discussed, let's use icons in the Styles, just like we’re doing in the Food & Drinks block. You can go ahead and lay the code and I'll prep some icons for you. 

5. Let's use the same left/right content align icons from the core Media & Text block (the same SVGs) for the List style alignment icons. 

6. Let's use the following keywords instead: `posts`, `articles`

7. Ensure the block displays within the CoBlocks category of the Block Inserter

8. Need to prefix functions within the block's index.php file and run PHPCS.

9. Combine styles/editor.scss and carousel-settings/editor.scss

10. Move inspector components into its own inspector.js

11. Let's use the same edit icon as the Events block (instead of a separate button) to trigger the editing of the blog URL, instead of the button within the Settings Sidebar.

12. Resolve deprecation notices with the Gutenberg 6.1.1 plugin installed and activated.

PR #674, opened 2019-07-17
Fix block food-and-drink display in IE11

### LINKED ISSUE #654: ISBAT view the Food & Drinks block properly in Internet Explorer 11
The Food & Drinks block does not render properly in Internet Explorer 11. 

**Safari Screenshot**
Screen Shot 2019-07-05 at 3 52 28 PM

**IE 11 Screenshot**
Screen Shot 2019-07-05 at 3 52 39 PM

paranoia1906:
- Closes #654 
- Fix for `Object.fromEntries` for IE11 backwards compatibility.
  * According to [Mozilla.org](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries#Browser_compatibility) the incompatibilty also applies to Edge and Opera.
- Targeted CSS specifically for IE11. 

Notes: Due to compatibility issue with `object-fit` and IE 11 images do not resize correctly. This fix uses an absolute image position for IE11.


jrtashjian:
I updated the code so that images are contained within the container rather than stretched on IE 11. Object-fit is not supported but there does not seem to be an acceptable workaround without adding some JS or setting the image as a background-image. We don't want to use a polyfill on the front-end and we can't use the background-image for accessibility reasons since the image adds context to the menu.

Here's what it looks like now:

Screen Shot 2019-07-23 at 11 10 13 AM

src/blocks/food-and-drinks/food-item/edit.js
```diff
@@ -4,6 +4,7 @@
 import { hasEmptyAttributes } from '../../../utils/block-helpers';
 import InspectorControls from './inspector';
 import icons from './icons';
+import fromEntries from '../../../js/coblocks-fromEntries';
```
jrtashjian: Just wanted to note what we had already talked about for future reference: We should let Babel handle transpiling rather than adding a polyfill ourselves manually. I looked into this and it seems core-js v3 package includes this polyfill but we are currently using core-js v2. We'll need to upgrade and resolve any issues sometime in the future.

We can leave this here for now.

src/blocks/food-and-drinks/food-item/styles/style.scss
```diff
@@ -1,9 +1,25 @@
+
+:-ms-fullscreen .wp-block-coblocks-food-item,
+:root .wp-block-coblocks-food-item {
```
jrtashjian: I updated the selectors so we're only targeting our own block rather than using selectors that only exist within a theme. These selectors are a sweet add because it seems to just target IE like we want.

richtabor:
@paranoia1906 I found an issue when viewing on mobile (Chrome). 

Master: 
Screen Shot 2019-07-23 at 4 43 01 PM

Branch: 
Screen Shot 2019-07-23 at 4 43 39 PM

PR #681, opened 2019-07-21
Add placeholder and ascending plan titles to the Pricing Table block

richtabor:
Before: 
Screen Shot 2019-07-22 at 10 35 38 AM

After: 
Screen Shot 2019-07-22 at 10 36 08 AM

### DIFF
--- a/src/blocks/pricing-table/pricing-table-item/components/edit.js
+++ b/src/blocks/pricing-table/pricing-table-item/components/edit.js
@@ -27,7 +27,7 @@ const { RichText, InnerBlocks } = wp.blockEditor;
  * @type {string[]}
 */
 const ALLOWED_BLOCKS = [ 'core/button' ];
-const TEMPLATE = [ [ 'core/button', { text: __( 'Buy Now' ) } ] ];
+const TEMPLATE = [ [ 'core/button', { placeholder: __( 'Buy Now' ) } ] ];
 
 /**
  * Block edit function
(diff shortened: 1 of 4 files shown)


src/blocks/pricing-table/pricing-table-item/components/edit.js
```diff
@@ -27,7 +27,7 @@ const { RichText, InnerBlocks } = wp.blockEditor;
  * @type {string[]}
 */
 const ALLOWED_BLOCKS = [ 'core/button' ];
-const TEMPLATE = [ [ 'core/button', { text: __( 'Buy Now' ) } ] ];
+const TEMPLATE = [ [ 'core/button', { placeholder: __( 'Buy Now' ) } ] ];
```
jrtashjian: To my surprise the `core/button` block doesn't support setting a placeholder. It will always default to "Add Text..."

richtabor: Oh right, I should have made that clearer. It's not now, but it should be eventually via: https://github.com/WordPress/gutenberg/issues/16618

🤞

richtabor: > Should we allow html to be saved to the post_content if there is only placeholder content? We can utilize the `isEmpty` functionality being used in the Food and Drinks block to detect the empty child blocks.
> 
> Looks good otherwise. No changes required.

Good idea, I like that.

PR #704, opened 2019-07-28
Refactor gist block to allow gist URLs be deleted without issue

### LINKED ISSUE #702: Removing the URL from the gist block causes the block to be unusable
**Describe the bug**
If you remove the URL on an existing GIST block, the gist is snippet(s) is removed (expected), but the block becomes an empty block without the URL input field.

**To Reproduce**

1. Create a gist block with the following URL https://gist.github.com/mayeenulislam/5b4580fe49c9eb8c319ce3cc69cb0a68#file-tooltip-2-css and save the page
2. Edit the gist block and delete the URL
3. The gist block is now an empty block

**Screenshots**
gist-bug

**Expected behavior**
After removing the URL from the gist block, I expect the gist block to revert to it's initial state with an input field where I can add another gist URL.

**Isolating the problem:**

**WordPress Version**
Latest 5.2.2

**Gutenberg Version**
None

EvanHerman:
Resolves #702 

## Before Patch
gist-bug

## After Patch
delete-gist-url


paranoia1906:
When switching to a new Gist and no file is specified, the file property remains as set by previous Gist.

### Before 
gistFileAttribute

Made a slight change in edit.js to handle this behavior and correctly purge the file property.

### After
gistFileAttributeAfter

PR #706, opened 2019-08-01
add withNotices to handle Gist block errors

### LINKED ISSUE #601: Gist Block handle invalid Gist URL
**Describe the bug**
The bug is a lack of meaningful error message when a non-existent or invalid Gist url is used. Browser console shows a 404 on the URL fetch but error is not displayed visually to end user.

**To Reproduce**
1. Edit any page.
2. Add CoBlocks Gist.
3. Edit Gist and add url that does not contain a Gist (e.g. https://google.com)
4. Spinning wheel renders by default and does not resolve.
5. Open browser developer console to see the issue.

**Screenshots**
GistEndlessSpinner

**Expected behavior**
If the fetch returns a non 200 response then there should be a helpful message to end-user. A message such as “No Gist Found at ‘https://google.com’” or “Error Loading Gist! URL does not resolve to a valid Gist.” would identify the issue clearly.
GistSpinnerResolved

**Isolating the problem:**

-   [X] This bug happens without the Gutenberg plugin active

**WordPress Environment**

<details>
```
What version of WordPress are you using? 5.2.2
What version of the Gutenberg plugin are you using? N/A
```
</details>
```

paranoia1906:
- Adds withNotices HOC to the Edit component.
- Closes #601.

gistErrorHandler


fjarrett:
I think the error message could be better here. Something like:

> _The provided URL does not appear to be a Gist_

@richtabor Thoughts?

richtabor:
> I think the error message could be better here. Something like:
> 
> > _The provided URL does not appear to be a Gist_
> 
> @richtabor Thoughts?

Yea, that could be fine-tuned a bit. The image block uses "Sorry, this file type is not permitted for security reasons." when an unsupported image is uploaded. 

We could follow this same language as such: "Sorry, this URL is not a GitHub gist."

paranoia1906:
Error message text now says "Sorry, this URL is not a GitHub gist.".

PR #707, opened 2019-08-01
Add cancel button to Map block

### LINKED ISSUE #638: ISBAT cancel updating the Map block location
To follow core's design and UX patterns, a user should be able to cancel editing a Map location from the edit view, without having to first apply the address or click off of the block.

Add a "Cancel" button under the location/address field, just as the core Image block has added in Gutenberg 6. Pressing "Cancel" should revert any current changes and bring up the map view. Perhaps the location/address field should not update with every key, but instead with the Apply button?

**Additional context**
Screen Shot 2019-07-02 at 4 33 29 PM

paranoia1906:
- Cancel editing of URL with Map block.
- Similar implementation to that of core. 
- Closes #638 

mapBlockCancel


src/blocks/map/components/edit.js
```diff
@@ -47,6 +47,7 @@ class Edit extends Component {
 		this.state = {
 			apiKey: '',
 			address: this.props.attributes.address,
+			addressText: '',
```
jrtashjian: What is the purpose of this new property in state? If we need to track or compare the address to an edited value we can do that with a single property in state and the block attribute. The block attribute (`this.props.attributes.address`) should always hold the final and saved value. The state value should be used in the component. Only after clicking "apply" would the state value be saved as the block attribute value. And cancel should not update the block attribute (and we can reset the state value if we want).

src/blocks/map/components/edit.js
```diff
@@ -299,15 +302,28 @@ class Edit extends Component {
...
 							{ __( 'Apply' ) }
 						</Button>
 
+						<Button
+							className="components-placeholder__cancel-button"
+							title={ __( 'Cancel' ) }
+							isLink
+							onClick={ () => {
+								setAttributes( { pinned: ! pinned } );
+								this.setState( { addressText: address } );
```
jrtashjian: Instead of two properties in state you could change this to `this.setState( { address: this.props.attributes.address } )`

src/blocks/map/components/edit.js
paranoia1906: Thank you for pointing this out. I got a look today and I realized that I was clearly replicating the address state into two separate states. I went ahead and made the change to remove the addressText state. The block is tested and working as expected.

src/blocks/map/components/edit.js
paranoia1906: It seems so obvious now. Thanks for they eyes JR. I learn something new every time you review something of mine.

paranoia1906:
Up to date with master. Tested and still working.

PR #708, opened 2019-08-04
Add Hero toolbar color settings dropdown

### LINKED ISSUE #503: ISBAT easily add a background color without having to go into the Settings Sidebar
It is not currently possible to easily change the color of a block's background without diving into the Color Settings of a block's Settings Sidebar. 

While it is possible to add background media (image/video) in the Block Toolbar, adding a background color requires much more effort. Having background image and color settings in one area (Block Toolbar) would make the editing controls for adding a background much more usable. 

Let's explore how to add background color capability into the background toolbar control from a UX and UI standpoint.

**Current flow:** 

ScreenFlow

paranoia1906:
- Add `Color Picker`/ `Color Settings` item to Hero toolbar.
- Closes #503.

### After enhancement
heroColorPicker


paranoia1906:
Let me know what you think about this enhancement. For the SVG I just used what I thought made sense in the situation, we may want to change the icon.

jrtashjian:
This works well but I think we can make this a bit better using a different component.

If you look inside the [PanelColorSettings](https://github.com/WordPress/gutenberg/tree/wp/5.2/packages/block-editor/src/components/panel-color-settings) component you'll see it's utilizing the [ColorPalette](https://github.com/WordPress/gutenberg/tree/wp/5.2/packages/block-editor/src/components/color-palette) component that I think we can use to simplify the toolbar content a bit. The Panel component being used in that dropdown feels weird and we'll never need to toggle the panel.

I think it would be good to combine the color settings and background image toolbar buttons. I'd like to add some advanced functionality here in the future so combining them will most likely happen eventually. You'll have to prevent the background image toolbar button from showing the media library on click and instead have a button in the dropdown like "select a background image" or something more concise. After an image is added we have the options of "edit" or "remove" for the background image (already exists as it's functionality).

I think it's be great to use that ColorPalette component above the background image options, but all in that single toolbar dropdown.

Kapture 2019-08-05 at 13 15 56

richtabor:
Yup, I'm thinking a way to have Background Color, Edit, and Remove, all in the same control would be beneficial. The dripping paint can icon should be all background focused. 

I'm not sure exactly how that'd work (and it's something core is looking into).

richtabor:
I’m going to close this (at least for now) until we have a solid UI/UX plan in place for #503. Core is also thinking through background support, which we’d likely want to emulate (or hopefully can simply opt-into at the block level).

PR #709, opened 2019-08-04
add logic to maintain consistent placeholders for food-and-drinks block

### LINKED ISSUE #617: Second Food Item placeholder disappears with zero state
**Describe the bug**
Hey guys, great new Food & Drinks block, kudos! Now when I first insert the block and escape click on the inner blocks, one of the items disappear and doesn't come back.

**Update:** 
It does actually come back when I start entering content on the first menu item, but that was confusing as a user. Maybe it would be easier to use an appender instead for adding a new item, or a helper indicating that I will be able to add more. Expecting me to start typing on the first menu item to be able to see the second menu item was unexpected :)

**To Reproduce**

1. Insert a new food & drinks block
2. Click on the first menu item
3. See second menu item vanish! :)

**Screenshots**
foodblock

**Isolating the problem:**
- [* ] This bug happens with no other plugins activated
- [* ] This bug happens with and without the Gutenberg plugin active
- [* ] I can reproduce this bug consistently using the steps above

**WordPress Environment**

<details>

- Wordpress 5.2.2
- Gutenberg 6.0

</details>

paranoia1906:
- Add logic to maintain two placeholders for food-item blocks.
- Should also prevent deleting below two placeholders. 
- Closes #617 

### After changes
foodDrinkTwoPlaceholders


jrtashjian:
This is very close but we do want to maintain a single placeholder. We only want to ensure two placeholder exists if all of the inner blocks are placeholders. So if all children are placeholders, 2 placeholder blocks should be shown.

We already have a check to determine if a placeholder block exists and if not, we add it. I think that might be the place to alter our algorithm to check if all of the blocks are placeholders and ensure two are shown.

paranoia1906:
Good feedback. Made the suggested change to the algorithm.

### Maintain only a single placeholder
foodDrinkTwoPlaceholders

paranoia1906:
- Upgraded placeholder logic.
- Merged with master.
**Behavior after changes**
betterFoodAndDrinkPlaceholders

PR #711, opened 2019-08-07
fix Alert block transform to Paragraph block

### LINKED ISSUE #643: ISBAT transform the Alert block into Paragraph block
**Describe the bug**
When the `Alert` block is transformed into a `core/paragraph` block, multiple identical elements are generated as opposed to the expected 1-to-1 transformation.

**To Reproduce**

1. While editing a page or post add the `Alert` block type.
2. Within the block add several text elements.
3. After adding content to the `Alert` block click on the `Change block type or style` button
4. Next click on `Paragraph` from the drop down menu.
5. See the elements populate.

An example of the text elements. I used: 
"Here is some example text." - Bold
"here is some example text." - Underlined
"here is some example text." - Italics
"here is some example text." - Strike-through
"here is some example text." - Different Font.
Line breaks
(see the screen-capture for more context.)

**Screenshots**
AlertBlockTransfromToParagraph

**Expected behavior**
I would expect a single paragraph element to replace the `Alert` block.

**Isolating the problem:**

**WordPress Version**
What version of WordPress are you using? 5.2.2

**Gutenberg Version**
What version of the Gutenberg plugin are you using? N/A

paranoia1906:
- Closes #643 
- Drops the content in `createBlock` instead of looping through

### After Fix
AlertBlockTransfromToParagraph


jrtashjian:
I'm experiencing a couple issues:

1. HTML tags for bold, italic, link, etc are being escaped when transforming from a paragraph block to an alert block.

2. There is an extra line being added to the top of the content after transforming which might be related to the alert title logic.

Kapture 2019-08-13 at 14 56 42

jrtashjian:
I created a new issue (#742) for this last bit as it shouldn't be part of this PR.

PR #736, opened 2019-08-13
Add new Posts block to display blogroll or RSS feed posts

dzoniub:


richtabor:
@dzoniub Do you mind clearing these conflicts? Thanks!

EvanHerman:
Currently in the process of cleaning up some of these files so they pass PHPCS and ESLint checks in our CI/CD pipeline. PHPCS is done, now working on ESLint (`grunt eslint` to test)

EvanHerman:
Passing PHPCS and ESLint. Last issue I've found is on initial addition of the blogroll block, things do not render correctly. Only after re-selecting a style does the block render properly.

**Note:** Also, you can see in the demo that the 'List' style thumbnails do not crop/render properly.


Blogroll Demo

EvanHerman:
Also of note, there is a `// TODO` comment inside of the icons. I'm assuming this is an indication that we need to swap out the blogroll icon to match the rest of the CoBlocks styles.

https://github.com/godaddy/coblocks/blob/2c433902d5721bdcd3adb0902299360f7415e21a/src/blocks/blogroll/icons.js#L17

dzoniub:
> Also of note, there is a `// TODO` comment inside of `coblocks/src/blocks/blogroll/icons.js` line 17. I'm assuming this is an indication that we need to swap out the blogroll icon to match the rest of the CoBlocks icon styles.
> 
> https://github.com/godaddy/coblocks/blob/2c433902d5721bdcd3adb0902299360f7415e21a/src/blocks/blogroll/icons.js#L17
That is correct, Rich told me that he'd make some proper SVGs for it, but if the TODO is still there, I guess that is still not resolved.

Thanks for the review and the tweaks from your side. Btw, correct me if i'm wrong, but it seems that you accidentally pushed this piece of code here, and I can't find that you changed it in the following commits: https://github.com/godaddy/coblocks/pull/736/commits/fa3cbf4dbf1e14192d4515d0b197ca118414ab18#diff-876237abae7ee7465d178a3f71505f41R352?

EvanHerman:
@dzoniub Wow you are correct. So sorry about that! Thought I had reverted that. I pushed up a change to revert that piece of test code.

src/blocks/posts/inspector.js
```diff
@@ -0,0 +1,224 @@
...
+				<ToggleControl
+					label={ __( 'Post Content' ) }
+					checked={ displayPostContent }
+					help={
+						displayPostContent ?
+							__( 'Showing the post content.' ) :
+							__( 'Toggle to show the post content.' )
+					}
+					onChange={ () => setAttributes( { displayPostContent: ! displayPostContent } ) }
+				/>
```
jrtashjian: We should use the post excerpt WordPress core already provides (and has associated hooks for other plugins/themes to use).

src/blocks/posts/inspector.js
```diff
@@ -0,0 +1,224 @@
...
+				{ displayPostContent &&
+					<RangeControl
+						label={ __( 'Max words in content' ) }
+						value={ excerptLength }
+						onChange={ ( value ) => setAttributes( { excerptLength: value } ) }
+						min={ 5 }
+						max={ 75 }
+					/>
+				}
```
jrtashjian: If we utilize the post excerpt from WordPress, setting the max words in the content would be unnecessary. Filters that exist in core allow this value to be modified.

src/blocks/posts/index.php
```diff
@@ -0,0 +1,417 @@
...
+
+		$formatted_post['thumbnailURL'] = get_the_post_thumbnail_url( $post );
+		$formatted_post['date']         = esc_attr( get_the_date( 'c', $post ) );
+		$formatted_post['dateReadable'] = esc_html( get_the_date( '', $post ) );
+		$formatted_post['title']        = get_the_title( $post );
+		$formatted_post['postLink']     = esc_url( get_permalink( $post ) );
+
+		$post_excerpt = $post->post_excerpt;
+
+		if ( ! ( $post_excerpt ) ) {
```
jrtashjian: We do not need to check if the post excerpt exists. The WordPress function [`get_the_excerpt`](https://developer.wordpress.org/reference/functions/get_the_excerpt/) will automatically generate an excerpt based on the content that exists.

src/blocks/posts/index.php
```diff
@@ -0,0 +1,417 @@
...
+				),
+				'displayPostLink'    => array(
+					'type'    => 'boolean',
+					'default' => false,
+				),
+				'postLink'           => array(
+					'type'    => 'string',
+					'default' => __( 'Read more', 'coblocks' ),
+				),
+				'excerptLength'      => array(
```
jrtashjian: Remove this attribute.

src/blocks/posts/index.php
```diff
@@ -0,0 +1,417 @@
...
+				esc_url( $post['date'] ),
+				esc_html( $post['dateReadable'] )
+			);
+
+		}
+
+		if ( isset( $attributes['displayPostContent'] ) && $attributes['displayPostContent'] ) {
+
+			$post_excerpt    = $post['postExcerpt'];
+			$trimmed_excerpt = esc_html( wp_trim_words( $post_excerpt, $attributes['excerptLength'], ' &hellip; ' ) );
```
jrtashjian: Do not trim words here. We should be receiving the excerpt through the use of [`get_the_excerpt`](https://developer.wordpress.org/reference/functions/get_the_excerpt/)

src/blocks/posts/inspector.js
richtabor: As this is a bit different from the standard blogroll on a WP site, I'd say it's ok to have our own manipulation here, as otherwise users would need to go in and adjust their filter (or add one) to do so. 

We're renaming this block to the "Posts" block, so maybe that'll make the difference a bit more pronounced, as this block is not intended to replace your blogroll - but to display blog posts.

PR #738, opened 2019-08-14
Add another container div in the Hero block

richtabor:
Maverick requires another wrapper div within the Hero block, to ensure the max-width of the content within the block does not exceed the set max-width within the theme. This does not affect how the Hero block displays for any other theme, but it does require a deprecation.


richtabor:
Deprecation works ✅

richtabor:
Here's a screenshot with the extra div added in this PR: 
Screen Shot 2019-08-14 at 9 58 18 AM

Before: 
Screen Shot 2019-08-14 at 9 59 19 AM

PR #740, opened 2019-08-14
Update Typography Controls Toolbar Icon

richtabor:
This PR updates the Typography Controls toolbar so that it properly displays in Gutenberg 6.2+.

Previous (_note that the typography icon square is wider than the corresponding alignment icon_): 
Screen Shot 2019-08-14 at 11 16 23 AM

Now (same widths): 
Screen Shot 2019-08-14 at 11 24 13 AM

I've also tested and confirmed this does not break core block editor as well.


jrtashjian:
Just need to fix that one eslint issue preventing the pass.

PR #749, opened 2019-08-19
Add advanced image options

Zrgisa:
Image cropping within the Gutenberg block for the core image block and core cover block.


src/extensions/advanced-image/index.js
```diff
@@ -0,0 +1,138 @@
+import AttachmentCropControl from '../../components/image-crop-control/attachment-crop-control';
+
+import { assign } from 'lodash';
```
paranoia1906: I changed this to import Syntax so ESLint was not not warning.

src/extensions/advanced-image/index.js
```diff
@@ -0,0 +1,138 @@
...
+
+		// Only Gallery images supported at this time
+		if ( ! currentAttributes.id ) {
+			return (
+				<BlockEdit { ...props } />
+			);
+		}
+
+		const updateImage = function() {
+			jQuery.post( global.ajaxurl, {
```
paranoia1906: Changed `ajaxurl` to `global.ajaxurl`. This is to avoid ESLint warnings.

paranoia1906:
Tested and seems to be working well. Made a few changes to minimize ESLint warnings.

Functionality is enabled only for `core/cover` and `core/image` blocks. I tried enabled for `coblocks/gallery-masonry` and other gallery blocks but it seems that this extension may need further modification before it may be used by CoBlocks Blocks.

richtabor:
Current flow: ScreenFlow

src/extensions/advanced-image/index.js
EvanHerman: For future reference, we can also throw `/* global ajaxurl */` at the top of the file to mark it as global in ESLint.

richtabor:
@Zrgisa I tweaked the loading a bit to be a bit more Gutenberg-like. It's _ok_ for now, but do you think it's possible to employ the core `Spinner` component when it's loading?

Here's a screenshot:
sss

PR #761, opened 2019-08-21
Add blockBackgroundColor attribute to Social Profiles and Share blocks

### LINKED ISSUE #759: ISBAT apply a background color to the social blocks
### Is this enhancement related to a problem? Please describe.
Right now we can only set colors for the social block's button and icon color. We should be able to set a background color for the entire block. One problem from a code point of view is that we are already using the `backgroundColor` name for the button colors.

### Describe the solution you'd like
We need to add a third color option for the block background. The simplest solution is to add a new `blockBackgroundColor` type with a label of "Background Color". The `backgroundColor` label should be updated to say "Button Color".

I'd like to eventually call these variables `backgroundColor` (instead of `blockBackgroundColor`), `buttonColor` and `iconColor` for the current values. I think it will be easier to make this transform happen when all three variables exist in the future.

### Additional context
Screenshot from 2019-08-20 15-27-34

paranoia1906:
- Closes #759 

Note: 
Added contrast checker for comparison of `blockBackgroundColor` and `backgroundColor`. ContrastChecker component is intended to be used for `textColor` compared to `backgroundColor` so the messaging references text. Perhaps we should remove the contract checker for the new attribute all together? What are your thoughts?

blockBackgroundColorforSocialandShare


src/blocks/share/inspector.js
```diff
@@ -61,10 +64,15 @@ class Inspector extends Component {
...
+			{
+				value: blockBackgroundColor.color,
+				onChange: setBlockBackgroundColor,
+				label: __( 'Block Background Color' ),
+			},
 			{
 				value: backgroundColor.color,
 				onChange: setBackgroundColor,
-				label: __( 'Background Color' ),
+				label: __( 'Button Background Color' ),
```
jrtashjian: This should be "Button Color" and "Text Color" should change depending on which style is selected: ("Icon Color", "Text Color", "Icon and Text Color")

src/blocks/social-profiles/inspector.js
```diff
@@ -73,10 +76,15 @@ class Inspector extends Component {
...
+			{
+				value: blockBackgroundColor.color,
+				onChange: setBlockBackgroundColor,
+				label: __( 'Block Background Color' ),
+			},
 			{
 				value: backgroundColor.color,
 				onChange: setBackgroundColor,
-				label: __( 'Background Color' ),
+				label: __( 'Button Background Color' ),
```
jrtashjian: Same as in the share block. This should be "Button Color" and "Text Color" should change depending on which style is selected: ("Icon Color", "Text Color", "Icon and Text Color")

src/blocks/share/inspector.js
paranoia1906: Adjusted share to use dynamic labels depending on icon style. Also removed contrast checker for block background.

src/blocks/social-profiles/inspector.js
paranoia1906: Adjusted social profiles to use dynamic labels depending on icon style. Also removed contrast checker for block background.

richtabor:
I'd say we could make the Block Background Color label just be "Background Color", as we already have a descriptive label for buttons specifically.

jrtashjian:
Good call.

paranoia1906:
Agreed. The labels are as follows: Background Color, Block Color, Text Color, and Icon Color. The labels for Text and Icon are dynamically applied depending on the style chosen.

richtabor:
I'd like to review this too. Thanks.

richtabor:
I found an issue when trying to set `textAlign1`. Happens with and without Gutenberg, and only on this PR.  ScreenFlow

paranoia1906:
Alignment bug is all sorted out.

richtabor:
@paranoia1906 Now it just needs the customBlockBackgroundColor applied inline (when you pick a custom color apart from the color palette).

paranoia1906:
@richtabor, I am not sure about the applying inline styles to the block. I think I must be missing something. Here is what I am seeing. 

**Selecting from pallete**
customBackgroundColorAppliedInline

**Pasting in a valid hex**
customBackgroundColorAppliedInlinePaste

richtabor:
> @richtabor, I am not sure about the applying inline styles to the block. I think I must be missing something. Here is what I am seeing.
> 
> **Selecting from pallete**
> customBackgroundColorAppliedInline
> 
> **Pasting in a valid hex**
> customBackgroundColorAppliedInlinePaste

Check the front-end. ;)

paranoia1906:
Those pesky bugs are squashed on this one.

PR #765, opened 2019-08-21
Add enhanced Flickity options to the Carousel block

richtabor:
This PR adds support for `alignCells`, `freeScroll`, and `pauseAutoPlayOnHover` to the Carousel block, which allows carousel items to display fully on the carousel (instead of the current view, where items are centered and may/may not show a portion of the other carousel items.

- [x] I've added a proper deprecation support for older versions of the carousel as well. 

**Here's the new feature in action:**
ScreenFlow


src/blocks/gallery-carousel/edit.js
```diff
@@ -229,6 +218,21 @@ class GalleryCarouselEdit extends Component {
 			);
 		}
 
+		const flickityOptions = {
+			draggable: false,
+			pageDots: true,
+			prevNextButtons: true,
+			wrapAround: true,
+			autoPlay: false,
+			cellAlign: alignCells ? 'left' : 'center',
```
jrtashjian: This doesn't seem to be re-positioning the cells to the left or center when the "Align Cells" slider setting is toggled in the editor. However, if you save the post and reload the editor it works. No problems on the theme side.

src/components/slider-panel/index.js
```diff
@@ -87,6 +92,12 @@ class SliderPanel extends Component {
 						onChange={ () => setAttributes( { pageDots: ! pageDots } ) }
 						help={ this.getDotNavigationHelp }
 					/>
+					<ToggleControl
+						label={ __( 'Align Cells' ) }
+						checked={ !! alignCells }
+						onChange={ () => setAttributes( { alignCells: ! alignCells } ) }
```
jrtashjian: If you pass the `this.flkty` as a prop down from `edit.js` down through the `Inspector` component and then into the `SliderPanel` component, you can use it in this onChange to set the `cellAlign` option of Flickity. This resolves the issue with it not updating in the editor but there are some display issues.

We need to find a way to re-render the Flickity component entirely but there is no obvious method to call. I also expect React would do this behind the scenes since properties are changing.

paranoia1906:
Related issue here: https://github.com/godaddy/coblocks/pull/705

src/blocks/gallery-carousel/edit.js
richtabor: Yea I think it’s acceptable as is for now.

jrtashjian:
Background color is not being set on the front-end:

Screen Shot 2019-09-18 at 11 19 40 AM
Screen Shot 2019-09-18 at 11 19 47 AM

richtabor:
> Background color is not being set on the front-end:

Found this also occurs in master, though I've applied the fix here as well. The deprecation catches this as well.

PR #768, opened 2019-08-22
Move getDividerFromStyle to a new utils.js file

jrtashjian:
Resolves an issue with multiple exports within the `index.js` file for the Shape Divider block. This was causing block validation errors in the editor and preventing the block from rendering on the front-end.


richtabor:
Looks good to me :)

PR #794, opened 2019-08-28
change icon size selctor from button with popout to a select element

### LINKED ISSUE #791: Change the Icon size dropdown to use select
**Describe the bug**
The Icon block utilizes the same styles and component layout as the now deprecated dropdown from the FontSizePicker from Gutenberg 6.3. 

Pull #741 resolved this issue for our DimensionsControl. We should implement the same type of resolution for the Icon block, using the select instead of the dropdown. 

**To Reproduce**

1. Install/Activate The Gutenberg plugin
2. Add the Icon block
3. View the Sidebar Settings panel

**Current UI from the Icon block's Sidebar Settings panel:**
Screen_Shot_2019-08-22_at_5 40 08_PM
Screen_Shot_2019-08-22_at_5 40 18_PM

**Expected behavior**

**Isolating the problem:**

**WordPress Version**
5.2.2

**Gutenberg Version**
6.3

paranoia1906:
- Closes #791 
- Does not require deprecation.
- Looks good **with** or **without** Gutenberg 6.4 installed.
- Note: CSS added to handle hidden screen reader label. This will not be needed when Gutenberg 6.4 is merged into WP core.

image


richtabor:
We're missing the "Advanced" button here, similar to the DimensionsControls component, which pulls up the RangeControl. You'll see the RangeControl, but only if you manipulate the size of the ResizableBox component within the block content area.

paranoia1906:
I see my mistake now. Had to add an additional class to get the render correct but this should be workable.

image

richtabor:
ScreenFlow

PR #806, opened 2019-08-29
Add simpler labels to coBlocksSpacing Advanced settings

richtabor:


richtabor:
> Consider if no margins have been applied to the block yet.

Margins are auto-applied automagically :)

PR #809, opened 2019-08-31
Fix missing innerBlocks template when adding the Hero block from inserter

### LINKED ISSUE #750: Hero inner blocks are missing when added from inserter
**Describe the bug**
The inner blocks of the Hero blocks are missing when the block is added using the block inserter (the dialog that shows up after clicking on the '+' icon). However, the inner blocks appear as expected when adding the block typing the `/` shorcut.

This issue is not reproducible when the Gutenberg plugin is not activated.

**To Reproduce**
1. Activate the latest version of the Gutenberg plugin.
1. Click on the "+" icon to open the block inserted.
1. Select the Hero block.
1. Notice how the inner blocks are missing.
1. Repeat the above steps, but this time insert the block by typing `/hero` on the post content.
1. Note how the inner blocks are correctly rendered now.

**Screenshots**
Aug-20-2019 11-54-46

**Expected behavior**
Inner blocks should appear after inserting a hero block.

**Isolating the problem:**

**WordPress Version**
5.2.0

**Gutenberg Version**
6.3.0

jrtashjian:
Resolves #750 

We are filtering the `coblocks/row`, `coblocks/column`, `coblocks/features`, `coblocks/feature`, `coblocks/media-card`, `coblocks/shape-divider`, and `coblocks/hero` blocks to add a `coblocks` attribute so that we can save a generated `id` property. This property is used to generate the inline css we save as post meta in the `saveMeta()` function of the DimensionControls` component.

There was an issue within the `withAttributes filter below which prevented the block from fully rendering with Gutenberg 6.4. This PR resolves that issue.

https://github.com/godaddy/coblocks/blob/98803bb4c84afa3979caec50ca87de41f0597417/src/extensions/attributes/index.js#L27-L62


src/extensions/attributes/index.js
```diff
@@ -30,36 +29,23 @@ function addAttributes( settings ) {
...
  */
-const withAttributes = createHigherOrderComponent( ( BlockEdit ) => {
-	return ( props ) => {
-		const {
-			attributes,
-			setAttributes,
-		} = props;
+const withAttributes = createHigherOrderComponent(
+	BlockEdit => props => {
+		const { name: blockName } = props;
```
jrtashjian: This bit is just a simple refactor of the nested callbacks to improve readability.

src/extensions/attributes/index.js
```diff
@@ -30,36 +29,23 @@ function addAttributes( settings ) {
...
 
-		if ( typeof attributes.coblocks === 'undefined' ) {
-			attributes.coblocks = [];
-		}
+		if ( allowedBlocks.includes( blockName ) ) {
+			props.attributes.coblocks = props.attributes.coblocks || {};
```
jrtashjian: We don't want every block to have a `coblocks` attribute so we only set it up if the current block is an allowed block.

src/extensions/attributes/index.js
```diff
@@ -30,36 +29,25 @@ function addAttributes( settings ) {
...
+		const { name: blockName } = props;
+
+		if ( allowedBlocks.includes( blockName ) ) {
+			props.attributes.coblocks = props.attributes.coblocks || {};
+
+			if ( typeof props.attributes.coblocks.id === 'undefined' ) {
+				const d = new Date();
+				props.attributes.coblocks = Object.assign( {}, props.attributes.coblocks, {
+					id: '' + d.getMonth() + d.getDate() + d.getHours() + d.getMinutes() + d.getSeconds() + d.getMilliseconds(),
+				} );
```
jrtashjian: Here we are generating a `coblocks.id` if one doesn't already exist.

src/extensions/attributes/index.js
```diff
@@ -30,36 +29,25 @@ function addAttributes( settings ) {
...
 		}
 
-		return (
-			<Fragment>
-				<BlockEdit { ...props } />
-			</Fragment>
-		);
-	};
-}, 'withAttributes' );
+		return <BlockEdit { ...props } />;
```
jrtashjian: We don't need to wrap the `BlockEdit` component in a `Fragment`, we can just return it.

src/blocks/hero/deprecated.js
```diff
@@ -6,6 +6,7 @@ import classnames from 'classnames';
 /**
  * Internal dependencies
  */
+import { attributes } from './block.json';
```
richtabor: I think we need to export attributes from index.js

jrtashjian: Having trouble importing the attributes being exported from `index.js`. Modified the code to export/import just like we're doing in the alert block but the imported attributes continue to be `undefined`. Not sure why yet but since this is only a minor cleanup we can figure this out later.

Leaving it as is so that we can keep moving forward on completing the 1.12.2 milestone.

PR #810, opened 2019-08-31
Add example block content 

richtabor:
With the new block inserter help panel in place and the example block content API layer in progress, we can now add neat little previews of our blocks (https://github.com/WordPress/gutenberg/issues/16979). 

The Gutenberg plugin is necessary to see the previews. 

This PRs takes a first stab at adding examples to _most_ of our blocks. The innerBlocks examples are still being fleshed out, but this PR focuses on all our other blocks. Once innerBlocks support makes it into the Gutenberg plugin, we should do a second pass to add tailored content to our nested blocks.

Gif:
ScreenFlow


src/blocks/gallery-carousel/index.js
```diff
@@ -39,6 +39,17 @@ const settings = {
 	supports: {
 		align: [ 'wide', 'full' ],
 	},
+	example: {
+		attributes: {
+			gridSize: 'lrg',
+			gutter: 5,
+			images: [
+				{ url: '/wp-content/plugins/coblocks/dist/images/examples/gallery-1.jpg' },
```
richtabor: One  bit I'm not 100% about is how I'm pulling images from the plugin to use in the examples for the gallery blocks. Thoughts?

jrtashjian: I think this is perfectly fine. We're not adding too many images and the file sizes are not large at all. I would prefer loading local images over potentially breaking the experience if a remote image fails to load.

src/blocks/author/index.js
```diff
@@ -28,6 +28,13 @@ const settings = {
 	description: __( 'Add an author biography to build credibility and authority.' ),
 	icon,
 	keywords: [ __( 'biography' ), __( 'profile' ), 'coblocks' ],
+	example: {
+		attributes: {
+			name: 'Jane Doe',
+			biography: __( 'Born to express, not to impress. A maker making the world I want.' ),
+			imgUrl: '/wp-content/plugins/coblocks/dist/images/examples/author.jpg',
```
mtias: This might break if the user has specified a different wp-content folder.

gziolo: Unless base URL is properly set in the head of HTML document.

jrtashjian: Good call. Looks like we need to expose the url produced through [`plugins_url()`](https://developer.wordpress.org/reference/functions/plugins_url/) in the admin to create the proper URL here.

gziolo: This is how webpack can handle public paths:

https://webpack.js.org/guides/public-path/

If you will find it useful, we can add it to wp-scripts build, too.

jrtashjian: I'm still planning to figure this out so we can server images locally. For now we're working on using an external source for now in #1106.

PR #812, opened 2019-09-01
Tweak block descriptions for Pricing & Social Profiles

richtabor:
This PR adds more helpful descriptions to our existing Pricing Tables and Social Profiles blocks.


richtabor:
> Wording changes look good to me. What do you think about wording the social profiles slightly differently?
> 
> > Build your brand with links to social media profiles.
> 
> Either way I approve of the changes.

I like that as well, though audience is a bit less specific I suppose. Good suggestion though. 👍

PR #823, opened 2019-09-03
Fix missing MediaFilterControl dropdown indicator

richtabor:
This PR resolves an issue where the filter control within our gallery blocks was missing a necessary dropdown indicator in Gutenberg 6.3+. 

**Before:** 
Screen Shot 2019-09-03 at 9 32 55 AM

**After:** 
Screen Shot 2019-09-03 at 9 32 31 AM


richtabor:
> What do you think about grouping the drop-down elements @richtabor?

Those others are aded via Gutenberg so we can't merge them :)

PR #763, opened 2019-08-21
Apply proper spacing for nested blocks

richtabor:
This PR adds proper styling to the Services, Pricing Table, Features, and Form blocks, emulating how the core Columns block handles the new padding/spacing system introduced in Gutenberg 6.3+. 

Closes #762 

Screenshots:
new
ScreenFlow
Screen Shot 2019-08-31 at 7 14 10 PM
Screen Shot 2019-08-31 at 7 39 14 PM

PR #825, opened 2019-09-03
Social share icons auto center

### LINKED ISSUE #771: ISBAT have the social icons centered automatically if I set wide/full alignment
When setting the social icon to display fullwidth/wide-width, the user then has to set the alignment of the buttons to `center` in most (if not all) instances of having wide/full alignment. 

I'm proposing UX where when the user selects wide or full width, the "textAlign" attribute is set to 'center' automagically - reducing that step.

paranoia1906:
- Closes #771.
- Block now defaults to `text-align: center;` when setting block to full page alignment.
- Social Profiles auto-aligns in editor.
  * Font-end render is dependent on merge of #763.
- Share block auto aligns in editor and on front-end.
  * Share uses inline styles instead of classes.
  * Social-Profiles block uses classes for alignment (`has-text-align-center`).

autoCenterSocialShare


jrtashjian:
Added center alignment when wide block alignment has been selected.

I also made a small change that I'd like an opinion on: When wide/full block alignment has been selected, we will only center the icons if a text alignment has not been set. So if the user selects right alignment of icons and then hits wide/full block alignment, the icons would not center.

**In Action:**

Kapture 2019-09-04 at 16 58 46

PR #827, opened 2019-09-04
Add Row block layout selector dropdown

### LINKED ISSUE #821: ISBAT change the Row block layout with a familiar dropdown interface
Not a problem per se, but our existing layout toolbar control within the Row block does not adhere to new Gutenberg patterns for toolbar control dropdowns. 

Our layout toolbar control uses the same functionality as the new Pricing Table controls in #811 , instead of its own control which emulates the core styles panel. We already have small icons we're using for the active state. 

**Additional context**
Toolbar control to be replaced: 
Screen Shot 2019-09-03 at 8 45 44 AM

Gif from the Pricing Tables PR, to replace our existing control in the Row block:

paranoia1906:
- Closes #821.
- Using the Toolbar controls similar to PR `enhance/pricing-tables-controls`.
- Made a minor change to icons smallLayout3366 and smallLayout6633 as they were mixed up.

**Row block layout drop-down**
rowLayoutDropdown

**33 / 66 layouts before change**
image

**33 / 66 layouts after change**
image


richtabor:
Looking good. Let's tweak the labels a bit to provide a bit more detail. 

Let's try something like this:

 `Two Columns; Equal Split`
 `Three Columns; Equal Split`
 `Three Columns; Two Thirds / One Third` 

..etc

It may be too long, but I'd like to see that sort of descriptive label there.

paranoia1906:
@richtabor I went ahead and have updated the labels used for the various layouts. The size of the labels does not seem to be much of an issue but it would be nice if all of the labels were equal width. Let me know if you would like me to update any of these labels. I am thinking the 4 column layout labels might need to be changed. Any suggestions?

image

image

image

paranoia1906:
I also just added support for i18n into the layouts file. I figure now that the labels will be used in the UI as opposed to programmatically that they should be translated. Please let me know if there is a more correct way to implement i18n here. Also I recall that additional contextual notes might be needed to explain the context of the translation. I would like clarification on that if it is needed here.

PR #834, opened 2019-09-06
Introduce a filter to prevent wp_mail from firing in the form block

EvanHerman:
Introduce a new filter (`coblocks_form_disable_emails`) which allows users to short circuit the form submission and prevent `wp_mail()` from firing.

Additionally introduced a new action (`coblocks_before_form_submit`) that allows users to catch the form data before the emails are disabled. The thought here is that users can disable the email functionality, but still hook in and capture/utilize the submitted form data as they need.

### Usage
```php
/**
 * Disable form submission emails
 */
add_filter( 'coblocks_form_disable_emails', '__return_true' );

/**
 * Set a custom success message to mimic a successful form submission
 *
 * @return string Form submission success message
 */
function coblocks_form_sent_message() {

	return __( 'Your message was sent.', 'textdomain' );

}
add_filter( 'coblocks_form_sent_notice', 'coblocks_form_sent_message' );
```

### DIFF
--- a/includes/class-coblocks-form.php
+++ b/includes/class-coblocks-form.php
@@ -426,7 +426,20 @@ class CoBlocks_Form {
 
 		$form_submission = filter_input( INPUT_POST, 'action', FILTER_SANITIZE_STRING );
 
-		if ( ! $form_submission || 'coblocks-form-submit' !== $form_submission ) {
+		/**
+		 * Filter to disable the CoBlocks form emails.
+		 *
+		 * @param bool false Whether or not the emails should be disabled.
+		 */
+		$disable_emails = (string) apply_filters( 'coblocks_form_disable_emails', false );
+
+		if ( ! $form_submission || 'coblocks-form-submit' !== $form_submission || $disable_emails ) {
+
+			if ( $disable_emails ) {
+
+				return true;
+
+			}
 
 			return;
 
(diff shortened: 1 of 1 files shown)


fjarrett:
@EvanHerman A better name for this filter might be `coblocks_form_email_send` that is `true` by default and you filter it to `false` to skip send.

includes/class-coblocks-form.php
```diff
@@ -453,6 +453,27 @@ public function process_form_submission( $atts ) {
...
+		 * @param array $atts  Form block attributes.
+		 */
+		do_action( 'coblocks_before_form_submit', $_POST, $atts );
+
+		/**
+		 * Filter to disable the CoBlocks form emails.
+		 *
+		 * @param bool false Whether or not the emails should be disabled.
+		 */
+		$disable_emails = (string) apply_filters( 'coblocks_form_disable_emails', false );
```
fjarrett: @EvanHerman I think this should be cast as a bool

includes/class-coblocks-form.php
```diff
@@ -453,6 +453,27 @@ public function process_form_submission( $atts ) {
...
+
+		if ( $disable_emails ) {
+
+			return true;
```
jrtashjian: The docblock for this function should reflect a boolean return value. And since we are returning `true` here, we should ensure a proper return value is set for the others rather than null (just to be explicit). Looks like the return of this function is used for a conditional so a boolean value makes sense.

EvanHerman: Good call.

PR #835, opened 2019-09-06
Adding the groundwork for testing block deprecation with Jest.

jrtashjian:
This PR explores how we can utilize Jest to test our block deprecations.

~**Before Merging**
We need to revert the changes added in 187b4cf. This was added only to demonstrate how we can write a test for a deprecated block.~

**Edit:** Reverted demo deprecation in 85038b3.

#### What this PR includes:
- Added tests for the Highlight block save function.
- Added tests for the Highlight block transforms.
- Added tests for the Hero block save function.
- Added tests for the Hero block deprecations.


jrtashjian:
Added `jsdom` and `jest-dom` npm dependencies so we can better test the HTML output with the familiar DOM. I've also cleaned up the tests and removed snapshots from all except the "it should render with content" test. We should probably add snapshots tests for each attribute that changes the HTML output of the block. This will ensure we know when a block would become invalid and need a deprecation in place.

## Test Example
Here's a simple test of the highlight block when a text alignment has been applied:

```
it( 'should center align text with inline css', () => {
    block.attributes.align = 'center';
    serializedBlock = serialize( block );

    blockDOM = new JSDOM( serializedBlock );
    expect(
        blockDOM.window.document.querySelector( '.wp-block-coblocks-highlight' )
    ).toHaveStyle( 'text-align: center' );
} );
```

### How it works

A few variables are declared at the top of the file that we will reuse throughout all of the tests:
```
// Make variables accessible for all tests.
let block;
let blockDOM;
let serializedBlock;
```

The block we're testing is registered before all tests are run
```
beforeAll( () => {
    // Register the block.
    registerBlockType( name, { category: 'common', ...settings } );
} );
```

The block is then created with the bare minimum attributes before each test is run. This bit also resets our reused variables so they are fresh at the beginning of each test
```
beforeEach( () => {
    // Create the block with the minimum attributes.
    block = createBlock( name, { content: 'highlighted content' } );

    // Reset the reused variables.
    blockDOM = undefined;
    serializedBlock = '';
} );
```

### Writing a test

In each test we can set attributes for a block by simply mutating the property
```
block.attributes.align = 'center';
```

Because we want to test the output of our save function we serialize the block object using the `serialize` function from `@wordpress/blocks` package. This will give us a string representation of the actual HTML (and block comment) being saved to the post_content. From there we can create a new DOM object using the `JSDOM` class from the `jsdom` package. This allows us to query for elements by class, id, etc and check properties just like in the browser.
```
serializedBlock = serialize( block );
blockDOM = new JSDOM( serializedBlock );
```

In order to complete our test, we need to check if the `.wp-block-coblocks-hightlight` element has an `text-align` inline style that matches the attribute we've set. The `jest-dom` package makes this nice and easy. We just query our DOM object for the element and check the style.
```
expect(
    blockDOM.window.document.querySelector( '.wp-block-coblocks-highlight' )
).toHaveStyle( 'text-align: center' );
```

### Running the tests

Running the tests is as easy as running `npm test` in your console. You should see something like this if everything went well.
```
 PASS  .dev/tests/jest/something.test.js
 PASS  src/extensions/colors/transform.spec.js
 PASS  src/extensions/colors/apply-style.spec.js
 PASS  src/extensions/typography/apply-style.spec.js
 PASS  src/extensions/colors/classes.spec.js
 PASS  src/utils/helper.spec.js
 PASS  src/blocks/highlight/test/index.spec.js

Test Suites: 7 passed, 7 total
Tests:       28 passed, 28 total
Snapshots:   1 passed, 1 total
Time:        4.246s
```

Currently I'm placing tests for a block within a `test/` subdirectory in a file named `index.spec.js`. Jest is configured to look for any JS files ending in `spec.js`, `test.js` or any JS file within the `test/` directory.

## Testing Deprecations

Testing for invalid blocks after making changes is the primary reason for this effort. We need to ensure our blocks provide the proper deprecations so that our blocks are never invalid after an update.

### How to test deprecated versions

The most important thing we need in order to test deprecations is the post_content output for the deprecated version of the block. You can see from https://github.com/godaddy/coblocks/commit/187b4cfd5d053085aec022dcb9639b095ef9f6e1#diff-87e53b8dafda4789e5cdc0d36740fa83 that I've added a deprecation to demonstrate. Here I've changed the tag being output by the `RichText.Content` component from a `<mark>` to a `<div>`.
```
const postContent = `
    <!-- wp:coblocks/highlight -->
    <p class="wp-block-coblocks-highlight"><mark class="wp-block-coblocks-highlight__content">highlighted content</mark></p>
    <!-- /wp:coblocks/highlight -->
`.trim();
```

Next we will pass `postContent` into the `parse` function from the `@wordpress/blocks` package. This function will return an array of block objects from the postContent. Since we are only passing a single block we only need to use the first index (`0`). The parse function will also determine if the block is valid and will return `false` if we need a deprecation (invalid) and `true` if it's valid.

```
const block = parse( postContent );
expect( block[ 0 ].isValid ).toBe( true );
```

## Next Steps

It's not enough to just test deprecations with the base html output. We need to ensure any attribute that modifies the resulting HTML is also tested. However, it's not easy to generate this static content automatically. So far we can add the block to the editor, make changes and then view the code editor and copy the post_content. An alternative is to console.log the `serializedBlock` output from our individual tests and store that. I'm not sure how to solve this problem well. We also need to think about the organizational structure for all of these deprecation tests.

That's all for now... doing a bit more exploration around these points. Overall, I think this is a great and lightweight solution for testing the output of our blocks and ensuring the validity of those blocks from version to version.

jrtashjian:
Added some tests for the Hero block which includes actual deprecations and removed the demo deprecation and related tests from the Highlight block. I've explored how test our deprecations in a simpler way as well as all the possible attribute variations. The original method of storing the serialized block object and then parsing it to test its validity was not easy to manage. I believe I've found a better way and will go over that now.

## Testing Deprecations Improved

The improved way of testing our deprecations requires us to register and the old version of the block which is possible because the `attributes` and `save` function must be saved for a deprecation to work (in our `deprecated.js` file). For the Hero block  I've explicitly selected the first index in the `deprecated` definition (only one exists right now).

### Register the old block

First we want to setup the block settings for the deprecated version of our block. We can do this by taking the current block settings and removing the `attributes`, `save`, and `deprecated` properties. Then we add our deprecated attributes and save properties to the block settings. We don't need the `deprecated` property because we don't want to check for any other fallbacks that may exist.

```
// Register the v1 block.
const v1Settings = Object.assign(
    {}, omit( settings, [ 'attributes', 'save', 'deprecated' ] ),
    {
        attributes: settings.deprecated[ 0 ].attributes,
        save: settings.deprecated[ 0 ].save,
    }
);
```

Then we simply register and create our block.

```
registerBlockType( name, { category: 'common', ...v1Settings } );
// Create the block.
v1Block = createBlock( name );
```

### Store the serialized block

You can modify block attributes through the `v1Block` object just like it's done in other tests. We want to serialize the block and store the output so that later we can test the deprecation.

```
const v1Serialized = serialize( v1Block );
```

### Unregister deprecated block, register current block

Once we have the serialized content we can unregister the deprecated version of our block.

```
// Unregister the v1 block version.
unregisterBlockType( name );

// Register the current block version.
registerBlockType( name, { category: 'common', ...settings } );
```

### Validate the deprecation

Now that we have the serialized block and emulated a block upgrade, we need to parse the serialized block content to run the deprecation functionality. The `parse()` function from the `@wordpress/blocks` package handles this for us. All we have to do is check that the `isValid` property is `true`. This tells us Gutenberg was successfully able to migrate the block to the newer version.

```
const blocks = parse( v1Serialized );
expect( blocks[ 0 ].isValid ).toBe( true );
```

## Testing Deprecated Block Variations

Because attribute values could change the serialized output of our blocks, we need a simple way to test these variations and verify our deprecations still work. The best way I've determined to do this is to create an object where each property matches to a block attribute and the values represent an array of possible block attribute values.

```
// Mapping of each attribute and value variation.
const variations = {
    contentAlign: [ undefined, 'left', 'center', 'right' ],
    textColor: [ undefined, 'primary' ],
    customTextColor: [ undefined, '#123456' ],
    maxWidth: [ 750 ],
};
```

Then we just loop over each attribute and each possible value and dynamically create a test for it.

```
Object.entries( variations ).map( ( [ attribute, variations ] ) => {
    variations.map( variation => {
        it( `should support attribute.${ attribute } set to '${ variation }'`, () => {
            // test the deprecation
        } );
    } );
} );
```

Within our test we would do the exact same we did earlier except we will modify one block attribute before serializing.

```
Object.entries( variations ).map( ( [ attribute, variations ] ) => {
    variations.map( variation => {
        it( `should support attribute.${ attribute } set to '${ variation }'`, () => {
            // Setup the deprecated block...

            v1Block.attributes[ attribute ] = variation;
            const v1Serialized = serialize( v1Block );

            // Setup the current block and test the deprecation...
        } );
    } );
} );
```

You can see everything together here:
https://github.com/godaddy/coblocks/blob/0d3ae77dda4e38f485f3ef76f9cdd7ddfd285877/src/blocks/hero/test/deprecated.spec.js

## Next Steps

There isn't much left to figure out here around the tests. However we do need to determine a new organizational structure for our deprecations. We need to know exactly what deprecation is being tested rather than relying on an array index that could change at any update. This may require creating additional files to house a single deprecation and naming that file in a way that's identifiable.

PR #836, opened 2019-09-07
Trigger display of the add accordion item button with isSelected

richtabor:
This PR changes the behavior of the "Add" accordion item button, making it only display when the parent block is selected. 

Why? Because in Gutenberg 6.3+, a block's child blocks are not selectable right at first - you have to click once per "nest" essentially. So right now, the "Add" button doesn't actually let you click it, unless you're already selected on the block.

I'd like to investigate if we can leverage `hasSelectedInnerBlock` ([source](https://github.com/WordPress/gutenberg/blob/1eca362e30aac95d7f0226ee8c080e7332b88c17/docs/designers-developers/developers/data/data-core-editor.md#hasSelectedInnerBlock)) to also display the Add button if you're editing inside a child Accordion Item block (and if that's available in WordPress 5.2 or now). 

This PR does this:
accordion


jrtashjian:
`hasSelectedInnerBlock` is only available in the Gutenberg plugin right now, but should be available in WP 5.3 in November. However, we did implement this functionality in the Food and Drinks block by finding the parent of the selected block. If this parent id was the same as the parent block's id, we would pass the prop down to hide/show our inserter.

Issue #685, opened 2019-07-23
Hero Block with background image covering other block's settings


retnonindya:
**Describe the bug**
When creating a new block after Hero Block, the settings/options on the next block is covered by Hero Block if the Hero Block is having background image.

**To Reproduce**
Create Hero Block and add Image Background for the Hero Block.

![image](https://cldup.com/9bF817z1Vx.png)
*Direct link: https://cldup.com/9bF817z1Vx.png*

Create new block under Hero Block.

Click on the Block and try to edit or change the block. It's being covered by Hero Block above it.

![new block](https://cldup.com/db1FlowC0E.png)
*Direct link: https://cldup.com/db1FlowC0E.png*

The block's settings won't be obstructed/covered when the Hero Block is not using any image background.

![not covered](https://cldup.com/Mutw9g7CKb.png)
*Direct link: https://cldup.com/Mutw9g7CKb.png*

**Expected behavior**
I can still see the block's settings when the block is clicked.

**Isolating the problem:**
Mark completed items with an [x].

-   [x] This bug happens with no other plugins activated. This is a Simple site (Premium plan)
-   [x] This bug happens with a default WordPress theme active. I use Independent Publisher 2 theme
-   [x] This bug happens **without** the Gutenberg plugin active
-   [x] I can reproduce this bug consistently using the steps above

**WordPress Version**
Since the site is Simple, it would be the latest version (5.1.1, I presume.)

richtabor:
I've confirmed the issue. I believe adding a `z-index: 0` to `.wp-block[data-type="coblocks/hero"]` should resolve this.

retnonindya:
5 days late of me, but thank you so much, @richtabor! Thank you!

PR #838, opened 2019-09-09
Change z-index to 60 for hero block controls.

### LINKED ISSUE #826: Hero Block: Unable to use toolbar when Hero block is below Group block
**Describe the bug**
When there's Group block above the Hero block, it is not possible to click on the Hero block toolbar. One of the newer theme needs to be activated, e.g. Twenty Nineteen.

It seems the issue could be related to the following code:

```
.wp-block[data-type="coblocks/hero"][data-align="full"], .wp-block[data-type="coblocks/hero"][data-align="wide"] {
    z-index: 0;
}
```

in `/wp-content/plugins/coblocks/dist/blocks.editor.build.css`

**To Reproduce**

1. Make sure Gutenberg and CoBlocks plugins are installed and activated.
2. Activate one of the newer themes, e.g. Twenty Nineteen (it doesn't have to be this one).
3. Create an empty page and add Group block to it.
4. Right below add the Hero block.
5. Try to use the Hero block toolbar.

**Screenshots**

Screen Capture on 2019-09-04 at 13-53-35

**Expected behavior**
It is possible to use Hero block toolbar.

**Isolating the problem:**

-   [ ] This bug happens with no other plugins activated

**WordPress Version**
5.2.2

**Gutenberg Version**
6.4.0

paranoia1906:
- Closes #826.
- I found that z-index of 60 was the minimum index that works in this case. 
- Tested block as user and found no negative side effect. 

heroToolbarBelowGroup

### DIFF
--- a/src/blocks/hero/styles/editor.scss
+++ b/src/blocks/hero/styles/editor.scss
@@ -17,7 +17,7 @@
 
 	&[data-align="full"],
 	&[data-align="wide"] {
-		z-index: 0;
+		z-index: 60;
 	}
 
 	.wp-block-coblocks-hero {


jrtashjian:
This fix is valid so I'm approving it.

I did a little sleuthing to figure out why we are setting `z-index`'s. Modifying the layering of elements can get messy which was a flag for me. I found that Gutenberg is actually modifying the z-index of the toolbar in these two places:

1. https://github.com/WordPress/gutenberg/blob/master/packages/block-editor/src/components/block-list/style.scss#L496
2. https://github.com/WordPress/gutenberg/blob/master/packages/block-editor/src/components/block-list/style.scss#L396

This was added in order to prevent the toolbar from being under an adjacent floated block, specifically the `core/image` block. Disabling this z-index and our own z-index resolved this issue and the original issue for why this exists. It would also resolve the issue where our z-index originated from in #685. 

Just noting this for future reference and a potential contribution back into the Gutenberg project.

PR #839, opened 2019-09-09
Add thumbnail and responsive height controls to the Carousel block

grabovacnem:


richtabor:
Hey @grabovacnem, it looks like we're missing the controls within `InspectorControls` to toggle the new functionality. 

Screenshot:
Screen Shot 2019-09-17 at 5 15 00 PM

richtabor:
Resolved potential deprecation issues by including the bits that would require a deprecation into the deprecation we are adding to the 1.14 release with #904.

src/blocks/gallery-carousel/edit.js
```diff
@@ -327,6 +365,32 @@ class GalleryCarouselEdit extends Component {
 						</div>
 					</div>
 				</ResizableBox>
+				<div className={ className }>
+					<div
+						className={ innerClasses }
+						style={ innerStyles }
+						style={ navStyles }
```
jrtashjian: I'm unsure which style var we should be using here. ESLint threw an error about duplicate props here.

src/blocks/gallery-carousel/edit.js
```diff
@@ -327,6 +365,32 @@ class GalleryCarouselEdit extends Component {
 						</div>
 					</div>
 				</ResizableBox>
+				<div className={ className }>
```
jrtashjian: We shouldn't be re-using the main block className here, this className needs to represent the element.

src/blocks/gallery-masonry/index.js
```diff
@@ -35,7 +35,7 @@ const settings = {
 	title: _x( 'Masonry', 'block name' ),
 	description: __( 'Display multiple images in an organized masonry gallery.' ),
 	icon,
-	keywords: [	_x( 'gallery', 'block keyword' ), _x( 'photos', 'block keyword' )	],
+	keywords: [ _x( 'gallery', 'block keyword' ), _x( 'photos', 'block keyword' ), _x( 'lightbox', 'block keyword' ) ],
```
jrtashjian: This change is not related to the PR and instead looks like it should be moved into PR #840.

src/blocks/gallery-stacked/index.js
```diff
@@ -35,7 +35,7 @@ const settings = {
 	title: _x( 'Stacked', 'block name' ),
 	description: __( 'Display multiple images in an single column stacked gallery.' ),
 	icon,
-	keywords: [	_x( 'gallery', 'block keyword' ), _x( 'photos', 'block keyword' ) ],
+	keywords: [	_x( 'gallery', 'block keyword' ), _x( 'photos', 'block keyword' ), _x( 'lightbox', 'block keyword' ) ],
```
jrtashjian: This change is not related to the PR and instead looks like it should be moved into PR #840.

src/blocks/gallery-carousel/edit.js
richtabor: > We shouldn't be re-using the main block className here, this className needs to represent the element.

This is used to not have to duplicate much of the carousel styling. It's only within the editor.

PR #840, opened 2019-09-09
Add lightbox support to masonry and stacked galleries

grabovacnem:


richtabor:
Looking good, just a few additional thoughts we should look into: 

1. Is is possible to pull a higher res image from the src if available? For example, themes usually have `srcset` support, where higher images are served at specific resolutions/viewports. When the image is pulled for the gallery, it's the src, which is usually sized down a bit. 

It'd be great if we could use a better sized image here, perhaps even the original upload, instead of forcing a low quality image. 

Here's from the TwentyNineteen theme:

```
<img src="https://coblocks.test/wp-content/uploads/2019/09/mae-mu-Rz5o0osnN6Q-unsplash-819x1024.jpg" alt="" data-id="5103" data-link="https://coblocks.test/?attachment_id=5103" class="wp-image-5103" srcset="https://coblocks.test/wp-content/uploads/2019/09/mae-mu-Rz5o0osnN6Q-unsplash-819x1024.jpg 819w, https://coblocks.test/wp-content/uploads/2019/09/mae-mu-Rz5o0osnN6Q-unsplash-240x300.jpg 240w, https://coblocks.test/wp-content/uploads/2019/09/mae-mu-Rz5o0osnN6Q-unsplash-768x960.jpg 768w, https://coblocks.test/wp-content/uploads/2019/09/mae-mu-Rz5o0osnN6Q-unsplash-1568x1960.jpg 1568w" sizes="(max-width: 819px) 100vw, 819px">
```
 Screen Shot 2019-09-18 at 7 16 24 AM

2. Let's move the lighbox markup to before the closing `</body>` tag, instead of right at the top. 

3. Does the markup have to be present on the page if Lightbox is not enabled? Perhaps the JS could look for the `.coblocks-gallery.has-lightbox` class, and if found render the markup.

richtabor:
@grabovacnem Looking good! One last ask. I added lightbox support to the Stacked block, though I'm getting a console error on the front-end. Mind looking into that?

src/blocks/gallery-masonry/save.js
```diff
@@ -76,7 +78,7 @@ const save = ( { attributes, className } ) => {
 							href = image.imgLink;
 						}
 
-						const img = <img src={ image.url } alt={ image.alt } data-id={ image.id } data-imglink={ image.imgLink } data-link={ image.link } className={ image.id ? `wp-image-${ image.id }` : null } />;
+						const img = <img src={ image.url } alt={ image.alt } data-id={ image.id } data-link={ image.link } data-index={ index } data-imglink={ image.imgLink } className={ image.id ? `wp-image-${ image.id }` : null } />;
```
richtabor: We're going to need a deprecation for this here.

richtabor:
We're all good, just need deprecations for stacked and carousel blocks

richtabor:
> We're all good, just need deprecations for stacked and carousel blocks

Deprecations for both Masonry and Stacked are done. Stacked also has lightbox support now.

richtabor:
@grabovacnem This the only thing left. Looking into how we could use the original image in the lightbox, instead of the cropped image `src`. Any ideas?

> 1. Is is possible to pull a higher res image from the src if available? For example, themes usually have `srcset` support, where higher images are served at specific resolutions/viewports. When the image is pulled for the gallery, it's the src, which is usually sized down a bit.
> 
> It'd be great if we could use a better sized image here, perhaps even the original upload, instead of forcing a low quality image. If we're using the image from the `data-url` attribute, could we make that be the unsized/original image?
> 
> From the TwentyNineteen theme:
> 
> ```
> 
> ```
> 
> Screen Shot 2019-09-18 at 7 16 24 AM

src/js/coblocks-lightbox.js
```diff
@@ -0,0 +1,72 @@
...
+	$( '.has-lightbox ul li figure img' ).click( function( e ) {
+		$.each( images, function( i, img ) {
+			imagePreloader[ 'img-' + i ] = new window.Image();
+			imagePreloader[ 'img-' + i ].src = img.attributes[ 0 ].value;
+		} );
+
+		index = Number( e.target.attributes[ 4 ].value );
+		wrapper.css( 'display', 'flex' );
+		wrapperBackground.css( 'background-image', 'url(' + imagePreloader[ 'img-' + index ].src + ')' );
+		image.attr( 'src', imagePreloader[ 'img-' + index ].src );
```
jrtashjian: Using the index in order to select images to load is not taking into account multiple galleries on a single page nor the group of images in a specific gallery on a page.

If we have 2 galleries on a page, each with 4 images, we will have a single index assigned to two different images. This allows only the first image with that index to be shown in the lightbox. SO if a user selected the first image on the second gallery, the first image of the first gallery would be lightboxed.

src/blocks/gallery-masonry/save.js
```diff
@@ -76,7 +78,7 @@ const save = ( { attributes, className } ) => {
 							href = image.imgLink;
 						}
 
-						const img = <img src={ image.url } alt={ image.alt } data-id={ image.id } data-imglink={ image.imgLink } data-link={ image.link } className={ image.id ? `wp-image-${ image.id }` : null } />;
+						const img = <img src={ image.url } alt={ image.alt } data-id={ image.id } data-link={ image.link } data-imglink={ image.imgLink } data-index={ index } className={ image.id ? `wp-image-${ image.id }` : null } />;
```
jrtashjian: We shouldn't be utilizing the index of a loop here as there could be more than one gallery on a page.

src/blocks/gallery-stacked/save.js
```diff
@@ -93,7 +99,7 @@ const save = ( { attributes, className } ) => {
 							[ `has-shadow-${ shadow }` ]: shadow !== 'none' || shadow !== undefined,
 						} );
 
-					const img = <img src={ image.url } alt={ image.alt } data-id={ image.id } data-imglink={ image.imgLink } data-link={ image.link } className={ imgClasses } />;
+					const img = <img src={ image.url } alt={ image.alt } data-id={ image.id } data-imglink={ image.imgLink } data-link={ image.link } data-index={ index } className={ imgClasses } />;
```
jrtashjian: We shouldn't be utilizing the index of a loop here as there could be more than one gallery on a page.

jrtashjian:
Screen capture of the image indexing issue. Also take note of the top left numbers. Each gallery has 2 images and I would expect those numbers to be #/2, not #/4

Kapture 2019-09-20 at 16 42 49

src/js/coblocks-lightbox.js
grabovacnem: I have adjusted the script, you can check it now, should be working good with multiple galleries @jrtashjian

richtabor:
Looks and works great!

PR #877, opened 2019-09-12
Add waves masks on images feature

### LINKED ISSUE #842: ISBAT apply a waves mask to images
As a customer, I should be able to easily add a stylistic element to my pages by applying a waves mask to images on my page.

Experimental branch on GitHub: `bottom-waves-image-style`

AC:
1. Add two styles to the core Image block, which allows for a waves SVG mask applied to either the top or bottom of the image within the editor and the frontend.

2. This should work fine in TwentyNineteen as well.

Experiment Video: https://d.pr/free/v/gM07c6

paranoia1906:
- Closes #842.
- Added waves style for bottom and top of images.
- CSS is responsive and it tested and working well in Chrome, FireFox, Edge, and using mobile device emulators built into FireFox and Chrome. 
- Tested using Twentynineteen, Twentyseventeen, Twentysixteen, and underscores bare bones theme.
- Future releases could feature options for SVG horizontal flip, custom colors, and different styles of SVG masks. 

I was not able to figure out how to get the IE11 support working here. I was able to get a background image applied but the layering was not working as expected.

There are two bugs that I am still working on tracking down. They are both related to the image-mask **not** being 'contained'

When first inserted
imageMaskUncontainedNew

When switching page alignment
imageMaskUncontained


jrtashjian:
@paranoia1906 great work! I'm about to push a few updates that build upon what you've accomplished. We can actually use the `mask` css properties directly on the image which removes the need for the psuedo element being added and needing to match the width of the image. This solves all the issues you had above and also removes the need for us to hide the caption text.

As for IE11 support, `mask` is not a supported property and I think in this case it's ok not to have it working. I don't think it would be beneficial to add support for it either. If the browser doesn't support the `mask` property, it doesn't break the content and the image renders normally.

The major difference is the SVG we're using as the mask. When an SVG is used as a mask, the path added is actually the part of the image that is removed. So in order to get this working properly I made an SVG which is the reverse of the shape. So in your example above the white portion is the SVG shape whereas the new shape would be the red portion.

jrtashjian:
One final update: fixing the style preview for that bottom wave style. I'm hiding the caption in the preview and aligning the block so we can see the bottom wave.

Screen Shot 2019-09-17 at 1 34 30 PM

jrtashjian:
Preview:

Kapture 2019-09-17 at 13 41 58

src/extensions/image-styles/index.js
```diff
@@ -0,0 +1,30 @@
...
+registerBlockStyle( 'core/image', {
+	name: 'default',
+	label: __( 'Default' ),
+	isDefault: true,
+} );
+
+registerBlockStyle( 'core/image', {
+	name: 'bottom-wave',
+	label: __( 'Bottom Wave' ),
+	isDefault: false,
```
jrtashjian: We only need to set isDefault to the default image style only.

richtabor:
I think it's fine that captions display in the preview (as core allows it). I also removed some extraneous styling that was causing some display errors in the sidebar and aren't really needed.

kosiew:
The wave style is broken.

See http://xizunite.com/test-image/

Screenshot
*if image does not load, please click on this link  : https://d.pr/i/eGnc83*

PR #885, opened 2019-09-17
Fix display bug for images within Features block.

### LINKED ISSUE #848: ISBAT align images with Features inner blocks
**Describe the bug**
- Bug with 2 columns only effects FireFox. Tested with version 69.0
- Bug with 1 column effects all browsers. 

When using an image within the features block content is centered within the editor. On the front of site the image is not properly aligned with the inner blocks. 

**To Reproduce**

1. Add Features block
2. Swap paragraph block for image by typing /image.
3. Set image from gallery and publish.
4. View page to see error.

**Screenshots**
**Within the editor 1 column**
image

**On the front of site 1 column**
image

**Within the editor 2 column**
image

**On the front of site 2 columns**
image

**3 Columns renders correctly inside editor and on front of site.** 
image

**Expected behavior**
All inner blocks to features should align the same way. 

**Isolating the problem:**

**WordPress Version**
5.2.3

**Gutenberg Version**
N/A

paranoia1906:
- Closes #848.
- This change resolves the display bugs with single and double columns.

**Screenshot with fix.**
image


src/blocks/features/styles/style.scss
```diff
@@ -88,6 +88,10 @@
 	}
 
 	.wp-block-coblocks-feature__inner {
+		img {
+			display: inline !important;
```
jrtashjian: We should avoid using !important as much as possible.

PR #887, opened 2019-09-17
Handle resizable box display in style preview

paranoia1906:
This is a fix for the style preview of blocks after adjusting the resizable box. The screen captures below illustrates the issue. Note: We may want to re-locate this CSS to another file or location.

**Before Fix**
imageMaskPreviewBug

**After Fix**
imageMaskPreviewFix


richtabor:
This only affects pre Gutenberg 6.3. We can remove this once WordPress 5.3 is released.

PR #894, opened 2019-09-18
Save custom padding and margin into post content Row block

### LINKED ISSUE #893: ISBAT apply custom margin and padding to Row block
**Describe the bug**
When using dimensions controls for the Row block you are able to go into 'Advanced' options and set values for horizontal and vertical padding and margins. At this time CoBlocks is not saving content including custom margins applied inline as is shown within the editor. This will likely require a deprecation of the current save function and a change in the save function properly apply inline styles.

**Applied in editor**
https://github.com/godaddy/coblocks/blob/ee6ba42ba793528c9b0b65c50c3ce445968144b0/src/blocks/row/edit.js#L303

**Not applied in save function**
https://github.com/godaddy/coblocks/blob/ee6ba42ba793528c9b0b65c50c3ce445968144b0/src/blocks/row/save.js#L64

**To Reproduce**

1. Add a row block and insert blocks into columns. 
2. Select the row block and apply custom values for margins or padding.
3. Notice that changes are visible within the editor.
4. Publish changes and view page to see that custom values are not saved to front-end.

**Screenshots**
**Display within the editor**
image

**Front of site**
image

**Expected behavior**
Expect custom CSS values to be applied both in editor and on front-of-site.

**Isolating the problem:**

**WordPress Version**
5.2.3

**Gutenberg Version**
6.5 (optional)

**Inspired by WordPress.org report**
https://wordpress.org/support/topic/margins-on-columns/

paranoia1906:
- Closes #893 
- Adjust save function to apply inline styles when advanced margins or padding are used.
- Create deprecated save function to handle old block markup.

### DIFF
--- a/src/blocks/row/deprecated.js
+++ b/src/blocks/row/deprecated.js
@@ -2,7 +2,8 @@
  * Internal dependencies
  */
 import { BackgroundClasses, BackgroundVideo } from '../../components/background';
-import metadata from './block.json';
+import { attributes } from './block.json';
+import DimensionsAttributes from '../../components/dimensions-control/attributes';
 
 /**
  * External dependencies
@@ -32,7 +33,10 @@ function Deprecation( { attributes } ) {
 	} = attributes;
 
 	const textClass = getColorClassName( 'color', textColor );
-	const backgroundClass = getColorClassName( 'background-color', backgroundColor );
+	const backgroundClass = getColorClassName(
+		'background-color',
+		backgroundColor
+	);
 
 	let classlist = {
 		[ `coblocks-row--${ id }` ]: id,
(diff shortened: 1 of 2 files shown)


jrtashjian:
Pushed fixes around the deprecations here which resolve the `id` errors I was receiving.

@richtabor can you test this out for sanity?

paranoia1906:
Pushed a change into this branch. Preferred setting key value of the `classlist` object using bracket notation instead of Object.assign. This allowed proper use of `classnames` function. Now ID is set correctly and the classes are properly calculated. This resolved the outstanding deprecation bugs that I could find. 

Tested against master and deprecates with no issues.

jrtashjian:
We're going to punt this one so we can do some more testing on it before merging.

richtabor:
While this PR does add the inline styles for margin/padding, it's not quite the correct solution for the original issue. The styles should not be inline, but generated and targeted via the row block's particular class, i.e. `coblocks-row-9494758173` (as they originally were [here](https://github.com/godaddy/coblocks/blob/master/includes/class-coblocks-generated-styles.php#L59)). 

It seems there's a bug in _that_ functionality - inline styles were never added to the Row block's save function (as that wouldn't work with mobile/tablet settings).

richtabor:
Does this PR address this same issue? https://github.com/godaddy/coblocks/pull/543

paranoia1906:
Pull #543 has the correct solution to this problem. Closing this pull request in favor of #543.

PR #929, opened 2019-10-01
Add auto height and max-width to masonry block images

### LINKED ISSUE #837: Masonry: Incorrect height on site but correct in WP
## Prerequisites

-   [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate
-   [x] The issue still exists against the latest `master` branch of CoBlocks on Github (this is **not** the same version as on WordPress.org!)
-   [x] I have attempted to find the simplest possible steps to reproduce the issue

## Steps to reproduce the issue

1. Add the Masonry gallery block in a page.
2. Add images and save.

Am currently working in a local dev env so unfortunately I can't link to a page with the error.

## Expected/actual behavior

Screenshot 2019-09-07 at 10 10 03

## Isolating the problem

-   [x] This bug happens with only CoBlocks plugin active
-   [ ] This bug happens with a default WordPress theme active **(Not in Twenty Nineteen)**

I have narrowed down the issue to these lines of CSS:
```scss
.wp-block-coblocks-gallery-masonry {
    img {
        max-width: 100%;
        height: auto;
    }
}
```
If I simply add these img styles the problem is fixed. These lines are also present in the Twenty Nineteen theme.

## WordPress Environment

The issue is present for me in Chrome (76.0.3809.132) and Safari (12.1.2), and the issue is present in a bare bones theme with no theme css. The issue is NOT present in a standard WP theme (Twenty Nineteen).

```
Wordpress 5.2.3
Coblocks 1.13.0
```

richtabor:
This PR applies suggested styling to resolve #837.

### DIFF
--- a/src/blocks/gallery-masonry/styles/style.scss
+++ b/src/blocks/gallery-masonry/styles/style.scss
@@ -15,6 +15,8 @@
 	}
 
 	img {
+		height: auto;
+		max-width: 100%;
 		vertical-align: bottom;
 	}
 
--- a/src/blocks/gallery-masonry/styles/style.scss
+++ b/src/blocks/gallery-masonry/styles/style.scss
@@ -16,7 +16,7 @@
 
 	img {
 		height: auto;
-		max-width: 100%;
+		width: 100%;
 		vertical-align: bottom;
 	}
 


richtabor:
> Made a change here to set the width to 100% instead of setting the max width. I found this gives me a correct display in bare-bones theme, twentynineteen, twentyseventeen, twentysixteen, and soho. See what you think. @richtabor

Trying to decide if we should force images to display at 100% here. Thinking about it.

richtabor:
Yea this is fine. 🤘

Issue #850, opened 2019-09-10
ISBAT set text alignments without the use of inline styling


richtabor:
**Is this enhancement related to a problem? Please describe.**
Inline styles should be a last resort when applying styling to blocks. When WordPress 5.3 is released, Gutenberg will be updated to use utility classes, such as `.has-text-align-center` to apply alignments, so we should follow suite. 

Here are blocks that we should tweak and provide proper deprecation support for: 

- [x] Alert
- [x] ~~Author (is not necessary, as the #808 PR removes this capability anyhow)~~
- [x] Click to Tweet
- [x] Pricing Table
- [x] Row
- [x] Social Profiles (#931)
- [x] Share 
- [x] Icon

**Describe the solution you'd like**
Each inline text align style is replaced with `.has-text-align-{value}` classes.

jrtashjian:
Just a future reminder that we should create multiple PRs for each block that's changed.

jrtashjian:
All tasks completed.

PR #932, opened 2019-10-02
Convert Icon block to use alignment classes

richtabor:
This PR finishes up the work for #850, which replaces the inline style alignments with `.has-text-align-{value}`. Deprecation has been added to the Icon block.


richtabor:
> The edit.js file is still setting the text alignment via inline styles. Other pull requests related to #850 feature the change in both save and edit components. Should we consider the modification here for code continuity?

Added, thanks :)

PR #938, opened 2019-10-02
Add Reply-To headers to Form block

### LINKED ISSUE #916: ISBAT reply to Form block submissions
When a contact form is submitted by a user, the message shows as being sent from wordpress@domain.tld. If I attempt to reply to the email message, the to address is set to wordpress@domain.tld as opposed to the address of the person who originally submitted. 

We are able to improve this process by adding a 'reply-to' header to the sent email. This will enable a user to reply to the original form submitter without having to manually enter the address.

At this time a website owner would have to copy the address over the the 'to' field when replying.

**Inspired by** 
https://wordpress.org/support/topic/contact-form-252/

paranoia1906:
- Closes #916

### DIFF
--- a/includes/class-coblocks-form.php
+++ b/includes/class-coblocks-form.php
@@ -28,6 +28,13 @@ class CoBlocks_Form {
 	private $email_content;
 
 	/**
+	 * Address to use in the Reply-To message header
+	 *
+	 * @var string
+	 */
+	private $reply_to_email;
+
+	/**
 	 * Form hash
 	 *
 	 * @var string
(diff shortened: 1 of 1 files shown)


includes/class-coblocks-form.php
```diff
@@ -508,6 +515,9 @@ public function process_form_submission( $atts ) {
 
 			$this->email_content .= '<li>' . sanitize_text_field( $data['label'] ) . ': ' . sanitize_text_field( $data['value'] ) . '</li>';
 
+			if ( 'Email' === $data['label'] ) {
```
richtabor: This shouldn't rely on the label being "Email", as users can change that value.

EvanHerman:
Closing in favor of https://github.com/godaddy/coblocks/pull/953

PR #939, opened 2019-10-02
Remove allowedBlocks limitation from the Accordion Item

### LINKED ISSUE #888: ISBAT add reusable blocks to Accordion block
At this time there are a limited set of blocks allowed within an Accordion block. This is a proposal to open up options to allow reusable blocks within an Accordion block.

Ability to add reusable blocks to Accordion block.

We may also want to consider allowing others outside of reusable Row, Alert, Image, Share, Highlight, Button, Heading, List, or Columns blocks within accordion and perhaps allow most if not all blocks. 

**Additional context**
**Here is the interface for the row block.**
image

**Whereas the Accordion block only allows a few.**
image

**Inspired by WordPress.org support request**
https://wordpress.org/support/topic/accordion-doesnt-support-reuseable-blocks/

richtabor:
This PR fixes #888, which was caused by limited the kind of blocks that could be placed within an Accordion Item block. The limitation has been lifted and I added a slight style tweak to ensure that fullwidth images don't flow out of the bounds of the Accordion Item.

### DIFF
--- a/src/blocks/accordion/accordion-item/edit.js
+++ b/src/blocks/accordion/accordion-item/edit.js
@@ -21,8 +21,6 @@ const { InnerBlocks, RichText } = wp.blockEditor;
 /**
  * Constants
  */
-const ALLOWED_BLOCKS = [ 'core/button', 'core/paragraph', 'core/heading', 'core/list', 'core/image', 'core/columns', 'core/column', 'coblocks/row', 'coblocks/column', 'coblocks/highlight', 'coblocks/alert', 'coblocks/social' ];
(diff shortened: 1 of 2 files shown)


richtabor:
@paranoia1906 I thought I commented previously - but we can't account for every block's wide/full alignment - nor how the theme decides to treat those. For example, 2019 actually makes wide blocks go wider, but pull to the left. This affects how the block sits within the accordion. 

By opening the accordion to all sorts of blocks, we have to accept these potential hang-ups.

PR #949, opened 2019-10-07
Tag each release of CoBlocks on the WordPress.org SVN during deployment

### LINKED ISSUE #905: Composer and WPackagist compatibility
Hey guys, can you create tags into your WordPress.org SVN repository so WPackagist will show up all available versions?

I think that makes sense create a composer.json file within the plugin too.

Currently it's only possible install the latest version using Composer.
https://wpackagist.org/search?q=coblocks&type=plugin

My team and I started using Coblocks 1.13.x via Composer in the last week and this version is not available anymore in WPackagist so it may break any Composer-based workflow.

Let me know if there's anything else I can help you.

EvanHerman:
Resolves #905 

Create a tag directory on WordPress.org SVN during plugin deployments.

### DIFF
--- a/.dev/bin/deploy-plugin.sh
+++ b/.dev/bin/deploy-plugin.sh
@@ -7,6 +7,10 @@ svn co "http://svn.wp-plugins.org/coblocks" $HOME/coblocks
 rm -rf $HOME/coblocks/trunk/*
 cp -a build/coblocks/* $HOME/coblocks/trunk/
 
+# Create the tag on the SVN repo and copy over plugin files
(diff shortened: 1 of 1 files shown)


jrtashjian:
This definitely works but I believe using `svn copy` would be quicker. In this case we would be uploading the files twice to SVN whereas `svn copy` does the copy remotely from trunk.

```
# Instead of this:
mkdir $HOME/coblocks/tags/${CIRCLE_TAG}
cp -a build/coblocks/* $HOME/coblocks/tags/${CIRCLE_TAG}/

# We can do this:
svn cp trunk tags/${CIRCLE_TAG}
```

PR #958, opened 2019-10-07
Add a 'popular' menu item icon to Food and Drinks block

### LINKED ISSUE #902: ISBAT add a popular food item to my site's food menu
Many restaurants have popular or favorite dishes which may sell more then others. Perhaps its a dish that makes the restaurant famous. We should allow site owners to mark select food items as popular with a block attribute. 

Add checkbox to block sidebar to allow a mechanism indicating a popular dish.

At this time one would need to style a Food & Drinks block separately to get a special look for a single food item or a set of food items. 

**Additional context**
Here is a quick mock-up which shows where interface controls are expected.
image

**SVG Icon to use when marking food items poular**
[popular.zip](https://github.com/godaddy/coblocks/files/3677698/popular.zip)

paranoia1906:
Closes #902 

Question for you @jrtashjian Should the block have the `Popular` option display within the editor when selected similar to how spicy works and how this screenshot shows? OR should the icon be hidden until selected from the Sidebar like how Vegetarian is treated?

image


src/blocks/food-and-drinks/food-item/edit.js
```diff
@@ -227,6 +227,26 @@ class FoodAndDrinksEdit extends Component {
...
+										<IconButton
+											icon={ icons.popular }
+											className="wp-block-coblocks-food-item__attribute wp-block-coblocks-food-item__attribute--popular"
+											onClick={ () =>
+												setAttributes( { popular: ! attributes.popular } )
+											}
+											label={ __( 'Popular' ) }
+											isToggled={ attributes.popular }
+										/>
+									</span>
```
jrtashjian: https://github.com/godaddy/coblocks/pull/958/files#diff-a666a2948d2f4865c59523949605e3d2R231-R241
Let's go ahead and just show the toggle in the sidebar and remove this.

richtabor: Actually, I think it'd be better to have `popular` be the first item in the attribute shortcuts, dropping Gluten Free from these shortcuts.

jrtashjian: Adding my notes from Slack here for future reference, then resolving this request.

> Alternatively, should we instead make a style available to the menu-item block? I don’t think the icon really fits well with the category of icons there. Maybe a block style for a popular item (or special, or featured)?

> The existing icons describe the actual food item. At a quick glance someone could see that an item is spicy and either consciously gravitate towards those items or avoid them.

> The star for popular doesn’t feel like it has that same “filter”. I’m just wondering what type of food a star is.

> A style for a popular/special/featured item would draw the attention because it’s styled differently whereas the attribute icons would not but rather provide a mental filter. Similar to how the food items with huge pictures on a menu draw your gaze away from the general list of food items.

> I'm not suggesting we do a completely different layout for a popular item. Instead, applying minor styles to emphasize an item similar to how we have different styles for the alert block: same layout but different ways of communicating.

PR #964, opened 2019-10-08
Add local language support for Map block

### LINKED ISSUE #922: ISBAT view Google Maps in my site's language
**Describe the bug**
The Google Maps iframe URL is incorrectly formatted causing map output to not be localized.

**To Reproduce**
1. Change your site language to Spanish.
2. Load up a page containing the block and inspect the map iframe element.
3. Notice that `&language=ja` is hardcoded (Japanese).
4. The correct query arg to use contains an incorrect value: `&hl=%s`

**Screenshot w/ Incorrect URL**
skitch (4)

**Expected behavior**
1. The `language` query arg should be removed from the URL completely.
2. The `hl` query arg should be used to set the language: `&hl=es`
3. The WordPress locale should be automatically mapped to the appropriate Google locale.

Supported language codes can be found here: https://developers.google.com/maps/faq#languagesupport

Example of locale conversion in the Contact Widgets plugin: https://github.com/godaddy/wp-contact-widgets/blob/v1.6.2/includes/class-contact.php#L256-L288

**Screenshot w/ Correct URL**
skitch (5)

**Isolating the problem:**

**WordPress Version**
5.2.3

**Gutenberg Version**
N/A

paranoia1906:
- Closes #922.

- Pulls the user locale from body class name. 
  * Shout-out to @jrtashjian for the great idea.
- Function then handles string conversion for use with Google Maps API.
  * Thank you @fjarrett for the examples and docs.
- Added deprecated function to handle previous block markup.

image


src/blocks/map/edit.js
```diff
@@ -148,6 +148,37 @@ class Edit extends Component {
 			zoom,
 		} = attributes;
 
+		const userLocale = () => {
+			let locale = $( '[class*="locale-"]' ).attr( 'class' ).split( 'locale-' )[ 1 ].split( ' ' )[ 0 ];
+			switch ( locale.toLowerCase() ) {
```
jrtashjian: https://github.com/godaddy/coblocks/pull/964/files#diff-7ebdffc151d5a58795470ce8ca4ec512R153-R179

So just out of curiosity during testing I wanted to see if there was another way to accomplish this to see if we really needed a switch statement. Turns out we can very easily pull the language attribute from the `<html>` element like this:

```
const locale = document.documentElement.lang;
```

It seems like Google can accept this value without any modifications and load the correct translations. So we can remove this new function entirely and concatenate this value directly here: https://github.com/godaddy/coblocks/pull/964/files#diff-7ebdffc151d5a58795470ce8ca4ec512L243-R274

EvanHerman:
Tested things out this morning - looks like maps generated when an API key was present were not localizing properly. Made some tweaks and got that working, tested things out. Looks good on my end. 👍

PR #966, opened 2019-10-09
Add new Post Carousel block to display a slideshow of blog posts

richtabor:
This PR adds a new Post Carousel block for displaying posts in a slideshow. 

As seen in TwentyTwenty:
Screen Shot 2019-10-09 at 2 24 21 PM


richtabor:
This should merged after #736.

PR #974, opened 2019-10-10
Improve the Map block Google Maps API key UX

### LINKED ISSUE #918: Map Block UX Improvements
This issue lists out the details for a number of UX improvements for the Map block.

image

**Task 1**
The "Remove" button should not be visible, unless there is a key saved

**Task 2**
"Add your Google Maps API key..." should be "Add a Google Maps API key..."

**Task 3**
"Retrieve your key" should be "Get a key"

**Task 4**
If you're editing a saved key, the save button should revert to "Save"
map-api-saved

paranoia1906:
Closes #918

Added `componentDidMount` function to reassign the state for the `apiKey`. This makes it so that the input field is autofilled with the 'saved' `apiKey` whenever the inspector is opened.

mapSaveApiButton

### DIFF
--- a/src/blocks/map/inspector.js
+++ b/src/blocks/map/inspector.js
@@ -204,7 +204,7 @@ class Inspector extends Component {
 						initialOpen={ false }
 						className="components-coblocks-block-settings-sidebar"
 					>
-						<p>{ __( 'Add your Google Maps API key. Updating this API key will set all your maps to use the new key.' ) }</p>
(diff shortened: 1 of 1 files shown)


richtabor:
Looks good from the gif. I'd like to have either @EvanHerman or @jrtashjian take a look behind the scenes to approve.🤘

richtabor:
> Code wise it looks good to me. My only question is, why did we update the copy from 'Add your' to 'Add a' and 'Retrieve your key' to 'Get a key'?

Because they may not have one yet. Insinuating they do may make them think they’re missing something.

PR #979, opened 2019-10-11
Update .po files in build process

EvanHerman:
During the build process (`grunt build`) the .po files are not being updated. This PR runs our `update-po.sh` script to loop through locales and generate updated .po files for each locale. This was a script that needed to be run manually before. After generating the new .pot/.po files, the files will need to be committed back up to the repo.

Can also be triggered with `grunt update-pot`.

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -220,6 +220,7 @@ module.exports = function( grunt ) {
 			].join( ' && ' ),
 			translations: [
 				'npm run babel:build',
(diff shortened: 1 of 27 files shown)


EvanHerman:
Holding off on this while I think about this more. This does not seem like the right approach.

EvanHerman:
Closing as this is already implemented in `grunt build`/`npm run babel:build`. The issue is the translation (.pot/.po) files are not committed back up to Github after updating.

PR #984, opened 2019-10-11
Fix image links being erased when a Gallery block is updated

### LINKED ISSUE #983: Hyperlinks getting removed form pictures when editing the Masonry Gallery
**Describe the bug**
When editing a Masonry gallery to add a new picture, the hyperlinks that were set for the pics that were in the gallery get removed once the new picture is added. 

**To Reproduce**

1. Add a Masonry Gallery to a page or post.
2. Add a few images to the gallery.
3. Click on one of the images and select **Custom URL** under the **Link settings** tab. Then add URL to one or more pics.
4. Then, click on the **Edit** button on the gallery toolbar and add an image to the gallery. 
5. Check the images that were in the gallery already and notice that the URLs set in the previous step are gone.

**Isolating the problem:**

**WordPress Version**
WordPress 5.2.3

EvanHerman:
Resolves #983 

Fixes image URLs getting stripped when a gallery is updated.

### DIFF
--- a/src/blocks/gallery-carousel/controls.js
+++ b/src/blocks/gallery-carousel/controls.js
@@ -24,7 +24,7 @@ class Controls extends Component {
 
 	onSelectImages( images ) {
 		this.props.setAttributes( {
-			images: images.map( ( image ) => helper.pickRelevantMediaFiles( image ) ),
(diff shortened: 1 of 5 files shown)


jrtashjian:
Made a small update so that the fix is applied to all galleries.

PR #985, opened 2019-10-11
Resolve PHP warnings with WordPress 5.3 Beta

paranoia1906:
In 5.3 Beta, WordPress core has introduced a new parameter to be used when adding menu pages. Warnings were thrown until a parameter was properly passed into this function. 

Works on 5.2.3 and 5.3 Beta. Works with and without Gutenberg.

Resolves WordPress.org support requests:
https://wordpress.org/support/topic/add_submenu_page-call-uses-a-string-for-position/
https://wordpress.org/support/topic/activating-the-plugin-on-5-3-beta3-produces-php-warnings/

### DIFF
--- a/includes/admin/class-coblocks-getting-started-page.php
+++ b/includes/admin/class-coblocks-getting-started-page.php
@@ -46,7 +46,8 @@ class CoBlocks_Getting_Started_Page {
 				'CoBlocks',
 				apply_filters( 'coblocks_getting_started_screen_capability', 'manage_options' ),
 				'coblocks-getting-started',
-				array( $this, 'content' )
+				array( $this, 'content' ),
+				4
 			);
 
 			return;
@@ -59,7 +60,7 @@ class CoBlocks_Getting_Started_Page {
 			apply_filters( 'coblocks_getting_started_screen_capability', 'manage_options' ),
 			'coblocks-getting-started',
 			array( $this, 'content' ),
-			'data:image/svg+xml;base64,' . base64_encode( $svg )
+			4,
 		);
 	}
 
--- a/includes/admin/class-coblocks-getting-started-page.php
+++ b/includes/admin/class-coblocks-getting-started-page.php
@@ -60,7 +60,7 @@ class CoBlocks_Getting_Started_Page {
 			apply_filters( 'coblocks_getting_started_screen_capability', 'manage_options' ),
 			'coblocks-getting-started',
 			array( $this, 'content' ),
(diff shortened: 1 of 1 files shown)


richtabor:
This works fine pre WordPress 5.3, correct?

EvanHerman:
Yeah, it should work fine pre 5.3. It's strange because that last argument is optional, and should be set to `null` when not passed in. It's also weird that these functions would break backwards compatibility.

paranoia1906:
@richtabor Yes it works tested on 5.2.3. I was thinking the same @EvanHerman when I seen the $position = null in the function.

richtabor:
So do we need this PR anymore?

jrtashjian:
If it's resolved in Core then I say we close this. @paranoia1906 can you confirm?

I would also say that it's good that we are keeping up with the betas but let's avoid testing on it until it's closer to release date or after release. We don't want to be chasing a moving target.

paranoia1906:
That was a poor choice of wording there. @richtabor @jrtashjian This issue still exists in `5.3-RC1-46558` is what I had meant to indicate. This pull request resolves the issue in the latest RC.

paranoia1906:
There are two support requests on WP.org related to issues people encountered in working on 5.3 beta. This issue is mainly to track ongoing support requests and make sure that the issue is addressed when 5.3 finally drops. I'll go with your judgement on if we should close it or not.

includes/admin/class-coblocks-getting-started-page.php
```diff
@@ -59,7 +60,7 @@ public function screen_page() {
 			apply_filters( 'coblocks_getting_started_screen_capability', 'manage_options' ),
 			'coblocks-getting-started',
 			array( $this, 'content' ),
-			'data:image/svg+xml;base64,' . base64_encode( $svg )
```
jrtashjian: After inspecting further, the issue isn't that we're missing the `$position` parameter (because it is optional and defaults to `null`), it's because we were passing the base64 encoded svg icon as the `$position` parameter. The passed parameters were not updated when switching this to [`add_management_page`](https://developer.wordpress.org/reference/functions/add_management_page/) from [`add_menu_page`](https://developer.wordpress.org/reference/functions/add_menu_page/).

jrtashjian: I've pushed an update which removes the `$position` parameter from our calls since we only needed to remove the svg icon.

PR #987, opened 2019-10-11
Introduce a filter to conditionally disable the typography controls

EvanHerman:
Original idea came from https://wordpress.org/support/topic/hide-typography-controls-to-users/

Allow users to turn on/off the typography controls using a filter. When the controls are disabled the blocks remain formatted, but prevents users from touching pre-set styles.

As discussed in slack, another solution (options within the plugin etc.), may be a better alternative.

#### Post Limits (Only shows on post ID 1)
```php
add_filter( 'coblocks_disable_typography_controls', function( $bool, $post_id ) {

  return ( 1 === $post_id );

}, 10, 2 );
```

#### User Role (Only shows to administrators)
```php
add_filter( 'coblocks_disable_typography_controls', function( $bool, $post_id ) {

  $current_user = wp_get_current_user();

  return in_array( 'administrator', $current_user->roles, true );

}, 10, 2 );
```


src/extensions/typography/controls.js
```diff
@@ -18,6 +19,13 @@ class Controls extends Component {
 			attributes,
 		} = this.props;
 
+		if ( ! coblocksBlockData.disableTypography ) {
```
paranoia1906: This should be `( coblocksBlockData.disableTypography )`. in the current state the typography controls are always disabled.

paranoia1906: I see now that this was related to the administrators filter. This effectively always disables the extension unless a user has applied a filter to allow administrators or specific pages access.

jrtashjian: Fixed.

PR #988, opened 2019-10-12
Add PHP development dependencies with composer

jrtashjian:
We've run into issues trying to install and run the PHP unit tests and running PHP linting with the current configs. This PR makes this easier by utilizing composer to install all dependencies and run the scripts locally as well as on CircleCI.

**To run PHPUnit and PHP_CodeSniffer locally:**

1. Run `composer install` to install all dependencies.

2. Run `sh .dev/bin/install-wp-tests.sh <db-name> <db-user> <db-pass> [db-host] [wp-version]`. The param `db-host` will default to `localhost` and `wp-version` will default to `latest`. This can be re-run in order to setup a different version of WordPress for testing. Example: `sh .dev/bin/install-wp-tests.sh wordpress_test root '' localhost "5.2.3"`

3. Run `composer run lint` to run PHP_CodeSniffer.

4. Run `composer run test` to run PHPUnit.


jrtashjian:
And now we will see the test summary in CircleCI for all three builds.

Screen Shot 2019-10-12 at 3 10 46 AM

includes/class-coblocks-font-loader.php
```diff
@@ -37,16 +37,16 @@ public static function register() {
...
-	private $_version;
+	private $version;
 
 
 	/**
 	 * The Constructor.
 	 */
 	public function __construct() {
-		$this->_version = COBLOCKS_VERSION;
+		$this->version = COBLOCKS_VERSION;
```
EvanHerman: Is this needed? Can we just reference `COBLOCKS_VERSION` throughout?

includes/class-coblocks-form.php
```diff
@@ -405,21 +403,24 @@ private function render_submit_button( $atts ) {
...
 
-			$styles = " style='{$styles}'";
+			$styles = ' style="' . esc_attr( $styles ) . '"';
 
 		}
 
+		// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
 		?>
 
 		<button type="submit" class="wp-block-button__link <?php echo esc_attr( $btn_class ); ?>"<?php echo $styles; ?>><?php echo esc_html( $btn_text ); ?></button>
```
EvanHerman: Does the output of `$styles` need to be escaped here?

includes/class-coblocks-form.php
```diff
@@ -405,21 +403,24 @@ private function render_submit_button( $atts ) {
...
 
 		<?php
-
+		// phpcs:enable WordPress.Security.EscapeOutput.OutputNotEscaped
```
EvanHerman: What is this ignoring?

includes/admin/class-coblocks-crop-settings.php
```diff
@@ -92,6 +114,7 @@ public function get_original_image() {
 	 * Cropping.
 	 */
 	public function api_crop() {
+		wp_verify_nonce( $_POST['cropSettingsNonce'], 'cropSettingsNonce' );
```
EvanHerman: We'll want to sanitize the nonce value before using it.

```php
$nonce = filter_input( INPUT_POST, 'cropSettingsNonce', FILTER_SANITIZE_STRING );

if ( ! $nonce ) {

   return;

}

wp_verify_nonce( $nonce, 'cropSettingsNonce' );
...
```

EvanHerman:
This is looking good. Good idea updating that stuff. One thought, should we kill the `php71-build` and possibly setup a `php56-build` running PHP 5.6 (minimum supported PHP version) and keep our 7.2 build? That way were testing the minimum and latest supported PHP versions? Testing PHP 7.1 and 7.2 might be redundant since they're so close in versions.

includes/class-coblocks-font-loader.php
jrtashjian: Good catch. Didn’t think to look into how it was used, just removed the unnecessary define abstraction.

includes/class-coblocks-form.php
jrtashjian: I don’t think it does but Wordpress rules will always trigger the warning. I will solve this a better wat so that we don’t disable rules.

includes/class-coblocks-font-loader.php
jrtashjian: Fixed.

includes/class-coblocks-form.php
jrtashjian: Fixed.

includes/admin/class-coblocks-crop-settings.php
jrtashjian: Fixed.

includes/class-coblocks-form.php
jrtashjian: Removed.

jrtashjian:
@EvanHerman Replaced the PHP 7.1 build with PHP 5.6!

includes/admin/class-coblocks-crop-settings.php
```diff
@@ -19,8 +19,24 @@ class CoBlocks_Crop_Settings {
...
+	/**
+	 * This class's instance.
+	 *
+	 * @var CoBlocks
+	 * @since 1.14.0
+	 */
 	private static $instance;
 
+	/**
+	 * Main CoBlocks Instance.
```
richtabor: This should be updated, and the description below as well.

richtabor:
How does this work with `npm`? Does this change the "[Development](https://github.com/godaddy/coblocks#development)" section of our readme? Just trying to understand what the flow would be for someone setting up CoBlocks for development (if it's changed).

jrtashjian:
> How does this work with npm?

This is unrelated to `npm`, instead we use `composer` to run the scripts. Although if and when we migrate to `wp-scripts`, `npm` would be utilized.

> Does this change the "Development" section of our readme?

No. Everything here is the same. We should add a new doc for settings this stuff up similar to how Gutenberg has for their [Testing Overview](https://github.com/WordPress/gutenberg/blob/master/docs/contributors/testing-overview.md).

includes/admin/class-coblocks-crop-settings.php
jrtashjian: Fixed.

jrtashjian:
I've opened a new issue for adding the docs later. I don't think we need them now in order to merge this. I will resolve the conflicts with master and merge shortly.

PR #993, opened 2019-10-13
Remove "Leave a Review" feedback notice

richtabor:
This PR removes the feedback notice that asks users for a review, in an effort to de-clutter the admin.


fjarrett:
💯

PR #995, opened 2019-10-14
Fix the block validation error when the Hero block background color is removed

### LINKED ISSUE #729: ISBAT clear/default background color in Hero block
**Describe the bug**
When clearing the background color set under the Color Settings and saved, the Hero block fails block validation. 

Block validation fails due to missing className of 'has-background'.

**To Reproduce**

1. 'Add Hero block'
2. 'Click on the Hero block to edit'
3. 'Access Color Settings and clear the background color'
4. 'update/save page'
5. 'refresh the browser page'
6. See error

**Screenshots**
heroBackgroundClearError

**Expected behavior**
I expect hero to either allow a transparent background or to default to a set color when hitting the clear button.

**Isolating the problem:**

**WordPress Version**
5.2.2

**Gutenberg Version**
N/A

Issue originally reported on wordpress.org
https://wordpress.org/support/topic/hero-block-background-color-clear-error/

EvanHerman:
Resolves #729 

Removing the default value for the hero is the simplest approach to resolving #729 . When you clear a color value, it's set to `undefined`. During the save routine, in core, if the value is `undefined` and a `default` setting exists, it falls back to that.

Without introducing additional login and checks, removing the default value from the background color for the hero block is the easiest approach. @richtabor @jrtashjian @paranoia1906 Do you think this is our best approach?

### DIFF
--- a/src/blocks/hero/block.json
+++ b/src/blocks/hero/block.json
@@ -48,8 +48,7 @@
 			"default": 60
 		},
 		"customBackgroundColor": {
-			"type": "string",
-			"default": "#f3f3f3"
+			"type": "string"
 		},
 		"height": {
 			"type": "number",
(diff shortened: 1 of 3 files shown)


richtabor:
Will this require a deprecation for folks who're using the current default background color?

jrtashjian:
The snapshot was updated but if it’s invalid, the deprecation tests should have failed. I don’t think in this case it would. Removing the default value doesn’t change the existing value of the added block.

PR #996, opened 2019-10-14
Ensure style previews of the Icon block display the selected icon

### LINKED ISSUE #899: ISBAT see Icon block icons as filled our outlined style
**Describe the bug**
The style preview located in the inspector sidebar shows the differences with Icon block icons using either the `filled` or `outlined` styles. Since Gutenberg 6.4 the icons are not properly randomized and will now incorrectly display style previews.

**To Reproduce**

1. Add Icon block.
2. Click on block and view sidebar.
3. Open styles panel to see icon previews.

**Screenshots**

random-icons

**Expected behavior**
Expect style previews to accurately reflect the `filled` or `outlined` style respectively. We may also want to settle for a default logo shown instead of randomizing the shown icon.

**Isolating the problem:**

**WordPress Version**
5.2.3

**Gutenberg Version**
6.5

paranoia1906:
- Closes #899 
- Removed `iconRand` attribute.
- Set icon default to `''` (unset).
- Static icon used for style previews.
  * I chose aperture because this style shows the difference between filled and outlined well. 
- No deprecation needed from my testing.

iconsStaticStylePreviews

### DIFF
--- a/src/blocks/icon/edit.js
+++ b/src/blocks/icon/edit.js
@@ -27,18 +27,19 @@ const MAX_ICON_SIZE = 400;
 
 export { MIN_ICON_SIZE, MAX_ICON_SIZE };
 
-/**
- * Block edit function
- */
 class Edit extends Component {
-	componentDidMount() {
-		// Randomized the default icon when first added.
-		const defaultIcons = [ 'aperture', 'gesture', 'scatter_plot', 'waves', 'blocks', 'coblocks', 'drafts', 'device_hub', 'marker' ];
-		const rand = defaultIcons[ Math.floor( Math.random() * defaultIcons.length ) ];
+	isStylePreview() {
+		return ( $( `#block-${ this.props.clientId }` ).parent().parent().hasClass( 'block-editor-block-preview__content' ) );
+	}
 
-		// CHeck if the icon is the default.
-		if ( this.props.attributes.iconRand === true ) {
-			this.props.setAttributes( { icon: rand, iconRand: false } );
+	componentDidMount() {
+		if ( this.isStylePreview() ) {
+			this.props.setAttributes( { icon: 'aperture' } );
+		} else {
+			// Randomized the default icon when first added.
+			const defaultIcons = [ 'aperture', 'gesture', 'scatter_plot', 'waves', 'blocks', 'coblocks', 'drafts', 'device_hub', 'marker' ];
+			const rand = defaultIcons[ Math.floor( Math.random() * defaultIcons.length ) ];
+			this.props.setAttributes( { icon: rand } );
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


src/blocks/icon/edit.js
```diff
@@ -27,18 +27,19 @@ const MAX_ICON_SIZE = 400;
...
+	isStylePreview() {
+		return ( $( `#block-${ this.props.clientId }` ).parent().parent().hasClass( 'block-editor-block-preview__content' ) );
+	}
 
-		// CHeck if the icon is the default.
-		if ( this.props.attributes.iconRand === true ) {
-			this.props.setAttributes( { icon: rand, iconRand: false } );
+	componentDidMount() {
+		if ( this.isStylePreview() ) {
+			this.props.setAttributes( { icon: 'aperture' } );
```
jrtashjian: This conditional shouldn't be needed and we should avoid changing the content for the style preview. The point of the style preview is to see your content in the different styles.

src/blocks/icon/edit.js
```diff
@@ -27,18 +27,19 @@ const MAX_ICON_SIZE = 400;
...
 
-/**
- * Block edit function
- */
 class Edit extends Component {
-	componentDidMount() {
-		// Randomized the default icon when first added.
-		const defaultIcons = [ 'aperture', 'gesture', 'scatter_plot', 'waves', 'blocks', 'coblocks', 'drafts', 'device_hub', 'marker' ];
-		const rand = defaultIcons[ Math.floor( Math.random() * defaultIcons.length ) ];
+	isStylePreview() {
```
jrtashjian: Remove this after removing the conditional in `componentDidMount`

jrtashjian:
ok, so I figured something out here which required looking directly at how the `BlockStyles` component works. And after seeing the code, it makes 100% sense why we’re only having this problem in Gutenberg. Look here: https://github.com/WordPress/gutenberg/blob/b5fe68187f94c0a204a336825e7c6aa6042c48aa/packages/block-editor/src/components/block-styles/index.js#L147-L170

The `BlockStyles` component is rendered with the our block being passed to it (through `withSelect`). Basically anytime we change attributes in our block, the `BlockStyles` component reacts to it. And when we change the style, the `BlockStyles` component can update our `className`’s (through `withDispatch`)

Here is where the problem happens: https://github.com/WordPress/gutenberg/blob/b5fe68187f94c0a204a336825e7c6aa6042c48aa/packages/block-editor/src/components/block-styles/index.js#L125-L135

The actual block is cloned into the preview _unless_ we have a block example defined!

So the issue is actually with our defined example: https://github.com/godaddy/coblocks/blob/2fac03a3b14e8b3336653dbe88596790afde49b8/src/blocks/icon/index.js#L40-L44

We’re only setting the width attribute and no others, which is why our icon was always undefined. We need to pass in a minimum set of attributes. We’ll need to remove the example for this block. We also need to add checks in the `inspector.js` file for the `icon` attribute because it’s assuming `attributes.icon` is set and that breaks things.

PR #1013, opened 2019-10-16
Use number input for height controls Carousel block

### LINKED ISSUE #1011: ISBAT tweak the Carousel block height using a number input
Gutenberg uses number inputs for "Height in pixels", we should do the same for the Carousel's Height in pixels control, to follow core UX patterns.

Screen Shot 2019-10-15 at 7 17 10 PM

paranoia1906:
Closes #1011 

**In action**
carouselHeightControls

**Implementation in `core/cover`**
image

@richtabor. Should the height control be in a separate `Dimensions` panel like the example above?


richtabor:
> I think it would be best to remove the `max` attribute on the number input field and remove the CSS additions. We removed the slider control for height on the map block and just decided to remove the `max` attribute altogether - which will fix the width of the number input.
> 
> #977

Agreed!

richtabor:
> @richtabor. Should the height control be in a separate `Dimensions` panel like the example above?

No, I'd like to keep the panel count as low as possible. Core is doing that because it's the only option in that panel.

paranoia1906:
Styles are removed and max height is not set.

image

PR #1017, opened 2019-10-17
Allow dynamic subject lines with Form block submissions

### LINKED ISSUE #975: ISBAT use dynamic subject lines in Form block submissions
The Form block allows the site owner to set a static subject line used with form submissions. 
image

This functionality could be enhanced by allowing a user to implement dynamic subject lines generated from user submitted content. For example of this we can look to Contact Form 7
image
This plugin uses variables for user submitted data.

Common practice from many hosting providers is to scan and flag subject lines when being sent repeatedly without change as potential spam. This change could prevent this scenario. 

The Form block should allow a site owner to have more dynamic control over subject lines.

AnthonyLedesma:
Closes #975 

Add filters to pull out form submission data to use in subject lines. Also added regex logic to grab possible variations of the mail-tags that we allow. 

**Supported tags:** 
- `[name]`
  * Uses the value from the name input 
- `[email]`
  * Uses the value from the email input

**Supported fields:**
`Subject`

**Variations are caught**
`[ email]` `[ EmAiL ]` `[E m a Il  ]` `[  NAME  ]` 

**If does not contain exactly the characters than the tag is used as plain text.**
`[1email]` `[nam]`

**Question**: How do we feel about the label wording and placement?

image


includes/class-coblocks-form.php
```diff
@@ -680,6 +698,18 @@ public function process_form_submission( $atts ) {
 			$post_id
 		);
 
+		$short_code_regex = '/\[(.*?)\]/i';
+		if ( preg_match_all( $short_code_regex, $subject, $matches_out ) ) {
+			foreach ( $matches_out[0] as &$value ) {
```
fjarrett: @paranoia1906 are we sure `$matches_out[0]` could never throw an undefined index notice?

richtabor:
Do you think we could automate this? 

Subject: `{subject}, from {email address}`

includes/class-coblocks-form.php
paranoia1906: I was not able to get an undefined index here. Since the `$matches_out[0]` is conditional to a 'truthy' `preg_match_all` execution. This returns null without a match.

paranoia1906: I think this covers all options a user may try. Let me know if I am off track.

paranoia1906:
> Do you think we could automate this?
> 
> Subject: `{subject}, from {email address}`

Automate as in, always use dynamic subject lines? Should the control be taken from the user? Or do you mean more like setting a default that can be changed?

richtabor:
> Automate as in, always use dynamic subject lines? Should the control be taken from the user? Or do you mean more like setting a default that can be changed?

Yea always use them so users' emails won't get flagged by hosts as potential spam.

paranoia1906:
Default subject for Form block changed to `Form submission from [email]`.

AnthonyLedesma:
Conflicts resolved and working on 5.2.4 & 5.3 with and without GB running.

AnthonyLedesma:
Note that this branch has 'fix/email-field-id' merged in. I suggest we merge 'fix/email-field-id' into master before we merge this PR.

includes/class-coblocks-form.php
```diff
@@ -897,6 +904,66 @@ public function process_form_submission( $atts ) {
...
+	 *       [name] will be replaced with the value of field-name etc.
+	 *
+	 * @param  array  $atts           Block attributes array.
+	 * @param  string $email_field_id Email field ID.
+	 * @param  string $name_field_id  Nane field ID.
+	 * @return string Email subject.
+	 */
+	private function setup_email_subject( $atts, $email_field_id, $name_field_id ) {
+
+		$default_subject = __( 'Form submission from [email]', 'coblocks' );
```
EvanHerman: @AnthonyLedesma It would be ideal if we could set this in a class property or something so that we're not re-defining it. It's also defined here https://github.com/godaddy-wordpress/coblocks/pull/1017/files#diff-6889c8f6a940ab86e23a2cf5809324ccR114

includes/class-coblocks-form.php
```diff
@@ -897,6 +904,66 @@ public function process_form_submission( $atts ) {
...
+		if ( isset( $matches[1] ) ) {
+
+			array_walk(
+				$matches[1],
+				function( $match, $key ) use ( $matches, &$subject, &$email_field_id, &$name_field_id ) {
+					$slug_match = strtolower( preg_replace( '/\s+/', '', $match ) );
+
+					// phpcs:disable WordPress.Security.NonceVerification.Missing
+					if ( 'name' === $slug_match ) {
+						if ( isset( $_POST[ $name_field_id ]['value'] ) ) {
```
EvanHerman: We'll probably want to add an empty line before this, and again between lines 942 and 943.

includes/class-coblocks-form.php
```diff
@@ -897,6 +904,66 @@ public function process_form_submission( $atts ) {
...
+
+							$name_field_value = is_array( $_POST[ $name_field_id ]['value'] ) ? sanitize_text_field( implode( ' ', $_POST[ $name_field_id ]['value'] ) ) : sanitize_text_field( $_POST[ $name_field_id ]['value'] );
+							$value            = ( '' === $name_field_value ) ? $matches[0][ $key ] : $name_field_value;
```
EvanHerman: We can probably run a check to see if `$name_field_value` is empty instead of an empty string.
```php
$value = empty( $name_field_value ) ? $matches[0][ $key ] : $name_field_value;
```

includes/class-coblocks-form.php
jrtashjian: Also, the `[email]` part of the string should not be translatable. I'd use `sprintf` so we can add placeholders here. There are examples in the codebase already.

src/blocks/form/edit.js
```diff
@@ -280,6 +280,7 @@ class FormEdit extends Component {
 					}
 					onChange={ this.onChangeSubject }
 				/>
+				{ __( 'In the Subject field you can use these mail-tags: [name] [email]' ) }
```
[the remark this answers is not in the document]
jrtashjian: Same as above: Also, the `[name]` and `[email]` part of the string should not be translatable. I'd use `sprintf` so we can add placeholders here. There are examples in the codebase already.

includes/class-coblocks-form.php
AnthonyLedesma: These have both been changed.

includes/class-coblocks-form.php
AnthonyLedesma: Spaced out per phpcs standards now.

includes/class-coblocks-form.php
AnthonyLedesma: using `empty()` now.

src/blocks/form/edit.js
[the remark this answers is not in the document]
AnthonyLedesma: Changed these around. Ready for another reivew.

AnthonyLedesma:
formSubjectButtons

richtabor:
Let's replace the help text with "You may use the following tags in the subject field:" and move it closer to the Subject field, as if it were actual help text.

Here's an example from the Image block:  
Screen Shot 2019-12-03 at 3 09 15 PM

Also, is there a reason why there are spaces inside the `[ email ]` and `[ name ]` tags?

AnthonyLedesma:
> Let's replace the help text with "You may use the following tags in the subject field:" and move it closer to the Subject field, as if it were actual help text.

Sounds good to me!

> Also, is there a reason why there are spaces inside the [ email ] and [ name ] tags?

Just for readability reasons. Here is a screenshot without the spacing. 
image

What I could do here is target these buttons and add a little bit of margin to help with readability.

AnthonyLedesma:
@richtabor Now using the help prop for the subject input field.
image

I don't like the idea of adding more CSS to add margins to the buttons so I tried to use our utility classes here `mr-1 ml-1`. Unfortunately, I found that Styles from core specifically `.components-button.is-link` were overriding our utility class settings with a `margin: 0;`. In lieu of adding more styles, I suggest we stick with current spacing.

PR #1023, opened 2019-10-18
Use descriptive element name for Form block radio buttons

### LINKED ISSUE #1020: Form radio fields are named 'choose-one'
**Describe the bug**
Each form field is named appropriately, `name`, `email`, `phone` etc. The radio fields are generating using the name `choose-one`, which does not follow the naming convention as above. This should be `field-radio`.

**Screenshots**
image

**Expected behavior**
Radios use the name `field-radio`.

**Isolating the problem:**

AnthonyLedesma:
- Closes #1020 

Use wording to reflect the element type in markup. Issue resolved by altering the label slug before using the value.

image


richtabor:
@AnthonyLedesma Mind taking a look at these tests? Thanks!

PR #1024, opened 2019-10-18
Ensure translations are loaded in the Block Editor

EvanHerman:
Begin fixing the JS translations by adding the proper `coblocks` text domain.


class-coblocks.php
```diff
@@ -168,7 +168,7 @@ public function load_textdomain() {
 		 */
 		public function block_localization() {
 			if ( function_exists( 'wp_set_script_translations' ) ) {
-				wp_set_script_translations( 'coblocks-editor', 'coblocks' );
+				wp_set_script_translations( 'coblocks-editor', 'coblocks', __DIR__ . '/languages' );
```
fjarrett: @EvanHerman you may wanna do `COBLOCKS_PLUGIN_DIR . 'languages'` here since that constant seems to be the pattern used elsewhere

PR #1032, opened 2019-10-18
Improve Row block stability through block validation testing

jrtashjian:
Provides deprecation tests for the Row block as part of #872.

```
Test Suites: 2 passed, 2 total
Tests:       297 passed, 297 total
Snapshots:   1 passed, 1 total
Time:        4.967s, estimated 5s
```


src/blocks/row/test/__snapshots__/save.spec.js.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`coblocks/row should render 1`] = `
+"<!-- wp:coblocks/row {\\"columns\\":2} -->
+<div class=\\"wp-block-coblocks-row\\" data-columns=\\"2\\"><div class=\\"wp-block-coblocks-row__inner has-medium-gutter\\"></div></div>
+<!-- /wp:coblocks/row -->"
```
paranoia1906: I think this markup may not match that master branch markup for row block save function. Here is the markup I generated with bare minimum attribute of 2 columns.
```javascript
"<!-- wp:coblocks/row {\\"columns\\":2} -->
<div class=\\"wp-block-coblocks-row\\" data-columns=\\"2\\"><div class=\\"wp-block-coblocks-row__inner has-medium-gutter has-no-padding has-no-margin is-stacked-on-mobile\\"></div></div>
<!-- /wp:coblocks/row -->"
```

I realize that you have also made the necessary changes to edit.js, save.js, and deprecated.js so that the `has-no-padding` and `has-no-margin` classes are not applied. This class has no value and I approve of the specific change but please note that that change should also be made within `Column`, `Features`, `Feature`, and `Media Card` blocks. Each of these blocks use the same format of 
```javascript
( paddingSize !== 'advanced' ),
```
as opposed to 
```javascript
( ! [ 'no', 'advanced' ].includes( marginSize ) ),
```
For the reasons of pull request scope, and the uncovering of a larger issue with the  logic for processing of `DimensionsControl` attributes, I suggest that the changes in `edit.js`, `save.js`, and `deprecated.js` be revisited in a separate pull request.

paranoia1906:
Followup on this. #1036  is created to detail the issue with classes improperly generated from Dimensions Control attributes.

src/blocks/row/test/__snapshots__/save.spec.js.snap
jrtashjian: Good catch! This indeed causes a block validation issue. I've reverted the condition change.

PR #1038, opened 2019-10-21
Use the jQuery global to avoid script conflicts with Gif block

paranoia1906:
While testing WP 5.3 RC I found that the Gif block references `$` variable to access `jQuery`. At time of testing 5.3 RC has removed `$` from `window` while `jQuery` variable remains. This change would allow compatibility **with** and **without** WordPress 5.3 and would also follow the best practice that is already used by CoBlocks to reference the `jQuery` variable.

### DIFF
--- a/src/blocks/gif/edit.js
+++ b/src/blocks/gif/edit.js
@@ -113,7 +113,7 @@ class Edit extends Component {
 
 			setAttributes( { fetching: true } );
 
-			$.getJSON( GIPHY_URL + encodeURI( search ) )
+			jQuery.getJSON( GIPHY_URL + encodeURI( search ) )
 				.success( function fetchSuccess( data ) {
 					setAttributes( { fetching: false, matches: data.data } );
 				} )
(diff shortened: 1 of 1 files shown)


JJJ:
@paranoia1906 Can you provide context into the WordPress core change that makes this necessary? I've searched Trac and the Make sites, and haven't found anything, though it's likely I'm not searching for the right things.

paranoia1906:
@JJJ I am not able to find any evidence that the `$` should have been exposed on the window object one way or the other. I am finding evidence to the contrary, that the $ symbol should never have been available in the global scope. When/where that symbol became undefined in the window object is unclear. 

Getting a look at official documentation, and trac tickets we can see that the best practice is to use scoped function expressions. Get a look at these resources from WordPress which specify the best practice on accessing the Global jQuery library.

https://developer.wordpress.org/reference/functions/wp_enqueue_script/#jQuery_noConflict_wrappers
 
https://core.trac.wordpress.org/ticket/16494#comment:41

I am curious to know what changed, so do let me know if you find out.

spencerfinnell:
For anyone finding this in the future, there is some background here: https://core.trac.wordpress.org/ticket/48568

Issue #1051, opened 2019-10-22
Style the Row block placeholder layout buttons to resemble core


richtabor:
**Describe the bug**
Our Row block's placeholder layout picker buttons are not appropriately styled according to the style changes coming to Gutenberg in WordPress 5.3

**Screenshots**
<img width="771" alt="Screen_Shot_2019-10-13_at_8 27 02_PM" src="https://user-images.githubusercontent.com/1813435/67278780-a52dec00-f497-11e9-9ef3-5cda1b23eece.png">

**Expected behavior**
Our Row block layout selector buttons are styled with the appropriate colors to remember core's styling. We should be using `theme(primary)`  and other values to ensure the colors are translated properly with particular admin color schemes.

PR #1055, opened 2019-10-22
Allow theme colors in Row block placeholder

paranoia1906:
I have two commits on this branch and am looking for design feedback.

Commit `2f15c96` accomplishes a similar look to WP 5.3. This route give the placeholder a typical look in the editor when viewing in 5.2.4. 

**WP 5.3**
https://prnt.sc/pmpsyk

**WP 5.2.4**
https://prnt.sc/pmpt84

Commit `9b8962b` is using `theme(primary)` to set SVG and border colors for button. This route accomplishes a similar look to WP 5.3  but effects 5.2.4 as well. 

**How it looks in 5.3 and 5.2.4**
https://prnt.sc/pmpsit

### DIFF
--- a/src/components/visual-dropdown/styles/editor.scss
+++ b/src/components/visual-dropdown/styles/editor.scss
@@ -31,14 +31,8 @@
 		margin: 0;
 		padding: 4px;
 
-		&:hover {
-			background: $light-gray-200;
-			color: $dark-gray-900;
-		}
-
 		&:focus {
 			box-shadow: 0 0 0 2px theme(primary);
-			color: $dark-gray-900;
 			outline: 2px solid transparent;
 			outline-offset: -2px;
 		}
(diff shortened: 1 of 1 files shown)


paranoia1906:
closes #1051

richtabor:
I'd like feedback from the others as well, but I think it's fine to merge the styles prior to 5.3 landing.

src/components/visual-dropdown/styles/editor.scss
```diff
@@ -186,29 +169,13 @@
 	width: 63px;
 }
 
-.components-coblocks-visual-dropdown__popover {
```
richtabor: Let's make sure this doesn't affect the Map block's styles dropdown (if you have a Google Map API Key)

paranoia1906: It was indeed effected. I had to adjust lines 175-181 to get the styles to look correct but that is now adjusted. Looks right in 5.2.4 and 5.3. Looks good with and without GB running.

PR #1065, opened 2019-10-23
Bump WordPress Core support to 5.3

paranoia1906:
Looks like grunt references this version for the find/replace actions. This change is part of the 5.3 compatibility testing.


jrtashjian:
This should not be merged in until we have resolved all 5.3 tagged issues. Putting the waiting tag on until then.

PR #1066, opened 2019-10-23
Fix code display styles of the Gist block with the TwentyTwenty theme

### LINKED ISSUE #1064: ISBAT use Gist block with TwentyTwenty theme
**Describe the bug**
The Gist block does not reflect the editor view when published on Twenty Twenty theme. 

**To Reproduce**

1. Add Gist block and provide Gist url.
2. Publish and view page.
3. See error

**Screenshots**
**Gist with `table-layout: fixed;`**
image

**Gist with `table-layout: auto;`**
image

Styles are set from `wp-content/themes/twentytwenty/style.css?ver=1.0`

**Isolating the problem:**

**WordPress Version**
5.3 RC2

**Gutenberg Version**
N/A

paranoia1906:
Closes #1064.

We can accomplish this by setting the `table-layout` explicit auto. This allows the gist table to render correctly in TwentyTwenty, TwentyNineteen, TwentyEighteen, SoHo, and Barebones theme (underscores).

### DIFF
--- a/src/blocks/gist/styles/style.scss
+++ b/src/blocks/gist/styles/style.scss
@@ -57,4 +57,8 @@
 		margin-top: 0.5em;
 		text-align: center;
 	}
+
+	table {
+		table-layout: auto;
+	}
 }


src/blocks/gist/styles/style.scss
```diff
@@ -57,4 +57,8 @@
 		margin-top: 0.5em;
 		text-align: center;
 	}
+
+	table {
+		table-layout: auto;
+	}
```
paranoia1906: In terms of targeting only `.is-twentytwenty`, I am not sure that we should limit in this case. `table-layout: auto` seemed to render correctly in the different themes. What are your thoughts?

richtabor: If `table-layout: auto` doesn't cause any issues to add, then we wouldn't need to target `is-twentytwenty`. 🤘

jrtashjian: The TwentyTwenty theme seems to be setting `table-layout: fixed` intentionally. I don't understand why this behavior is being set but `auto` is fine (and the browser's default).

Issue #1029, opened 2019-10-18
ISBAT ensure the integrity of the Posts block through automated tests


jrtashjian:
Automated tests need to be added to the `posts` block which follow the groundwork from #835.

#### Tests Required:
- [x] The save function (server rendered block #1069)
- [x] Block transforms #1233
- [x] Block deprecation (not applicable)

#### AC
- Any attribute that modifies the serialized block output needs to have tests.
- Any existing transforms to or from a block needs to have tests (if applicable).
- Any existing block deprecations need to have tests (if applicable).

AnthonyLedesma:
Closing this issue with the completion of #1233

PR #1069, opened 2019-10-23
Improve Posts block stability through block validation testing

jrtashjian:
This block did not have deprecations. Instead I added basic PHP tests on the server rendered function. Related to #1029.


.dev/tests/phpunit/src/blocks/posts/test-index.php
```diff
@@ -0,0 +1,92 @@
...
+				$this->fail( "$action_data[0] is not attached to $action_data[1]." );
+			}
+		}
+
+		$this->assertTrue( true );
+	}
+
+	/**
+	 * Test the social block markup returns correctly
+	 */
```
paranoia1906: change this to `posts block`

PR #1074, opened 2019-10-24
Hide avatar placeholder on Author block when not in use

### LINKED ISSUE #1050: ISNBAT see the Author block avatar placeholder if I'm not adding an avatar
I am presented with an avatar placeholder, even though I may not have any intentions of adding an avatar to the Author block.

If there is no avatar, and I am not selected on the block, the avatar placeholder should not be visible. I may still however drop an image into the Author block's dropZone, to add one. 

This is actually already how the front-end works.

AnthonyLedesma:
Closes #1050 

AuthorBlockImagePlaceholder

An issue here is that the child block `core/button` does not store a prop of `isSelected`. Reading props by using the following: 
```javascript
const childButton = wp.data.select( 'core/block-editor' ).getBlocksByClientId( clientId )[ 0 ].innerBlocks[ 0 ];
```

Any ideas on how to conditionally show when the `core/button` is selected?


paranoia1906:
Here is the updated behavior after implementing the suggested changes. 
authorBlockHidePlaceholderImage

richtabor:
Currently the `dropZone` is not available unless there's an image already there - or it's selected. Let's make the `dropZone` available at all times.

AnthonyLedesma:
@richtabor Made the suggested change. Now `dropZone` is available at all times. Also added the necessary logic to persist the image when it exists.

richtabor:
@jrtashjian Mind taking another peek at this one? Thanks!

PR #1068, opened 2019-10-23
ISBAT use Dynamic HR block in TwentyTwenty theme

paranoia1906:
Closes #1063.

Three rules are causing unexpected behavior with the Dynamic HR block. These are the specific changes from TwentyTwenty that cause the behavior.

https://github.com/WordPress/twentytwenty/blob/f257dfff755c415dadc56a2c70a412421f39aad7/style.css#L427

https://github.com/WordPress/twentytwenty/blob/f257dfff755c415dadc56a2c70a412421f39aad7/style.css#L444

https://github.com/WordPress/twentytwenty/blob/f257dfff755c415dadc56a2c70a412421f39aad7/style.css#L439

When these lines are commented out, the block renders correctly. Looking for advice on how to properly override these styles. I am finding this is not a trivial step.

PR #1067, opened 2019-10-23
Fix Collage Block layouts for the TwentyTwenty theme

paranoia1906:
Closes #1062.

Pull request add `margin: inherit` to `li` elements scoped for collage block. This change allows correct display in TwentyTwenty. This change still looks correct in TwentyNineteen, TwentySeventeen and Barebones theme (underscores).

PR #1078, opened 2019-10-24
Add Twenty Twenty as an auto-generated theme classes

richtabor:
This PR adds a the Twenty Twenty default theme to the list of theme names that get appended to the body class if active. This way we can add theme-specific support for default themes.


paranoia1906:
Class is being added to body now. Need to consider if #1066 #1067 #1068 should be updated to target this class.
image

richtabor:
> Class is being added to body now. Need to consider if #1066 #1067 #1068 should be updated to target this class.
> image

As long as those fixes are not negatively affecting styling across other themes/Go/core Gutenberg (without editor styles) then they're fine. But if they do affect them, we'll need to specify with `.is-twentytwenty`.

PR #1079, opened 2019-10-24
Fix Dynamic HR block display on TwentyTwenty

richtabor:
This PR is a test to see if we can inherit the core Separator's styling, so we wouldn't need to continue supporting individual styles. It's essentially a Separator + Spacer block combined. 

If we go this route we'll need to add a proper deprecation. 

A possible alternative to #1068.

### DIFF
--- a/src/blocks/dynamic-separator/deprecated.js
+++ b/src/blocks/dynamic-separator/deprecated.js
@@ -1,0 +1,47 @@
+/**
+ * External dependencies
+ */
+import classnames from 'classnames';
+
+/**
+ * Internal dependencies
+ */
+import { attributes } from './block.json';
+
+/**
+ * WordPress dependencies
+ */
+import { getColorClassName } from '@wordpress/block-editor';
+
+const deprecatedClassBlockSeparator = ( { attributes } ) => {
+	const {
+		color,
+		customColor,
+		height,
+		className,
+	} = attributes;
+
+	const colorClass = getColorClassName( 'color', color );
+
+	const classes = classnames(
+		className, {
+			'has-text-color': color || customColor,
+			[ colorClass ]: colorClass,
+		} );
+
+	const styles = {
+		color: colorClass ? undefined : customColor,
+		height: height ? height + 'px' : undefined,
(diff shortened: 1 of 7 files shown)


paranoia1906:
Added block deprecation. Experiencing an issue with setting text colors (hr divider colors) in twentytwenty theme. Found a few conflicts that should be addressed. 

**View in editor**
image

**View when published**
image

Issue seems to be conflicting styles from twentytwenty.

https://github.com/WordPress/twentytwenty/blob/dea9290e7ca3d38b7067c3b7107787db6554249a/style.css#L2591
```css
:root .has-accent-color {
	color: #cd2653;
}
```

https://github.com/WordPress/twentytwenty/blob/dea9290e7ca3d38b7067c3b7107787db6554249a/style.css#L419
```css
.entry-content hr:not(.has-background),
hr.styled-separator {
	color: #6d6d6d;
}
```

jrtashjian:
> * Change markup to use a `div` instead of `hr` element.

@paranoia1906 What is the reason for this change? An HR is syntactically correct in this context.

paranoia1906:
@jrtashjian The reason for the change in markup is detailed in this comment. 
https://github.com/godaddy/coblocks/pull/1079#issuecomment-547590928

Essentially, TwentyTwenty was overriding color settings simply because it was an `hr` element. I would love to know what type of alternative could be used there.

richtabor:
We should stick to using `hr` here.

jrtashjian:
I'm having difficulty understanding what we're trying to accomplish here or maybe just the method we're doing it. It sounds like we should just remove our Dynamic HR block and migrate over to the `core/separator` block. We can then add a new "blank" style to the `core/separator` block and add our height control to it.

This could be a nice test for how we remove support for certain blocks in preference of core blocks and even migrating block gallery blocks to the CoBlocks versions.

I suggest we close this PR and start a new one for enhancing the `core/separator` block.

paranoia1906:
This request is closed in favor of https://github.com/godaddy/coblocks/pull/1093

PR #1084, opened 2019-10-30
Improve the column control of the Posts and Post Carousel blocks

### LINKED ISSUE #1044: ISNBAT select a number of columns smaller than the number of posts
**Describe the bug**
As a user, I should not be able to set a number of columns lower than the number of posts I'd like to display. This would prevent display issues. For example, if I select to display two posts - but have a three column layout - my posts will not display properly.

**Screenshots**
Screen Shot 2019-10-21 at 7 40 49 PM

**Expected behavior**
I should not be able to set a number of columns lower than the number of posts value. 

**Blocks to update**
- [x] Posts block
- [x] Post Carousel block

paranoia1906:
Closes #1044 

postsBlockColumns

postsCarouselColumns

### DIFF
--- a/src/blocks/post-carousel/inspector.js
+++ b/src/blocks/post-carousel/inspector.js
@@ -34,14 +34,14 @@ class Inspector extends Component {
 			columns,
 		} = attributes;
 
-		const columnsOnChange = ( selectedColumns ) => {
+		const columnsCountOnChange = ( selectedColumns ) => {
 			const { postsToShow } = attributes;
 			setAttributes( { columns:
 				( selectedColumns > postsToShow ) ? postsToShow : selectedColumns,
 			} );
 		};
 
-		const postsOnChange = ( selectedPosts ) => {
+		const postsCountOnChange = ( selectedPosts ) => {
 			const { columns } = attributes;
 			const changedAttributes = { postsToShow: selectedPosts };
 			if ( columns > selectedPosts || ( selectedPosts === 1 && columns !== 1 ) ) {
@@ -85,7 +85,7 @@ class Inspector extends Component {
 					<RangeControl
 						label={ __( 'Columns', 'coblocks' ) }
 						value={ columns }
-						onChange={ ( value ) => columnsOnChange( value ) }
+						onChange={ ( value ) => columnsCountOnChange( value ) }
 						min={ 1 }
 						max={ Math.min( 4, postCount ) }
 						required
@@ -121,7 +121,7 @@ class Inspector extends Component {
 							<RangeControl
 								label={ __( 'Number of posts', 'coblocks' ) }
 								value={ postsToShow }
-								onChange={ ( value ) => postsOnChange( value ) }
+								onChange={ ( value ) => postsCountOnChange( value ) }
 								min={ 1 }
 								max={ 20 }
 							/>
--- a/src/blocks/post-carousel/inspector.js
+++ b/src/blocks/post-carousel/inspector.js
@@ -34,23 +34,20 @@ class Inspector extends Component {
 			columns,
 		} = attributes;
 
-		const columnCountByPosts = ( selectedColumns ) => {
+		const columnsOnChange = ( selectedColumns ) => {
 			const { postsToShow } = attributes;
 			setAttributes( { columns:
 				( selectedColumns > postsToShow ) ? postsToShow : selectedColumns,
 			} );
 		};
 
-		const postCountByColumn = ( selectedPosts ) => {
+		const postsOnChange = ( selectedPosts ) => {
 			const { columns } = attributes;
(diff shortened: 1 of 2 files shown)


src/blocks/post-carousel/inspector.js
```diff
@@ -34,6 +34,25 @@ class Inspector extends Component {
 			columns,
 		} = attributes;
 
+		const columnCountByPosts = ( selectedColumns ) => {
+			const { postsToShow } = attributes;
+			setAttributes( { columns:
+				( selectedColumns > postsToShow ) ? postsToShow : selectedColumns,
+			} );
+		};
```
jrtashjian: This function should be named in a way that describes what it does. Right now it implies that it will get the column count when in fact, this is setting the column count based on the "posts to show" value. Making this `setColumnCountByPosts` is more descriptive.

src/blocks/post-carousel/inspector.js
```diff
@@ -34,6 +34,25 @@ class Inspector extends Component {
...
+		const postCountByColumn = ( selectedPosts ) => {
+			const { columns } = attributes;
+			if ( columns > selectedPosts || ( selectedPosts === 1 && columns !== 1 ) ) {
+				setAttributes( {
+					columns: selectedPosts,
+					postsToShow: selectedPosts,
+				} );
+				return;
+			}
+			setAttributes( { postsToShow: selectedPosts } );
```
jrtashjian: This can be simplified a bit because there is a little duplication here. A new object type variable with the `postsToShow` key/value added would be good. The conditional can then push on the `columns` key/value if true. Then `setAttributes` is only called once.

Look at [`Object.assign()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) for "pushing" new keys onto an existing object.

src/blocks/post-carousel/inspector.js
```diff
@@ -69,10 +88,8 @@ class Inspector extends Component {
 					<RangeControl
 						label={ __( 'Columns', 'coblocks' ) }
 						value={ columns }
-						onChange={ ( value ) => {
-							setAttributes( { columns: value } );
-						} }
-						min={ 2 }
+						onChange={ ( value ) => columnCountByPosts( value ) }
+						min={ Math.min( 2, postCount ) }
```
jrtashjian: I don't think a minimum of `2` columns is necessary here. This might have been missed previously so let's fix it here. We have defaults for setting 1 or 2 columns based on the style if the value was not explicitly changed. One column is perfectly acceptable.

src/blocks/posts/inspector.js
```diff
@@ -74,6 +74,25 @@ const Inspector = props => {
...
+		if ( columns > selectedPosts || ( selectedPosts === 1 && columns !== 1 ) ) {
+			setAttributes( {
+				columns: selectedPosts,
+				postsToShow: selectedPosts,
+			} );
+			return;
+		}
+		setAttributes( { postsToShow: selectedPosts } );
+	};
+
```
jrtashjian: See my comments in the [`post-carousel/inspector.js`](https://github.com/godaddy/coblocks/pull/1084/files#diff-2e5f26fcb8648e88a57c0ffad902109b) file as they apply here as well.

src/blocks/posts/inspector.js
```diff
@@ -111,9 +130,9 @@ const Inspector = props => {
 					value={ columns }
 					onChange={ ( value ) => {
 						onUserModifiedColumn();
-						setAttributes( { columns: value } );
+						columnCountByPosts( value );
 					} }
-					min={ isStackedStyle ? 2 : 1 }
+					min={ isStackedStyle ? Math.min( 2, postCount ) : 1 }
```
jrtashjian: See my comments in the [`post-carousel/inspector.js`](https://github.com/godaddy/coblocks/pull/1084/files#diff-2e5f26fcb8648e88a57c0ffad902109b) file as they apply here as well.

src/blocks/post-carousel/inspector.js
paranoia1906: > I don't think a minimum of 2 columns is necessary here. This might have been missed previously so let's fix it here. We have defaults for setting 1 or 2 columns based on the style if the value was not explicitly changed. One column is perfectly acceptable.

I would appreciate a bit of clarification here. `Math.min( 2, postCount )` should be either 1 or 2 at lowest value depending on the style of posts-carousel (number of pooled posts). Is there a preferred way to accomplish this?

paranoia1906:
@jrtashjian The minimum is now properly set. Also handling case of style change and setting the column number accordingly. 

postsBlockColumnsControls

@richtabor This request handles the fix for maxing column count to post count on both the posts block as well as the post carousel block.

PR #1086, opened 2019-10-31
Customize Flickity to use CoBlocks Lightbox button style

### LINKED ISSUE #1047: ISBAT have one style of navigational arrows
The Carousel block navigation arrows and our lightbox navigation arrows should be styled in the same fashion. The lightbox navigation arrows are more modern, so the Carousel block should inherit those styles.

We should ensure that these can be easily styled by themes, so we should:

1. Reuse selectors to minimize duplicate styling
2. Implement utility classes where necessary

Carousel Arrows:
Screen Shot 2019-10-22 at 6 28 30 AM

Lightbox Arrows:
Screen Shot 2019-10-22 at 6 24 05 AM

AnthonyLedesma:
Closes #1047 
`flickity.min.js` is the file which is creating the navigational buttons for the Carousel block. Using the newest un-minified [source](https://unpkg.com/flickity@2.2.1/dist/flickity.pkgd.js) and modified code to use navigational styles from Lightbox. Used `npx grunt uglify` to generate the new minified file to be used for CoBlocks.
 
Edit function on line 2924.
Edit function on line 2964.

### DIFF
--- a/src/js/vendors/flickity.js
+++ b/src/js/vendors/flickity.js
@@ -1,5 +1,6 @@
+/* eslint-disable no-mixed-operators */
 /*!
- * Flickity PACKAGED v2.1.2
+ * Flickity PACKAGED v2.2.1
  * Touch, responsive, flickable carousels
  *
  * Licensed GPLv3 for open source use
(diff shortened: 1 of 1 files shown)


paranoia1906:
This file is generated when building the plugin. Users who would install the plugin or update would have these changes as deployed changes. As developers the file is not yet created/or the file still contains the old code `(/dist/js/vendors/flickity.min.js)`. Run `npx grunt uglify` to generate the new distributed flickity vendor files which are sourced from this change.

The arrows should reflect the correct styles.

carouselNavButtons

paranoia1906:
Note about this change. Styles used in editor and live site do not match. This is related to the fact that we are using distributed vendor js files to render the live site, as opposed to react component contained within node-modules and compiled into a source. A major issue with updating the styles in the editor is that we would have to either source the component within CoBlocks official source (perhaps as a utility) and modify the code OR we would need to edit the node-modules themselves to achieve style changes on the Flickity react component. 

One part of this problem is the actual methodologies used for how the Lightbox renders its buttons over how Flickity component renders its buttons. The change is not simply style but is also markup related. Flickity uses rendered SVG elements in the DOM whereas the lightbox uses a button and div with styles applied by classnames.

Another potential issue was investigated in https://github.com/godaddy/coblocks/pull/1026 regarding use of the Flickity react component and its current issue with passing focus to RichText. 

A move away from the React component would solve both of these issues right off. I am looking for opinions on the approach here.

src/js/vendors/flickity.js
```diff
@@ -9,5 +10,4305 @@
...
+
+		const style = getStyle( elem );
+
+		// if hidden, everything is 0
+		if ( style.display === 'none' ) {
+			return getZeroSize();
+		}
+
+		const size = {};
+		size.width = elem.offs
```
paranoia1906: Adding this comment for additional context. This change removes the SVG element which Flickity was using and instead uses div + button styled with the appropriate classes to achieve the correct style. Classes used are defined in lightbox utility.

https://github.com/godaddy/coblocks/blob/23c76198775064f1717d652ea71e4814c450dcfe/src/components/block-gallery/styles/style/_lightbox.scss#L64

paranoia1906: Adding this comment for additional context. This change removes the SVG element which Flickity was using and instead uses div + button styled with the appropriate classes to achieve the correct style. Classes used are defined in lightbox utility.

https://github.com/godaddy/coblocks/blob/23c76198775064f1717d652ea71e4814c450dcfe/src/components/block-gallery/styles/style/_lightbox.scss#L64

paranoia1906: Adding this comment for additional context. This change removes the SVG element which Flickity was using and instead uses div + button styled with the appropriate classes to achieve the correct style. Classes used are defined in lightbox utility.

https://github.com/godaddy/coblocks/blob/23c76198775064f1717d652ea71e4814c450dcfe/src/components/block-gallery/styles/style/_lightbox.scss#L64

paranoia1906: Adding this comment for additional context. This change removes the SVG element which Flickity was using and instead uses div + button styled with the appropriate classes to achieve the correct style. Classes used are defined in lightbox utility.

https://github.com/godaddy/coblocks/blob/23c76198775064f1717d652ea71e4814c450dcfe/src/components/block-gallery/styles/style/_lightbox.scss#L64

paranoia1906: Adding this comment for additional context. createDiv function added to Flickity to handle element generation.

AnthonyLedesma:
Closed in favor of https://github.com/godaddy-wordpress/coblocks/pull/1166

PR #1087, opened 2019-10-31
Enhance DimensionsControls with vw and vh units

### LINKED ISSUE #1054: Add vw and vh units to Block inspector
For more responsive layouts, being able to specify margin/padding in vh or vw units, as well as the current px, em and %, would be very helpful.

Adding  them to the const unitSizes = [] array appears to work with very easily...

/src/components/dimensions-control/index.js

const unitSizes = [
			//Append these...
			{
				name: _x( 'VH', 'A size unit for CSS markup', 'coblocks' ),
				unitValue: 'vh',
			},
			{
				name: _x( 'VW', 'A size unit for CSS markup', 'coblocks' ),
				unitValue: 'vw',
			},
		];

paranoia1906:
Closes #1054 

image

### DIFF
--- a/src/components/dimensions-control/index.js
+++ b/src/components/dimensions-control/index.js
@@ -394,12 +394,12 @@ class DimensionsControl extends Component {
 				unitValue: 'em',
 			},
 			{
-				name: _x( 'Viewport Height', 'A "vh" size unit for CSS markup', 'coblocks' ),
-				unitValue: 'vh',
-			},
-			{
 				name: _x( 'Viewport Width', 'A "vw" size unit for CSS markup', 'coblocks' ),
 				unitValue: 'vw',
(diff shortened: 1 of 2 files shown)


richtabor:
I don't think placing the units with this many selections is going to work UX wise here. Perhaps we need to tweak this where we're using a `select` somehow. 

I'm hesitant on complicating the interface by adding more visible top-level controls as such. :/

richtabor:
Added UI tweaks while we were here. 

Screen Shot 2019-11-04 at 10 16 55 AM

richtabor:
And here's a screenshot for < WP 5.3 and < Gutenberg 6.4. 

Screen Shot 2019-11-04 at 10 22 29 AM

PR #1093, opened 2019-11-06
Dynamic HR migration to core/separator

### LINKED ISSUE #1063: ISBAT use Dynamic HR block with TwentyTwenty theme
**Describe the bug**
Running TwentyTwenty on WP 5.3 RC2 the Dynamic HR block does not render correctly on the published page.

**To Reproduce**

1. Add Dynamic HR block.
2. Publish page and view as user.
3. See error

**Screenshots**
**Dynamic HR in editor**
image

**Dynamic HR on published page**
image

**Isolating the problem:**

**WordPress Version**
5.3 RC2

**Gutenberg Version**
N/A

AnthonyLedesma:
Closes #1063 

**Description**
The arrival of TwentyTwenty theme has revealed styling conflicts that are used with Dynamic HR block and the TwentyTwenty theme.

**Previous iterations**
Pull request #1068 investigated using styles to correct appearance of Dynamic HR block within twenty twenty.

Pull request #1079 considered a fix by inheriting the core Separator’s styling, so we wouldn’t need to continue supporting individual styles. It’s essentially a Separator + Spacer block combined and deprecating as needed.

The two previous requests did not seem to be the appropriate approach due to either complexity of style changes, or due to a shift away from using proper markup (as a work-around).

**Current changes**
This pull request is a shift away from the CoBlocks Dynamic HR block to the core/separator block with appropriate controls and deprecations.

**Notes:**
- Color controls have been removed from Dynamic HR yet the ability remains when selected the inner block `core/separator`. 
  * We can consider future enhancements to dispatch attributes to the inner blocks.
- Styles from index.js have been removed as styles are provided by the `core/separator` block.
  * Again we can consider future enhancements to dispatch classnames to the inner blocks.
- An eventual goal here would be to mirror the panel controls for style and color directly within the sidebar if the Dynamic HR block `isSelected`.

dynamicSeparatorMigration


paranoia1906:
I have one outstanding issue that that I am really struggling with. Hoping for some assistance.

**When using the `default` style**
Correct view in editor: 

image

Incorrect published view:
image

jrtashjian:
This is more complex than I think it needs to be. We should not need a deprecation or the innerBlocks component. Instead, approach this like we are removing the Dynamic Separator block entirely.

These are the steps I would take:
- [ ] Add a new style to the `core/separator` block with [`registerBlockStyle`](https://developer.wordpress.org/block-editor/developers/filters/block-filters/#block-style-variations)
- [ ] Filter the block settings through the [`block.registerBlockType`](https://developer.wordpress.org/block-editor/developers/filters/block-filters/#blocks-registerblocktype) to add a height attribute
- [ ] Add our height inspector control panel through the [`editor.BlockEdit`](https://github.com/godaddy/coblocks/pull/665/files#diff-7c3b28639720b37762013010416a6ab7)

You can see how I've accomplished this with filters when experimenting with abstracting the column controls here in #665: https://github.com/godaddy/coblocks/pull/665/files#diff-7c3b28639720b37762013010416a6ab7

- [ ] Remove all Dynamic Separator block code except for anything related to just registering the block (bare bones registration)
- [ ] Add a "raw" transformation to the Dynamic Separator block that looks for the old HTML and creates a `core/separator` block from it

Example of a raw transform in the Highlight block:
https://github.com/godaddy/coblocks/blob/master/src/blocks/highlight/transforms.js#L22-L34

I think the raw transform is a valid method of converting to the `core/separator` block in the most simple way. This last portion is where we need to experiment a bit to figure out the best way of removing a block from CoBlocks.

AnthonyLedesma:
Closed in favor of https://github.com/godaddy/coblocks/pull/1109

PR #1097, opened 2019-11-06
Remove CoBlocks getting started guide

AnthonyLedesma:
This pull request will remove the CoBlocks Getting Started guide from WordPress Admin. Also removes all assets and tests.


richtabor:
I believe we can now remove `lity.js` from `/vendors/` as well.

PR #1098, opened 2019-11-07
Ensure SVGs which provide additional context are accessible

AnthonyLedesma:
WordPress provides the `import { G, Path, SVG }` from `'@wordpress/components` drop-in replacement for the svg element that adds the required accessibility attributes for SVG elements across browsers. 

This pull request converts all SVGs to use the WordPress provided components and removes width and height attributes.


src/blocks/food-and-drinks/food-item/icons.js
```diff
@@ -10,9 +10,7 @@ const icons = {};
 
 icons.glutenFree = (
 	<SVG
-		height="14"
```
richtabor: Does removing these height/widths cause deprecation issues?

richtabor: We should test to confirm; I have a feeling it may.

richtabor:
@paranoia1906 Did you confirm that there were no visual regressions with the icons within the editor UI?

And could you resolve that conflict? Thanks!

src/blocks/food-and-drinks/food-item/icons.js
paranoia1906: Shoot, you are correct. Deprecation's needed all around.

paranoia1906:
@richtabor Blocks have been deprecated and I think I caught all visual regressions.

Issue #1081, opened 2019-10-28
Pricing Table: Block is missing block mover controls


kwight:
**Describe the bug**
The Pricing Table block shows no mover controls at normal width; at other widths, the controls appear, but are obscured with other UI.

**To Reproduce**

1. Create a new block editor post.
2. Insert the Pricing Table block.
3. Notice it has no mover controls.
4. Set it to Wide or Full display.
5. Notice the mover controls appear, but conflict with other UI.

**Screenshots**

<img width="844" alt="Screen Shot 2019-10-28 at 1 15 09 PM" src="https://user-images.githubusercontent.com/349751/67714464-319f4980-f985-11e9-8d95-17028e30f550.png">

<img width="862" alt="Screen Shot 2019-10-28 at 1 15 16 PM" src="https://user-images.githubusercontent.com/349751/67714489-3b28b180-f985-11e9-80a6-83aaee3070ef.png">

**Expected behavior**
I expect the block to have accessible mover controls, regardless of display width.

**Isolating the problem:**
Mark completed items with an [x].

-   [x] This bug happens with no other plugins activated
-   [x] This bug happens with a default WordPress theme active
-   [x] This bug happens **without** the Gutenberg plugin active
-   [x] I can reproduce this bug consistently using the steps above

**WordPress Version**
5.2.4

**Gutenberg Version**
n/a

PR #1100, opened 2019-11-07
Add block movers to Pricing Table Items

richtabor:
This PR adds block movers back to the Pricing Table Items, so you would be able to move them around inside the Pricing Tables parent block - closing #1081 

The reason they were disabled originally is because of the way we have the `count` selected. We needed to lock the InnerBlocks template so that we could manipulate the nested block count. I added `updateTables` (derived from the core Columns block) so that we could have both. 

I'm also using the `__experimentalMoverDirection` prop for `InnerBlocks`, which falls back to `vertical` without support. 

tables


jrtashjian:
I can't seem to see these new movers. I tried the latest version of WordPress with Gutenberg 6.8.0 as well as the nightly build of WordPress (5.4-alpha-46674). The `__experimentalMoverDirection` is being detected by React Inspector Tools in Chrome.

Screen Shot 2019-11-07 at 5 49 26 PM

Screen Shot 2019-11-07 at 5 45 35 PM

jrtashjian:
Looks like we have an odd behavior when removing the `templateLock`:

Kapture 2019-11-07 at 17 53 48

richtabor:
@jrtashjian are you sure you built the files?

richtabor:
Ah, it looks like `__experimentalMoverDirection` is not available in Gutenberg 6.8 or WP 5.3 just yet. I added back templateLock, but only `insert`, which removes the "Remove Block" control from that popover - but still maintains the switching table count capability.

If all is good on your end, I'm good with merging this and then coming back around later when `__experimentalMoverDirection` is officially supported - even though the movers will be vertical.

PR #1105, opened 2019-11-08
Row block: Fix - 2-column row not centring at alignfull in editor

broken8:
In various themes (plugins disabled) the 2-column row does not align properly in the editor when set to alignfull. 1, 3 and 4-column rows display acceptably in the editor but this fix also improves their positioning. The front-end displays correctly.

# Before
row-block-fullwidth-issue
# After
row-block-fullwidth-fixed
The issue appears to be fixed. Blocks placed within the row block display correctly on the front and back-end.


broken8:
I've updated my fork to 1.17.0 and at first sight I believe this back-end editor issue does not exist in this release. HOWEVER, using it on my test rig - no more than an hour after release - the front-end is not rendering properly. I think it's a front-end CSS build issue because in the editor everything renders properly. BUT the Posts block is empty when loaded. All options are zeroed and have to be selected from scratch. Hopefully, it's just me. I'll try again later with a fresh platform and watch this space.

AnthonyLedesma:
Hi @broken8, 

Thank you for testing this again and for your contribution. This morning a few requests came in regarding that Posts block as well as other server side rendered blocks. We have released 1.17.1 which should resolve the issue you are experiencing with the Posts block. Please create a new issue if you see any other bugs.

PR #1106, opened 2019-11-08
Use external Wikipedia CC licensed images in block examples

AnthonyLedesma:
This pull request uses images from https://commons.wikimedia.org/wiki/Main_Page just as core blocks have done. Used lower resolution images for the examples to help with load times.

examplesGalleryBlocks

image

image

image


jrtashjian:
Adding examples will override the style panel preview. So if we set example images in the Collage Gallery block, those images will show in the Styles panel preview rather than the ones uploaded. ([Related comment](https://github.com/godaddy/coblocks/pull/996#issuecomment-541865250))

@richtabor is this the behavior we want?

Screen Shot 2019-11-08 at 5 16 45 PM

richtabor:
I think it’s fine for galleries as it adds value to see what it actually looks like.

richtabor:
@jrtashjian Do you know if core is tracking this as a bug?

jrtashjian:
I'm not sure if they are but I was going to submit one if I couldn't find one. It's on my todo list. If we're ok with this behavior I approve.

jrtashjian:
I'll merge this after I deploy our next release today so it goes into the following release.

PR #1107, opened 2019-11-08
Add !important properties for the advanced top/bottom margin

richtabor:
This PR applies a margin-top and margin-bottom !important property to top-level `.wp-block` classes, where the Advanced spacing controls are adding `mb-0` and `mt-0` classes. 

We don't want our base utilities to override, so this PR addresses only `mb-0` and `mt-0` classes assigned to the top-level block div.

### DIFF
--- a/src/extensions/advanced-controls/styles/style.scss
+++ b/src/extensions/advanced-controls/styles/style.scss
@@ -10,3 +10,11 @@
 		}
 	}
 }
+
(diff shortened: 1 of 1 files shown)


AnthonyLedesma:
Pull request resolves recently received support request on WordPress.org.

https://wordpress.org/support/topic/cant-remove-spacing-between-rows/

PR #1123, opened 2019-11-12
Add top/bottom waves to core Cover block

richtabor:
This PR brings our custom waves styles from the Image block, to the Cover block. Once merged we could probably set the waves as variables and use them within both the Image and Cover block styling.

ScreenFlow


richtabor:
Does this same behavior happen on the core image block styles? That they’re available before an image is uploaded?

richtabor:
Does this same behavior happen on the core image block styles? That they’re available before an image is uploaded?

richtabor:
Good catch; fixed. 👍 ScreenFlow

PR #1128, opened 2019-11-13
Add Form block templates

AnthonyLedesma:
Adds pre-made templates to the Form block. 
`Contact`
`RSVP`
`Appointment`

**This PR resolves a bug in `master` branch**
formBlockSettingsUndefined

Bug is resolved by wrapping settings methods with 
```javascript
if ( typeof settings !== 'undefined' ){
[...]
}
```

**Deprecating a Form block**
If a Form block was created and configured prior to CoBlocks 1.17.1, the block would maintain its inner blocks but present a placeholder to choose a template. Added componentDidMount logic to allow previously configured blocks to continue working.

```javascript
	componentDidMount() {
		const { innerBlockCount, innerBlocks } = this.props;
		if ( innerBlockCount > 0 ) {
			this.setState( { template: innerBlocks } );
		}
               [...]
	}
```


jrtashjian:
Pushed a fix for the undo and redo editor functions. Prior to the fix if you selected a layout preset you would have to click "undo" twice to get back to the placeholder. This is because setting the layout attribute on the Form block would create an undo snapshot, and when the innerBlocks component was updated with the template another undo snapshot was created.

This is solved by subscribing to changes of the innerBlocks on the Form block and tracking the selected template in the Form block's state rather than as an attribute. Doing so allows the undo and redo editor functions to work as expected.

**Undo and Redo in action:**

Kapture 2019-11-22 at 18 20 37

jrtashjian:
@AnthonyLedesma is there a reason for saving the template as an attribute in this commit: a481862? It feels like we're unnecessarily polluting the `post_content`.

jrtashjian:
@AnthonyLedesma is there a reason for saving the template as an attribute in this commit: a481862? It feels like we're unnecessarily polluting the post_content.

AnthonyLedesma:
> @AnthonyLedesma is there a reason for saving the template as an attribute in this commit: 

@jrtashjian No. I thought it may be necessary but I found a more elegant solution while troubleshooting 5.2.4 support. I'll get that cleaned up and out of draft soon.

AnthonyLedesma:
Tested and working in WordPress 5.2.4 & 5.3.
Tested and working with Gutenberg 6.9, 7.0-RC1 and without Gutenberg.

PR #1135, opened 2019-11-15
ensure all __() functions use a text domain

AnthonyLedesma:
Four files in CoBlocks are using `__()` function for string translation but do not properly use the text-domain 'coblocks'. This PR adds the necessary domain for these functions.

### DIFF
--- a/src/components/background/panel.js
+++ b/src/components/background/panel.js
@@ -293,7 +293,7 @@ class BackgroundPanel extends Component {
 								}
 							} }
 						>
-							{ __( 'Remove ' + backgroundType ) }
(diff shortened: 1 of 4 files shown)


src/components/background/panel.js
```diff
@@ -293,7 +293,7 @@ class BackgroundPanel extends Component {
 								}
 							} }
 						>
-							{ __( 'Remove ' + backgroundType ) }
+							{ __( 'Remove ' + backgroundType, 'coblocks' ) }
```
fjarrett: @AnthonyLedesma we should be using a placeholder here

```js
sprintf(
    /* translators: %s: background type */
    __( 'Remove %s', 'coblocks' ),
    backgroundType
)
```

PR #1141, opened 2019-11-18
Add lightbox support to Collage block

AnthonyLedesma:
This enhancement will add a Lightbox toggle, script and cursor styles to the Collage block. 

collageBlockLightbox

### DIFF
--- a/includes/class-coblocks-block-assets.php
+++ b/includes/class-coblocks-block-assets.php
@@ -187,19 +187,11 @@ class CoBlocks_Block_Assets {
 		}
 
 		// Lightbox.
-		if ( has_block( $this->slug . '/gallery-masonry' ) || has_block( $this->slug . '/gallery-stacked' ) ) {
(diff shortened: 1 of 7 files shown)


includes/class-coblocks-block-assets.php
```diff
@@ -187,22 +187,14 @@ public function frontend_scripts() {
...
 				true
 			);
-
-			wp_enqueue_script(
-				$this->slug . '-masonry',
-				$dir . $this->slug . '-masonry' . COBLOCKS_ASSET_SUFFIX . '.js',
-				array( 'jquery', 'masonry', 'imagesloaded' ),
-				COBLOCKS_VERSION,
-				true
-			);
```
AnthonyLedesma: This seems to be a bug. Removed this snippet because it caused the masonry script to load even if masonry block is not in use. On line `150` the script is enqueued appropriately.

PR #1142, opened 2019-11-19
Update Gallery blocks to use core UX patterns

AnthonyLedesma:
This PR updates CoBlocks Galleries to use core UX patterns. Inspiration for changes are sourced from here: 

https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/gallery/edit.js#L271-L289


AnthonyLedesma:
Only works well with GB running. Putting a work in progress until I can figure out the work arounds.

AnthonyLedesma:
Pushed out some fixes for behavior **without** Gutenberg running. Now it works well with and without GB.

AnthonyLedesma:
@richtabor Tested and functional in 5.2.4 **with** and **without** GB running.

PR #1144, opened 2019-11-20
Introduce Offset Gallery block

AnthonyLedesma:
Offset Gallery block displays images in a dynamic grid format and has no limit on images. Allows standard gallery controls and has lightbox.

**Editor**
offsetGalleryIntroduction

**Published**
offsetGalleryIntroductionPublished

Note that this block is affected by Bug reported here: https://github.com/godaddy/coblocks/issues/1143


richtabor:
We'll need to add the relevant tweaks that we need to the existing gallery blocks in #1142 (if they're not already present).

AnthonyLedesma:
Using patterns defined in #1142. Ready for review.

src/components/block-gallery/styles/style/_captions.scss
```diff
@@ -1,5 +1,5 @@
 .has-caption-style-light {
-	.coblocks-gallery--item .coblocks-gallery--figure figcaption {
```
AnthonyLedesma: Removed class `.coblocks-gallery--figure` specification for support of Offset block. No visual regressions detected.

jrtashjian:
There is a weird behavior with breakpoints right at a width of 959px, 831px, and 832px.

Kapture 2019-11-27 at 14 56 10

richtabor:
I think that’s just because two images at those widths can’t display next to each other at those particular widths. Can you try with different images of different sizes?

AnthonyLedesma:
@richtabor @jrtashjian 
I have different behavior with different images.

offsetResponsive

richtabor:
That's what I thought would be the case. Different images will produce different marks where images will move onto the next "row".

PR #1146, opened 2019-11-20
Allow for multiple field types with same label

### LINKED ISSUE #1145: ISBAT select a date from a calendar and have it input to the appropriate input
**Describe the bug**
Users cannot add multiple date fields to a form block. When clicking the calendar on the second (or later) date field, the date gets added to the first date input field.

This is because all of the date input fields share the same ID.

**To Reproduce**

1. Add a form block to a page
2. Add two or more date fields to it
3. Try and set a date for date field 2 or 3
4. Notice how the selected date gets input to the first input

**Expected behavior**
I should be able to select a date from any of the date inputs in the form block and have the selected date be input to the appropriate input field.

**Isolating the problem:**

**WordPress Version**
5.3

EvanHerman:
Closes #1145  

This PR allows for fields to have multiple of the same field type with the same label. Previously, data was overwritten in the global `$_POST` when two or more of the same field types were used with identical field labels.

image

#### Additional Fixes:
- [x] Allow for multiple date fields to be used on a page and the proper date the be input to the correct text input field. (Updated the date field IDs)
- [x] Added `autocomplete="off"` on the date field to prevent autocomplete dropdown from covering the calendar

##### Submitted Data Before Patch
image

##### Submitted Data with Patch
image


richtabor:
@EvanHerman, looks like tests are failing here.

PR #1164, opened 2019-11-26
Generate translation files on master changes

EvanHerman:
Generate CoBlocks translation files in `i18n` job on CircleCI and upload them to the artifacts for the build when any changes to `master` are detected. This includes any PRs merged into `master`, or any hot fixes pushed directly to `master`.

https://circleci.com/gh/godaddy-wordpress/coblocks/6319#artifacts/containers/0

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,7 +156,9 @@ jobs:
             npm run json2po
             npm run po2mo
             npm run po2jed
+            grunt compress:coblocks
             mv languages/coblocks-*.json /tmp/artifacts
+            mv coblocks-translations.zip /tmp/artifacts
       - store_artifacts:
           path: /tmp/artifacts
           destination: coblocks-translations
@@ -199,7 +201,7 @@ jobs:
           command: |
             mkdir -p /tmp/artifacts
             grunt build
-            grunt compress
+            grunt compress:coblocks
             mv build/*.zip /tmp/artifacts
       - deploy:
           name: Deploy the plugin
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,9 +17,9 @@ workflows:
             tags:
               only: /.*/
       - i18n:
-          branches:
-            only:
-              - master
+          filters:
+            branches:
+              only: master
       - deploy:
           requires:
             - php56-build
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -124,6 +128,42 @@ jobs:
...
+          command: sudo apt-get install gettext
+      - run:
+          name: Install node packages
+          command: npm install
+      - run:
+          name: Build the plugin
+          command: |
+            mkdir -p /tmp/artifacts
+            npm run babel:build
+            npm run pot2json
```
fjarrett: @EvanHerman These lines can be removed. The `babel:build` script already does `makepot` which calls `pot2json`, and the other scripts are for dev use only. Basically all this job has to do is ensure two files are updated with every `master` commit: `coblocks.pot` and `coblocks.json`.

.circleci/config.yml
```diff
@@ -124,6 +128,42 @@ jobs:
...
+            npm run json2po
+            npm run po2mo
+            npm run po2jed
+            grunt compress:coblocks
+            mv languages/coblocks-*.json /tmp/artifacts
+            mv coblocks-translations.zip /tmp/artifacts
```
fjarrett: @EvanHerman what is the `coblocks-translations.zip` actually for?

.circleci/config.yml
fjarrett: Lines 156-159 can be removed

.circleci/config.yml
```diff
@@ -124,6 +128,42 @@ jobs:
...
+          command: npm install
+      - run:
+          name: Build the plugin
+          command: |
+            mkdir -p /tmp/artifacts
+            npm run babel:build
+            npm run pot2json
+            npm run json2po
+            npm run po2mo
+            npm run po2jed
```
fjarrett: @EvanHerman These lines can be removed. The babel:build script already does makepot which calls pot2json, and the other scripts are for dev use only. Basically all this job has to do is ensure two files are updated with every master commit: coblocks.pot and coblocks.json.

fjarrett:
@EvanHerman I think what we really want here is for the `coblocks.pot` and `coblocks.json` files to be regenerated and just committed back to the `master` branch, no artifacts needed

.circleci/config.yml
EvanHerman: Sorry, I misunderstood. I thought we wanted to generate translation files for the translation team to test with on each commit. PR has been updated to reflect the actual needed changes.

PR #1166, opened 2019-11-27
Use consistent styling for the next and previous buttons in the Carousel block and Lightbox

### LINKED ISSUE #1047: ISBAT have one style of navigational arrows

AnthonyLedesma:
Closes #1047

Custom jQuery script to replace Flickity nav buttons on both front and back end.

To test this locally you must first compile the script by running `npx grunt uglify`.

image


AnthonyLedesma:
Made flickity script a dependency for the coblocks-flickity-buttons script. Improves load speed and appearance especially on the front-end.

PR #1168, opened 2019-12-03
Remove hard coded field-email in form block reply to headers

EvanHerman:
Allows for the email field label to be updated while still setting the proper reply to headers in the email that is sent.


richtabor:
@AnthonyLedesma Does this change affect forms already added to pages?

AnthonyLedesma:
@richtabor Testing shows that this does not negatively affect already-setup form blocks. This is one great thing about server-rendered blocks. 😃

AnthonyLedesma:
Changes merged in to master from here:
https://github.com/godaddy-wordpress/coblocks/pull/1017

PR #1178, opened 2019-12-06
Serve full size images with Post Carousel block

AnthonyLedesma:
The editor provides full-size images to preview the block but published pages use the thumbnail-sized imaged. This change will bring in full-size images to be used on published pages. The default behavior of `get_the_post_thumbnail_url()` is to retrieve the thumbnail size image for the provided post ID. Here we can pass the string 'full' to retrieve full-size images. 

**Inspired with WordPress.org support request**
https://wordpress.org/support/topic/featured-images-in-post-carousel-low-res/

### DIFF
--- a/src/blocks/post-carousel/index.php
+++ b/src/blocks/post-carousel/index.php
@@ -241,7 +241,7 @@ function coblocks_get_post_carousel_info( $posts ) {
 
 		$formatted_post = null;
 
-		$formatted_post['thumbnailURL'] = get_the_post_thumbnail_url( $post );
+		$formatted_post['thumbnailURL'] = get_the_post_thumbnail_url( $post, 'full' );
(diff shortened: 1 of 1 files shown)


richtabor:
> Here we can pass the string 'full' to retrieve full-size images.

Is `full` the standard size, or is that from the theme's declared sizes?

AnthonyLedesma:
The acceptable paramers do not seem to be well documented, but I was able to find this list looking through source code.

```
 * The `$size` parameter accepts either an array or a string. The supported string
 * values are 'thumb' or 'thumbnail' for the given thumbnail size or defaults at
 * 128 width and 96 height in pixels. Also supported for the string value is
 * 'medium', 'medium_large' and 'full'. The 'full' isn't actually supported, but any value other
 * than the supported will result in the content_width size or 500 if that is
 * not set.
```

PR #1208, opened 2019-12-18
Fix media placeholder visibility when using Spotlight mode

AnthonyLedesma:
When a block is selected in the typical editor view, an `is-selected` class is appended to the block. In the case that Spotlight mode is enabled, the class `is-focused` is appended instead. This change adds the correct conditional to hide the placeholder. 

galleryBlocksHiddenPlaceholder

galleryBlocksHiddenPlaceholderFix


AnthonyLedesma:
@jrtashjian I'd love to see this one get included in the next release today if we have the time for review.

PR #1225, opened 2019-12-20
Break each icon keyword into its own i18n wrapper

fjarrett:
1. Making this one long string is really confusing to translators.
2. In most languages each keyword translated into multiple words, breaking the space-separated strategy altogether.
3. Individual strings are easier to maintain (add/remove) in source control than one mega line.


fjarrett:
Ugh. I should have totally taken this as an opportunity to alphabetize the keywords. I will do that before we merge.

PR #1242, opened 2020-01-09
Fix CoBlocks for Gutenberg 7.2.0: replace legacy editor- classes with block-editor-

mmtr:
Gutenberg 7.2.0 has removed the legacy `editor-` classes in https://github.com/WordPress/gutenberg/pull/19050 breaking the CoBlocks styles. This PR updates all those classes to use the `block-editor-` prefix.

**Screenshots**

| Before | After |
|---|---|
| Screenshot 2020-01-09 at 14 35 24 | Screenshot 2020-01-09 at 14 34 06 |
| Screenshot 2020-01-09 at 14 35 29 | Screenshot 2020-01-09 at 14 34 10 |
| Screenshot 2020-01-09 at 14 35 33 | Screenshot 2020-01-09 at 14 34 14 | 

**How to test**
- Activate Gutenberg 7.2.0
- Generate a new build of CoBlocks using this branch and activate it.
- Verify the blocks look correct.


gwwar:
@mmtr if CoBlocks supports multiple Block Editor versions, we'll want to have additive css selectors instead of updating them. Eg styling would support both `.editor` and `.block-editor`

@aduth did we have a migration warning for this? Other block builders might run into this.

aduth:
[It is marked](https://github.com/WordPress/gutenberg/pull/19050) as "Needs Dev Note", which by the [backwards compatibility guidelines](https://github.com/WordPress/gutenberg/blob/master/docs/designers-developers/developers/backward-compatibility/README.md) is meant as an indication of needing to publish an associated [`dev-notes`-tagged post on the Make/Core blog](https://make.wordpress.org/core/tag/dev-notes/).

In other words: It is expected that a migration guide would be authored as part of an upcoming major WordPress release in which it's expected to land.

As to how this relates to interoperability between the Gutenberg plugin and other plugins... I'm not sure there are guidelines or expectations around this. The Gutenberg plugin is tended to be considered as being more unstable in its changing APIs, but even then it's typically tended to have [some deprecation period](https://developer.wordpress.org/block-editor/developers/backward-compatibility/deprecations/) (at least for breaking API changes).

I can see how this might be an area for workflow improvement. I might imagine it could be something where the dev-notes associated with a "Needs Dev Notes" pull request should be authored at the time of the pull request, and should ideally be shared as part of a plugin release post and/or the tagging of the plugin release candidate. Third-party integrators may also consider using "Needs Dev Note" as a [label filter](https://github.com/WordPress/gutenberg/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3A%22Needs+Dev+Note%22+) for potential future compatibility issues.

aduth:
There was also this note published for dev-notes of WordPress 5.2:

>
>The components that moved from the editor module to the block-editor module are internally using CSS class names that follow the package they’re declared in. For example, the `editor-inserter__toggle` class name is now renamed `block-editor-inserter__toggle`. The old class names have been retained to minimize any backward compatibility concern, but the CSS styles are now targeting the new class names. **Ideally, you should rely on components in your own code instead of the internal class names used in the WordPress admin. But if you do use those classes, make sure to rename them accordingly.**

https://make.wordpress.org/core/2019/04/09/the-block-editor-javascript-module-in-5-2/

aduth:
Worth noting there are a few more of these set to land as part of https://github.com/WordPress/gutenberg/pull/19489 . It shouldn't be a problem if this pull request captured all of those legacy class names.

richtabor:
Appreciate it @aduth, we'll double-check against that PR. Thanks!

PR #1250, opened 2020-01-10
Fix custom links added to images in the Offset Gallery block redux

### LINKED ISSUE #1227: Offset Gallery: Adding urls to images breaks block once editor page gets reloaded
## Prerequisites

-   [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate
-   [x] The issue still exists against the latest `master` branch of CoBlocks on Github (this is **not** the same version as on WordPress.org!)
-   [x] I have attempted to find the simplest possible steps to reproduce the issue

## Steps to reproduce the issue

1. Add "Offset Gallery" block to a page
2. Add some images to the new Offset Gallery
3. Set "Link to" to "Custom URL"
4. Add url to at least one image
5. Save
6. Reload the website

## Expected/actual behavior

Wordpress shows an error that the block got corrupted. It's possible to repair it/revert it to a block or transform it to HTML. In best case the gallery is still there but added urls are gone.

Issue doesn't occur when no urls are added to any images in the Offset Gallery.

Expected behavior:
Offset Gallerys can be saved with urls added to single images and offset gallery being able to be edited even after reloading the editor.

## Isolating the problem

-   [x] This bug happens with only CoBlocks plugin active

## WordPress Environment

<details>
Wordpress 5.3.2
</details>

AnthonyLedesma:
Closes #1227.

Shared gallery attributes were not being correctly cloned here: https://github.com/godaddy-wordpress/coblocks/pull/1239. 

The previous PR resulted in an improper pointer in memory from the attempted cloned attributes. The code had inadvertently altered shared attributes instead of providing a new set of values to deprecate with.  

Implemented changes suggested by @jrtashjian.
https://github.com/godaddy-wordpress/coblocks/pull/1250#discussion_r365383916

<del>Old code
const deprecatedAttributes = { ...GalleryAttributes, ...metadata.attributes };
Object.assign( deprecatedAttributes.images.query, {	imgLink: { source: 'attribute', selector: 'a', attribute: 'href' } } );</del>

<del>New code - This PR
const clonedAttributes = JSON.parse( JSON.stringify( GalleryAttributes ) );
Object.assign( clonedAttributes.images.query.imgLink, { source: 'attribute', selector: 'a', attribute: 'href' } );
</del>


src/blocks/gallery-offset/deprecated.js
```diff
@@ -0,0 +1,119 @@
...
+					);
+				} ) }
+			</ul>
+		</div>
+	);
+};
+
+//prepare attributes
+const clonedAttributes = JSON.parse( JSON.stringify( GalleryAttributes ) );
+Object.assign( clonedAttributes.images.query.imgLink, { source: 'attribute', selector: 'a', attribute: 'href' } );
```
jrtashjian: This is smart but we will need to change how we do this in the future if we need to add a second deprecation for this block. This doesn't have any context to what deprecation this is part of but with only one deprecation this is fine for now.

src/blocks/gallery-offset/deprecated.js
```diff
@@ -0,0 +1,119 @@
...
+
+const deprecated = [
+	{
+		attributes: { ...clonedAttributes, ...metadata.attributes },
+		save: deprecatedCustomImageLinks,
+	},
```
jrtashjian: To provide better context on the attributes we are changing for the assigned deprecation, I've used the spread operator in the assignment (without creating a separate variable). This works exactly the same without modifying the original object.

```
{
	attributes: {
		...GalleryAttributes,
		...metadata.attributes,
		images: {
			...GalleryAttributes.images,
			query: {
				...GalleryAttributes.images.query,
				imgLink: {
					source: 'attribute',
					selector: 'a',
					attribute: 'href',
				},
			},
		},
	},
	save: deprecatedCustomImageLinks,
}
```

It’s a little longer but it also has the context of being associated with a specific deprecation which is what we want.

AnthonyLedesma: I like this better. Implementing now.

PR #1263, opened 2020-01-15
Bundle languages in the canary package

fjarrett:
* Bundle all language files with the canary package (POT/PO/MO/JED)
* Remove any GlotPress translations from the i18n staging site after deploy
* Append the current commit hash to the end of canary versions
* Add disclaimer to the canary description that is should not be used in production
* Once merged, we can remove the `STAGING_SITE_PLUGIN_PATH` env var from CircleCI

Screenshot_2020-01-15_16_58_45
Screenshot_2020-01-15_16_59_56

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -344,6 +344,9 @@ jobs:
           command: |
             mkdir -p /tmp/artifacts
             grunt build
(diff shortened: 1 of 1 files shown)


fjarrett:
This PR ended up requiring a hotfix for the `sed` line. Done in https://github.com/godaddy-wordpress/coblocks/commit/c39d2c799134aa6488349d8d3fbc5e3e777d31b1

PR #1288, opened 2020-01-21
Improve Gist block stability through E2E testing

AnthonyLedesma:
New E2E tests for Gist block.

Test changes using
```bash
 npx cypress run --config video=false --browser chrome --spec '**/gist.cypress.js,'
```
```javascript
  Running:  src/blocks/gist/test/gist.cypress.js                                            (1 of 1)


  Test CoBlocks gist Block
    ✓ Test gist block saves with color values set. (25883ms)
    ✓ Test gist block saves with url. (12036ms)
    ✓ Test gist block saves with custom classes. (13416ms)


  3 passing (54s)
```


AnthonyLedesma:
From testing local it does not seem that a gist deprecation is needed in this case.

PR #1308, opened 2020-01-27
Remove one column option from the Row block

richtabor:
This PR removes the one column option from our Row block to make things less confusing. 

Current: 
Screen Shot 2020-01-27 at 12 24 29 PM

### DIFF
--- a/src/blocks/row/edit.js
+++ b/src/blocks/row/edit.js
@@ -91,7 +91,6 @@ class Edit extends Component {
 		);
 
 		const columnOptions = [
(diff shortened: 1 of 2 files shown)


DustinWStout:
I'm disappointed with this decision... I had clear purpose for using one column rows! And I use them in every single layout I've built with CoBlocks.

aj-adl:
Seconded, Ive just upgraded and am sorely missing this feature. the built-in groups block lacks the padding controls / colour controls, so it can be hard to keep the wrappers and padding consistent. using 1col row blocks was the strategy for this so far (using Go theme)

richtabor:
Thanks for the feedback champs. We'll revert this. 👍

PR #1328, opened 2020-02-08
Optimize CircleCI Workflow

jrtashjian:
### TODO

- [x] Run the build process at the beginning of our workflow and share the built plugin with the rest of the workflow. Right now we are running a new build on each job.
- [x] Make CSS, JS, and PHP linting processes part of the workflow.
- [x] Look into utilizing Yarn for speedy dependency installs.
- [x] Rename jobs and steps for readability.
- [ ] ~Rework Cypress parallelism so that we ensure we're processing tests in parallel.~
- [x] Utilize CircleCi caching to improve install speed.
- [x] Run linting only on changed files, if any.
- [x] Run tests only on changed files, if any.
- [x] Run tests on all files when running on the `master` or `develop` branches.

## NPM vs Yarn
I tested `npm install` vs `yarn install` and yarn is insanely faster (and pre-installed on the CircleCI image). This PR moves over to utilizing yarn for dependency management and is interchangeable with NPM. All of the normal NPM commands will continue to work but we will be persisting a `yarn.lock` file over a `package-lock.json` file. Both Yarn and NPM utilize the same `package.json` file.

#### Cold install (no cache)
`npm install`: ~2m 15s
`yarn install`: ~1m 13s

#### Warm install (using previous cache):
`npm install`: 46s (33s restoring cache, 13s for install)
`yarn install`: 15s (14s restoring cache, <1s for install)

## Finished! What has changed?

Our CircleCI config, job paths, processes, etc have all been optimized to improve the speed of our CI/CD feedback loop. Simple reorganizing of build tasks early in the workflow massively improved speed of each job as we were building the plugin at each job previously.

1. The first job in the workflow is the "build" job and is responsible for updating cached dependencies, compiling scripts and styles, and persisting files to the workspace for the other jobs to use.

2. The linting jobs (`lint_css`, `lint_js`, `lint_php`) perform our code quality/formatting checks prior to unit test running. These jobs short-circuit the workflow if there are any failures to prevent provisioning unecessary machines. _These jobs will only check files changed in the branch unless ran against the master branch, in which case all files are checked._

3. The unit testing jobs (`unit_testing_js`, `unit_testing_php_(56|73)`) run our unit tests prior to the end-to-end testing jobs. These jobs short-circuit the workflow if there are any failures to prevent provisioning unecessary machines. _These jobs will only test files changed in the branch unless ran against the master branch, in which case all files are tested._

[...]

PR #1336, opened 2020-02-10
Run composer install and switch path to WPCLI in e2e build

EvanHerman:


EvanHerman:
This should actually be done in https://github.com/godaddy-wordpress/coblocks/pull/1328

Closing this PR so it can be updated there.

PR #1344, opened 2020-02-11
Tweak attributes for consistency

richtabor:
This PR adds a bit more consistency across how attributes are defined. 

For example, taking `attributes.displayPostContent` format - that we've used in a few blocks - to:

```
const {
displayPostContent,
} = attributes;

{ displayPostContent }
```

Blocks I've worked on and tested here: 
- [x] Features
- [x] Events
- [x] Food & Drinks
- [x] Logos
- [x] Posts
- [x] Services
- [x] Row


jrtashjian:
I'm fine with merging this but here's a couple reasons why we might want to use the `attributes.displayPostContent` over deconstructing the variable like this:

```
const {
displayPostContent,
} = attributes;

{ displayPostContent }
```

1. It's obvious from a readability standpoint that we're using the value directly from the parsed block attributes.

2. New eslint rules included with `wp-scripts` has [`no-shadow`](https://eslint.org/docs/rules/no-shadow) enabled which "disallow variable declarations from shadowing variables declared in the outer scope". Basically deconstructing this variable will prevent us from using a nice shorthand within out callback functions.

**Example of deconstructing the variable as well as satisfying eslint:**

```
const { contentAlign } = this.props.attributes;

<AlignmentToolbar
    value={ contentAlign }
    onChange={ ( nextContentAlign ) => setAttributes( { contentAlign: nextContentAlign } ) }
/>
```

**Example using `attributes.{attribute-name}` as well as satisfying eslint:**

```
const { attributes } = this.props;

<AlignmentToolbar
    value={ attributes.contentAlign }
    onChange={ ( contentAlign ) => setAttributes( { contentAlign } ) }
/>
```

I would prefer we move towards the second method. They might not all be exactly the same immediately, but over time (or someone with time on their hands) they will be updated as eslint is run on changed files.

Regardless, we can merge this and make updates later.

richtabor:
Sounds good. I'm all for consistency and following whatever standards core/gutenberg is using. 👍

PR #1357, opened 2020-02-13
Introduce Jest tests [Phase 3] 

AnthonyLedesma:
Introduce Jest phase-3 snapshot/save tests for the following blocks:

- [x] Gif
- [x] Gist
- [x] Hero
- [x] Logos
- [x] Map
- [x] Media Card
- [x] Post Carousel
- [x] Posts

Closes: #868, #869, #870, #865, #866


AnthonyLedesma:
@EvanHerman, This branch has a failing test regarding post counts. This attribute test requires at least 3 posts to display. Not sure why this did not catch earlier. Should we consider a helper to generate a few sample posts for the E2E tests?

AnthonyLedesma:
@EvanHerman 
We could also add a script like this to the E2E env: https://developer.wordpress.org/cli/commands/post/generate/

Issue #1361, opened 2020-02-17
ISBAT use coblocks without jQuery being required by default. (Events block issue)


stephenheron:
### Describe the bug:
Hi, 

I updated coblocks to the latest version (1.21.0) and I noticed that now jQuery appears in the source even when no blocks are being used. Using the previous version that we were using (1.15.0) this was not the case.

I narrowed down the issue to Events block enqueuing 'coblock-slick' (https://github.com/godaddy-wordpress/coblocks/blob/master/src/blocks/events/index.php#L139).

If we compare this to the Posts Carousel block (https://github.com/godaddy-wordpress/coblocks/blob/master/src/blocks/post-carousel/index.php#L322) which also uses Slick, it instead uses 'wp_register_script' and 'editor_script' so it is only included for the editor. 

Interestingly in the [class-coblocks.block.assets.php](https://github.com/godaddy-wordpress/coblocks/blob/master/includes/class-coblocks-block-assets.php#L299) it seems it looks for the events block and enqueues the scripts. 

### To reproduce:
Install coblocks 1.21.0.
View source on page not using any blocks.
jQuery is being included.

### Expected behavior:
Install coblocks 1.21.0.
View source on page not using any blocks.
jQuery is not being included.

josephscott:
@jrtashjian this appears to be coming from this change: https://github.com/godaddy-wordpress/coblocks/commit/950cdb9b36714982c43bc1692536d33877490d9d#diff-2c58a1e4d8686ecb878210a36c71d29dR139

As noted above, CoBlocks now adds extra JS and CSS on every page view, just by activating the plugin.  These include:

- jQuery ( 33 KB )
- jQuery Migrate ( 4.3 KB )
- slick.js ( 10.9 KB )
- coblocks-style.css ( 19.6 KB )

That is an extra 67.8 KB over the wire for every page view, just by activating the CoBlocks plugin.

While #1365 is a good long term solution, not loading these extra resources on every page view would be a big help right now.

PR #1379, opened 2020-02-25
Stop unconditionally enqueuing slick.js

brandonpayton:
### Description
Currently, the events block unconditionally enqueues slick.js. It is enqueued regardless of whether it is needed.

This PR updates coblocks so slick.js is only enqueued for the editor and for front-end pages that include the events block.

Because slick.js depends on jQuery, this may address #1361, but I haven't been able to confirm yet because something else on my quite simple site is also enqueuing jQuery by default.

### Types of changes

Bug fix

### How has this been tested?

Setup:
Open browser dev tools to the network tab and search for slick.js

Testing:
1. On a test site, navigate to a page that does not contain the events block, and observe that slick.js is not requested.
2. Open the block editor to create a new post and note that slick.js is requested.
3. Add an events block, link to an external calendar with more than 5 events, and note that the event pages are rendered in a carousel.
4. Publish the new post, navigate to it, and note that slick.js is requested and that the events are rendered in a functioning carousel.

If a negative test is desired:
1. Comment out the line `add_action( 'enqueue_block_editor_assets', array( $this, 'editor_scripts' ) );`
2. Open the block editor to create a new post and note that slick.js is not requested.
3. Add an events block, and link to an external calendar with more than 5 events, and note that the events are not displayed in a carousel.

### Checklist:
- [x] My code is tested

### DIFF
--- a/includes/class-coblocks-block-assets.php
+++ b/includes/class-coblocks-block-assets.php
@@ -40,6 +40,7 @@ class CoBlocks_Block_Assets {
 	public function __construct() {
 		add_action( 'enqueue_block_assets', array( $this, 'block_assets' ) );
 		add_action( 'init', array( $this, 'editor_assets' ) );
+		add_action( 'enqueue_block_editor_assets', array( $this, 'editor_scripts' ) );
 		add_action( 'wp_enqueue_scripts', array( $this, 'frontend_scripts' ) );
 		add_action( 'the_post', array( $this, 'frontend_scripts' ) );
 	}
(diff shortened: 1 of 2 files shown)


josephscott:
@brandonpayton confirmed that jQuery and friends no longer load with this version.  However, the request for coblocks-style.css is still happening.

josephscott:
I tried the plugin with this method returning early to confirm it is the place triggering the load: https://github.com/godaddy-wordpress/coblocks/blob/master/includes/class-coblocks-block-assets.php#L71

It looks like the coblocks-style.css file contains styles for multiple blocks.  Normally I'd think that `has_block()` would be good here, but I guess we'd end up having to check against every block that CoBlocks registers.

PR #1385, opened 2020-02-28
Introduce 'Replace Image' CoBlocks extension

AnthonyLedesma:
### Description

Introduce a new extension which adds a 'Replace Image' button to the `core/image` block. This button opens directly to the media library. 

### Screenshots
ReplaceImageWithGutenberg

### Types of changes

Introduces filtered component for the `core/image` block allowing a user to easily swap images.

### How has this been tested?

Tested with and without Gutenberg plugin active.
Tested on WordPress 5.4beta2 and 5.3.2.
New test for extension added. Passes tests.

### Checklist:
- [x] My code is tested


.dev/tests/cypress/helpers.js
```diff
@@ -115,6 +117,40 @@ export function addCoBlocksBlockToPage( clearEditor = true, blockID = '' ) {
...
+/**
+ * From inside the WordPress editor open the core Gutenberg editor panel
+ *
+ * @param bool clearEditor Whether or not to clear all blocks on the page before
+ * adding a new block to the page.
+ * @param clearEditor
+ * @param blockID
+ * @param string blockID     ID to check for in the DOM.
+ */
+export function addCoreBlockToPage( clearEditor = true, blockID = '' ) {
```
jrtashjian: We now have two similar functions for adding a block to the page. My recommendation is that we combine these into a single `addBlockToPost` and update all calls to use this new function. We should also reorder the parameters and keep our naming consistent.

**Why `addBlockToPost` vs `addBlockToPage`?**
WordPress considers all pages as a post with the `post_type` of `page`.

**Change parameter `blockID` to `blockName` and make it the first parameter**
When registering a block, the namespaced slug is called the block name so we're just keeping it consistent here so there is no potential confusion. We want to make it the first parameter because it is absolutely required to add a block to the page. We also want to remove the default `= ''` bit so that we fail if a name isn't passed. `clearEditor` isn't required so it should be last. I would also set the default value of `clearEditor` to `false` because the function name is `addBlock` not `replaceBlocksWith`.

.dev/tests/cypress/helpers.js
```diff
@@ -115,6 +117,40 @@ export function addCoBlocksBlockToPage( clearEditor = true, blockID = '' ) {
...
+	if ( clearEditor ) {
+		clearBlocks();
+	}
+
+	if ( ! blockID.length ) {
+		return;
+	}
+	console.log( 'should be running' );
```
jrtashjian: Remove call to `console`.

.dev/tests/cypress/helpers.js
jrtashjian: And to be clear with the `blockName` change, we would now be passing the full namespaced block name such as `core/image` or `coblocks/alert` to this function rather than just the last portion.

.dev/tests/cypress/helpers.js
```diff
@@ -131,11 +167,13 @@ export function savePage() {
...
- *               retreive the block from the spec file.
- *               eg: accordion => div[data-type="coblocks/accordion"]
+ * Note: If no blockID is specified, getBlockSlug() attempts to
+ * retreive the block from the spec file.
+ * eg: accordion => div[data-type="coblocks/accordion"]
+ * @param blockID
+ * @param slug
  */
-export function checkForBlockErrors( blockID = '' ) {
+export function checkForBlockErrors( blockID = '', slug = 'coblocks' ) {
```
jrtashjian: Again, let's change `blockId` to `blockName` and pass the full namespaced block name to this function. Remove the new parameter.

.eslintignore
```diff
@@ -1,6 +1,8 @@
 build
 dist
 node_modules
-test
 vendor
 wordpress
+*.spec.js
+*.spec.js.snap
+!*/helpers.js
```
jrtashjian: Because we don't know how our `test/` directories could evolve in the future we want to ignore them completely. If we want to run linting on the test files, use the negation syntax like you did for the helpers.js file to pick enable linting on `*.spec.js` and `*.cypress.js` files.

src/extensions/remove-image/remove-image.cypress.js
```diff
@@ -0,0 +1,44 @@
...
+		cy.get( '.replace-image-button' ).should( 'not.exist' );
+
+		cy.fixture( pathToFixtures + fileName, 'base64' ).then( ( fileContent ) => {
+			cy.get( 'div[data-type="core/image"]' )
+				.find( 'div.components-drop-zone' ).first()
+				.upload(
+					{ fileContent, fileName, mimeType: 'image/png' },
+					{ subjectType: 'drag-n-drop', force: true, events: [ 'dragstart', 'dragover', 'drop' ] },
+				)
+				.wait( 2000 ); // Allow upload to finish.
```
jrtashjian: From what I've read about the behavior of Cypress, we don't need to wait here arbitrarily. Cypress should be waiting by default for the upload to finish and move on as soon as it's done. I think you would use `.then()` to perform an action or assertion after upload if you need to.

.dev/tests/cypress/helpers.js
AnthonyLedesma: Refactored these changes. Need to resolve a few of these tests then I'll ping you for review again.

richtabor:
@AnthonyLedesma is there a way to filter the actual `InspectorControls` so that the button may be applied _within_ the Image settings panel? Here's what happens if the panel is closed: 

Screen Shot 2020-03-05 at 1 27 39 PM

AnthonyLedesma:
@richtabor This was as close as I could get to appearing within the image settings panel.
replaceImageVisibility

Tested and confirmed working in 5.4RC1 and 5.3.2 with and without the Gutenberg plugin.

PR #1390, opened 2020-03-02
wp_json_encode needs an integer

szepeviktor:
currently true is cast to 1

Did you mean `JSON_HEX_TAG` which has the value of 1?

Discovered by @phpstan


jrtashjian:
We don't need to define anything here so I'm going to remove the passed parameter entirely.

szepeviktor:
All right.

PR #1397, opened 2020-03-03
Migrate the Form block's submit button to a child block

EvanHerman:
- Move submit button in as a component of the innerblocks in the form block


richtabor:
@jrtashjian We're still getting deprecation errors here and could use a hand. 🙏

src/blocks/form/edit.js
```diff
@@ -279,17 +273,22 @@ class FormEdit extends Component {
...
 			if ( isEqual( elem.template, layout ) ) {
+				// // Update the child block's attributes
 				submitButtonText = elem.submitButtonText;
 				if ( Array.isArray( submitButtonText ) ) {
 					submitButtonText = submitButtonText.join( '' );
 				}
+				setTimeout( function() {
+					let childBlocks = wp.data.select( 'core/editor' ).getBlocksByClientId( clientID )[0].innerBlocks;
+					wp.data.dispatch('core/editor').updateBlockAttributes( childBlocks[ childBlocks.length - 1 ].clientId, { submitButtonText: submitButtonText } );
+				}, 100 );
```
jrtashjian: We should be able to detect when changes are made to the template using `componentDidUpdate` to then modify inner blocks. We shouldn't wait an arbitrary time when we know exactly when something has updated.

src/blocks/form/edit.js
```diff
@@ -279,17 +273,22 @@ class FormEdit extends Component {
...
 			}
 		} );
 
 		this.setState( { template: layout } );
-		setAttributes( { submitButtonText } );
 	}
```
jrtashjian: There is some strange behavior within `setTemplate`. We shouldn't need to track the template in state since any inner block changes are tracked by the InnerBlocks component. I think we can rip out this state property and maintain undo/redo states so long as we are handling when innerBlocks is empty.

richtabor:
Looks good. It doesn't look like its possible (https://github.com/WordPress/gutenberg/pull/18118) but it'd be great if we were able to ensure the Submit block doesn't get deleted accidentally. 🤔

PR #1402, opened 2020-03-05
Fix image and price toggle behavior on the Food and Drinks block

### LINKED ISSUE #1276: ISBAT disable prices from Food and Drinks block
**Describe the bug**
If a price is set on the Food Item block then prices are disabled from the parent Food and Drinks block, we receive "This block contains unexpected or invalid content."

image
image

**To Reproduce**

1. Add Food and Drinks block.
2. Click on Food Item block and type in a price.
3. Click on parent Food and Drinks block.
4. Disable (toggle) prices attribute in the editor sidebar. 
5. Save.
6. Reload page to see the error.

**Expected behavior**
Should be able to disable attribute at block level without errors.

**Isolating the problem:**

**WordPress Version**
5.3.2

**Gutenberg Version**
Happens with and without GB 7.2 running.

### LINKED ISSUE #1277: ISBAT disable images from Food and Drinks block
**Describe the bug**
If an image is set on the Food Item block then images are disabled from the parent Food and Drinks block, we receive "This block contains unexpected or invalid content."

image
image

**To Reproduce**

1. Add Food and Drinks block and click on the block.
2. Enable images in editor sidebar.
3. Click on Food Item block and add an image.
4. Click on parent Food and Drinks block.
5. Disable images in editor sidebar.
6. Save.
7. Reload page to see the error.

**Expected behavior**
Expect to toggle image attribute at a block level.

**Isolating the problem:**

**WordPress Version**
5.3.2

**Gutenberg Version**
Happens with and without GB 7.2 running.

AnthonyLedesma:
### Description

Closes #1276 and closes #1277.

Food and Drinks block contains controls used to alter children's block attributes. The causes of the bugs reported in #1276 and #1277 were the attributes `url` and `price` failing to reset when toggled off. 

The benefit of the previous design was that we could restore the previously used values as they were never cleared from markup.

This new solution uses localStorage to keep the `url` and `price` attributes ready to reference so a user can experiment with controls without losing the configured data.

### Screenshots

foodAndDrinksShowAttributes

### Types of changes

- Resolved issue where blocks returned `"This block contains unexpected or invalid content."` error.
- Refactored the use of `select` and `dispatch` to use proper High Order Components and injecting into props. 
- Renamed the edit components for the `Food and Drinks` and `Food Item` blocks. These components seemed to have been named incorrectly.

### How has this been tested?

Passes tests. Tested with 5.4RC and 5.3.2. Tested with and without the Gutenberg plugin.

### Checklist:
- [x] My code is tested


src/blocks/food-and-drinks/food-item/edit.js
```diff
@@ -115,17 +109,65 @@ class FoodAndDrinksEdit extends Component {
...
+			} else if ( url !== '' ) {
+				localStorage.setItem( 'menuItemImages', JSON.stringify( {
+					...storedAttributes,
+					[ clientId ]: {	url	},
+				} ) );
+
+				setAttributes( { url: '' } );
+			}
+		}
+	}
```
jrtashjian: This seems overly complex for what we're trying to do. I would utilize `localStorage` only when we need something to persist beyond this page load. Instead we can use the components state which will prevent us needing to do any sort of clientId tracking since we'd be working with the current instance of the component.

I think this should accomplish what we want in a simpler way using state (inside `componentDidUpdate`)

```
if ( this.props.attributes.showPrice ) {
	this.props.setAttributes( { price: this.state.price || '' } );
}
```

AnthonyLedesma: Changes made 👍

PR #1006, opened 2019-10-15
Introduce the ability for themes to provide custom icons for the Icon block

EvanHerman:
Resolves #923 

Introduce the ability to use custom icons directly within the icon block.

No UI is introduced in this PR to manage custom icons (add/remove). Icons (`.svg` files) must be manually added to the proper directory (listed below) in your theme root. This PR lays the groundwork for the behind the scenes work - in a future release we could potentially create a UI to manage these icons.

### How To
1. Create a `coblocks/icons/` directory inside of your theme root. (eg: `wp-content/themes/primer/coblocks/icons/`)
2. Drop your `.svg` files into this directory
3. Check the icon block for your custom icons 🎉

When the `config.json` file (_see below_) is not present, these values are generated from the icons file name.
eg: `adobe-photoshop.svg` => Label: Adobe Photoshop, Keywords: `adobe, photoshop`

#### Optional Icon Configuration File
Users can also create an optional `config.json` file inside of the `coblocks/icons/` directory in the theme root. In this file, they can further define settings for each icon.
- `label` - The label for the icon
- `keywords` - The keywords used to search for the icon
- `icon_outlined` - The `name` of the outlined version of the icon (eg: `adobe-photoshop-outlined.svg`)

#### Demo Icons/config.json
[coblocks-custom-icons.zip](https://github.com/godaddy/coblocks/files/3730252/coblocks-custom-icons.zip)

PR #1403, opened 2020-03-05
Introduce a filter to disable the CoBlocks bundled svg icons

### LINKED ISSUE #1394: ISBAT Filter out and remove CoBlocks icons SVGs
Awesome work on #1006  and this works great! It would be great to have a filter or boolean to perhaps allow users to disable all of CoBlock's icons and then load in their own in theme. Whereas, right now CoBlocks' icons are always loaded along with the user's added SVGs.

Could even be a boolean flag in the theme's `config.json`? Or, perhaps an `apply_filters( 'use_coblocks_svgs', return true );` at the top of [`get_custom_icons()`](https://github.com/godaddy-wordpress/coblocks/blob/master/includes/class-coblocks-block-assets.php#L160). 

Then user could set in their theme:
``` php
add_filter( 'use_coblocks_svgs', return false );
```

This would hide all of CoBlocks' SVG, and just show what is available in the theme's `coblocks/icons/*.svg` dir.

EvanHerman:
Resolves #1394, #1405

### Description
Introduce a new filter to disable the bundled icons that ship with CoBlocks. This allows users who implement custom icons to remove the excess icons that they may not ever use.

### Types of changes
**New filter:** `coblocks_bundled_icons_enabled`
**Note:** I named the filter this way to match the syntax of the typography controls filter, `coblocks_typography_controls_enabled`.

### Usage
```php
add_filter( 'coblocks_bundled_icons_enabled', '__return_false' );
```

### How has this been tested?
Manually tested.

### Checklist:
- [x] My code is tested

### DIFF
--- a/includes/class-coblocks-block-assets.php
+++ b/includes/class-coblocks-block-assets.php
@@ -133,6 +133,13 @@ class CoBlocks_Block_Assets {
 		 */
 		$typography_controls_enabled = (bool) apply_filters( 'coblocks_typography_controls_enabled', true, (int) $post_id );
 
+		/**
+		 * Filter to disable all bundled CoBlocks svg icons
+		 *
+		 * @param bool    true Whether or not the bundled icons are displayed.
+		 */
+		$bundled_icons_disabled = (bool) apply_filters( 'coblocks_bundled_icons_disabled', false );
+
 		$form_subject = ( new CoBlocks_Form() )->default_subject();
 
 		wp_localize_script(
@@ -145,6 +152,7 @@ class CoBlocks_Block_Assets {
 				),
 				'cropSettingsOriginalImageNonce' => wp_create_nonce( 'cropSettingsOriginalImageNonce' ),
 				'cropSettingsNonce'              => wp_create_nonce( 'cropSettingsNonce' ),
+				'bundledIconsDisabled'           => $bundled_icons_disabled,
 				'customIcons'                    => $this->get_custom_icons(),
 				'typographyControlsEnabled'      => $typography_controls_enabled,
 			)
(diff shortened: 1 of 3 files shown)


richtabor:
Looking good! One last bit: can you add a quick doc article that references this - using [this doc page](https://github.com/godaddy-wordpress/coblocks/blob/master/docs/hooks/settings-panel.md) as a template?

After that it's good to merge.

colorful-tones:
@EvanHerman I downloaded a .zip of your branch `add/filter-disable-bundled-icons`, ran `npm install` and `npm run build`.

I then went to create a new Page and add an Icon block and see this:
Screenshot 2020-03-05 16 42 07

I have `add_filter( 'coblocks_bundled_icons_enabled', '__return_false' );` in theme's `functions.php` and I currently have a series of SVGs in the theme's `my-theme/coblocks/icons/*.svg`. Am I missing something? Because previously my SVGs were showing alongside CoBlocks icons.

It says the `config.json` is/was optional, but wondering if this addition makes it a requirement? https://github.com/godaddy-wordpress/coblocks/pull/1006#issuecomment-560614001

EvanHerman:
@colorful-tones I don't think that there's anything missing, from what you describe. I am also seeing the same error now, but I'm not sure what may have changed.

I didn't notice the error because during my testing I added the icon block to the page, published it, and then added the filter, which removes all of the bundled icons without throwing any errors.

I am going to do some testing and see if I can narrow it down and fix things up. I will post back here when I've figured it out.

colorful-tones:
@EvanHerman works like a charm! 👏 

One thing seemed confusing, but not a blocker to release this IMO. So I opened a new Feature Request #1405

PR #1437, opened 2020-03-27
Introduce Vertical Alignment controls to Row block.

AnthonyLedesma:
### Description
Introduce Vertical Alignment controls to the Row and Column blocks.

### Screenshots

rowBlockAlignmentToolbar

### Types of changes

New controls and styles.

### How has this been tested?

Tested in 5.3.2 with and without Gutenberg 7.8.
Tested in 5.5beta with and without Gutenberg 7.8.

### Checklist:
- [x] My code is tested


src/blocks/row/styles/editor.scss
```diff
@@ -147,6 +147,121 @@
...
+					height: 100%;
+					> div[data-block] {
+						height: 100%;
+						margin-bottom: 0;
+					}
+				}
+			}
+	}
+	
+	.wp-block-coblocks-row__inner {
```
richtabor, line 199: This works, but it seems overly complicated - is this all necessary? Core columns doe it [like this](https://github.com/WordPress/gutenberg/blob/c19d2d908cba695960cf8407bd0b0afc181aa657/packages/block-library/src/columns/style.scss#L76), which is much simpler.

AnthonyLedesma: Changes made. Good call.

AnthonyLedesma:
Made some major changes to the styles. This simplifies things a great deal. I would appreciate another review @richtabor.

PR #1440, opened 2020-03-27
Use columns/gutter utilities on the Pricing Table block

richtabor:
### Description
This PR tweaks Pricing Table block to utilize our column utilities. Requires deprecation.


richtabor:
@AnthonyLedesma Do you mind taking over this one from me? Get the tests passing and ensure the depreciation passes (it doesn't right now).

AnthonyLedesma:
In this PR I identified a bug with attribute imports among three block deprecations. Using an import alias to correct bug across blocks.

Before
`import { blockAttributes } from './block.json';`
`import { currentBlockAttributes } from './block.json';`

After
`import { attributes as blockAttributes } from './block.json';`
`import { attributes as currentBlockAttributes } from './block.json';`

richtabor:
Once the deprecation is resolved I think this is good to go. I pushed a few minor style improvements to make these (and Features/Services blocks) work better on mobile with three and four column layouts (in both the editor and front-end). 

ScreenFlow
ScreenFlow2

richtabor:
Should close this issue on Go https://github.com/godaddy-wordpress/go/issues/474

richtabor:
Deprecation works ✅

PR #1450, opened 2020-04-13
Add proper example to the Pricing Table block

### LINKED ISSUE #1441: Accordion block not displaying correctly on Edge browser
### Describe the bug:

Accordion blocks display always open and can't be closed on Internet Explorer and Edge browsers.

### To reproduce:

1. Add an accordion block to a post or page.
2. Display as open should be toggled off.
3. View the accordion block on Edge or Internet Explorer.
4. See error: accordion displays as open and can't be closed.

### Expected behavior:

I expected the accordion to display as closed and only open when I click, as it does on other browsers.

### Isolating the problem:

### WordPress version:

WordPress version 5.4

### LINKED ISSUE #1449: ISBAT insert the Pricing Table block using the Block Inserter
### Describe the bug:
The block inserter is unexpectedly closed when user hovers the Pricing Table block. 

This **does not happen** with core block editor but only with the Gutenberg plugin active. I have been reproducing this error consistently using latest CoBlocks and Gutenberg plugins straight from the plugin repo (CoBlocks 1.23.0, Gutenberg 7.8.1) on a fresh WP install.

No other CoBlocks block shows this behavior to my knowledge.

### To reproduce:
- open the block inserter (the very left top of the screen, "plus" icon)
- find Pricing Table in the list
- hover it

### Expected behavior:
See the example preview and be able to click the button to insert the block.

### Isolating the problem:

### WordPress version:
5.4

### Gutenberg version:
7.8.1

### More investigation
There is no console error which is super strange and hard to debug without knowing this codebase.

There is sometimes a console warning which has been already documented in #799:

```
wp.blockEditor.RichText formattingControls prop is deprecated. Please use allowedFormats instead.
```

I have tried to check whether this is to root cause by removing all `formattingControls` usages. No effect so I assume it does not cause the error with the inserter.

After some more debugging I have found that this error is likely connected to the way Pricing Table generates its inner blocks during the block initialization. I haven't dug much deeper but I have an understanding that when this block sees no inner blocks, it generates and inserts `pricing-table-item` blocks into itself, based on the count attribute.

I don't know what is the root cause responsible but I've been able to make the inserter work again by providing an `example` of the pricing table block with inner blocks included. This probably skips the block-generating logic in the Pricing Block and just shows it.

```
	example: {
		attributes: {
			count: 1,
		},
		innerBlocks: [
			{
				name: 'coblocks/pricing-table-item',
				attributes: {
					placeholder: sprintf( __( 'Plan %d', 'coblocks' ), 1 ),
				},
			},
		],
	},
```

Since this is not a proper fix of the root cause, I haven't done a PR but unless you can pinpoint the error, I think it might be good idea to get this in. The preview it generates matches the one that will get generated anyway and follows how core blocks define their examples (including inner blocks, for [`example in core/columns`](https://github.com/WordPress/gutenberg/blob/c3798f141028240312dc7e7d530ad82b38900151/packages/block-library/src/columns/index.js#L33-L89)).

richtabor:
### Description
This PR aims to resolve #1449, while also adding a proper example for the Pricing Table block. This is just a start, ensuring that it does resolve the issue. We still need: 

- [x] Features added to each pricing table item example ("Feature one, Feature two, Feature three"
- [x] Button text added to each example (likely via a deeper `innerBlocks`).

### Screenshots
Screen Shot 2020-04-13 at 3 44 16 PM

### Types of changes
Fix + new feature (non-breaking change which adds functionality)

### Checklist:
- [x] My code is tested


getdave:
Do you know why the issue occurs? Is it because there are no `innerBlocks` definitions for the `example`? Is there a way we could improve Core to avoid errors in this case?

AnthonyLedesma:
image

Issue #1454, opened 2020-04-14
Only load front end CSS when it is needed


josephscott:
### Describe the bug:
I should be able to activate the plugin and not immediately have coblocks-style.css load an extra 141 KB of CSS on the front end of the site.

### To reproduce:
- Install and activate the coblocks plugin
- Notice that even without using any coblocks blocks an additional 141 KB of CSS is being loaded on the front end of the site

### Expected behavior:
Coblocks CSS should only be loaded when blocks that need it are in use.  That will shift the performance penalty so that it is limited to page views that require the CSS.

PR #1455, opened 2020-04-15
Only load front end CSS when it is needed

josephscott:
This checks to make sure that a Coblock is being used before requiring the CSS to be loaded on the front end of the site.  This is to address #1454 

In my basic tests this worked fine, however I'd like to see it get broader testing to make sure I haven't missed other conditions.


richtabor:
Good stuff, we'll run it though some testing. I'll hand this to you @jrtashjian. Should work find, though we'll need to make sure the assets load on index/archive views.

josephscott:
@richtabor  / @jrtashjian  - any thoughts on when this could be reviewed and included in a new release?

jrtashjian:
Hey @josephscott, I’m planning to review this during the current release cycle. If all looks/works well (which at first glance it does) then it’ll make it into the release next Thursday.

josephscott:
Sounds good!

jrtashjian:
Closes #1454

PR #1457, opened 2020-04-15
Improve Events block errors handling

### LINKED ISSUE #1363: ISBAT see an error message with Events block if not a valid calendar URL
### Describe the bug:
Events block can be given a URL that does not have a compatible calendar resulting in a block without content. An example of such a URL would be: https://www.nasa.gov/calendar. 

### To reproduce:
1. Add post and edit post.
2. Add Events block and enable `Link a Calendar` in Sidebar.
3. Paste incompatible URL such as https://www.nasa.gov/calendar.
4. See bug.

### Expected behavior:

Events block should use consistent error reporting in case of no content. 

### Screenshots:

eventsNonError

### Isolating the problem:

- This bug happens with no other plugins activated
- This bug happens with a default WordPress theme active
- This bug happens **without** the Gutenberg plugin active
- I can reproduce this bug consistently using the steps above

### WordPress version:

5.3.2

### Gutenberg version:

N/A (Also occurs with GB 7.5)

### LINKED ISSUE #1434: ISBAT Show a message when RSS events doesn't retrieve anything
I should be able to specify a message to display when the RSS feed doesn't fetch anything. E.g: 'No upcoming events'.

AnthonyLedesma:
### Description
Closes #1434 
Closes #1363 
The events block can enter a state where the serverSideRender has no content or error message presented to the user. This PR aims to improve UX by providing a contextual error message.

### Screenshots

**before**
eventsHandleNoEventsBefore

**after**
eventsHandleNoEvents

### Types of changes

Changes to the PHP server-side render function to handle the case of empty events. This occurs when using an RSS URL with the Events block.
EG( https://www.nasa.gov/rss/dyn/breaking_news.rss )

### How has this been tested?

Tested using various RSS URLs.
5.4 with and without Gutenberg 7.8.1

### Checklist:
- [x] My code is tested


AnthonyLedesma:
Enhanced this PR further by wrapping the error message in a filter. User can now set a custom error message which would display for the user on front-end or back-end. 

image

image

jrtashjian:
I'm going to push back on the filter for the error message.

Although interesting to have I don't think it's absolutely necessary and gives us more to maintain. This is an Events block and so it will be used for showing events specifically. We can add our own verbiage here that makes sense and remove the filter.

Additionally, it might be useful to show a message like "No upcoming events" to normal visitors but if an admin is logged in show the actual error message so they can debug.

jrtashjian:
I've assigned this PR to myself in an effort to get it finished for the next release.

jrtashjian:
Pushing updates shortly that utilizes the Notices UI of the Placeholder component for our messaging within the Block Editor:
image

AnthonyLedesma:
Trying out these changes I am running into one specific issue here. This is a blocker for merge as its part of the originally reported issue. 

ical with zero events
eventsBlockErrors

same ical with one sample event
eventsBlockErrorsWithEvent

This logic being used does not seem to properly catch cases with no events.
```php
		if ( ! $ical->has_events() ) {
			return new \WP_Error(
				'no_events_found',
				__( 'No events were found or an unsupported URL was used.', 'coblocks' ),
				array( 'status' => 404 )
			);
		}
```

Tried playing around with getting it correct but was not successful this morning. I might revisit this here soon.

PR #1465, opened 2020-04-20
Allow core/shortcode in the Pricing Table block

AnthonyLedesma:
### Description

The pricing table uses Innerblocks with a locked template. This PR opens up the template lock, allows the use of the block appender, and adds `core/shortcode` to the allowed blocks list.

Inspired by WP.org support request.
https://wordpress.org/support/topic/error-this-block-contains-unexpected-or-invalid-content/

### Screenshots

image

### Types of changes

Code changes. No deprecation needed.

### How has this been tested?

Tested with WP 5.4 with and without the Gutenberg plugin running.

### Checklist:
- [x] My code is tested


richtabor:
I don't like that we're adding more weight to the existing interface (with the new block-list-appender in the UI) for such an edge-case scenario. I don't mind adding support for the shortcode block, but not at the expense of everyone else's experience.

Screen Shot 2020-04-28 at 3 16 19 PM

richtabor:
I pushed a tweak to hide the inserter again. It's still possible to drag a shortcode block into the pricing table though. 

I also tweaked the placeholder component to display a bit better inside the tables.

Screen Shot 2020-04-28 at 3 19 16 PM

AnthonyLedesma:
@richtabor Good call! I think the drag approach will work best.

PR #1468, opened 2020-04-23
Add form block e2e test to confirm emails are being sent

EvanHerman:
### Description
This PR introduces a Cypress e2e test which submits the form and confirms that an email was sent and received.

It also adds a test to test that custom subject line is set when `wp_mail()` fires and tests that the `[email]` and `[name]` link keywords are added to the subject line in the block editor and then replaced in the subject line when sent.

### Types of changes

### How has this been tested?
I have run the tests a number of times locally and on CircleCI.

### Checklist:
- [x] My code is tested

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -239,6 +239,9 @@ jobs:
       - run:
           name: "Install Mailhog"
           command: bash .dev/bin/install-mailhog.sh
+      - run:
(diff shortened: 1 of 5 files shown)


fjarrett:
@EvanHerman Let's also add a check that custom subject fields are sent and received as expected.

PR #1486, opened 2020-05-13
Add Lightbox support for Lazy Loading

### LINKED ISSUE #1421: The lightbox does not work with lazy loading
### Describe the bug:
A user of my lazy loading plugin reported (https://wordpress.org/support/topic/plugin-compatibility-coblocks-gallery-lightbox/) that there are problems using it with CoBlocks’ lightbox for, for example, the offset gallery block.

I took a look into the code and the issue seems to be that on load the `img` elements are scraped for the `src` attribute values, to store them in a variable. But at that point, the `src` holds the placeholder data uri from my plugin, so the lightbox tries to display that instead of the image that gets loaded later.

### To reproduce:

1. Install and activate Lazy Loader (https://wordpress.org/plugins/lazy-loading-responsive-images/) and CoBlocks.
2. Create an offset gallery with the lightbox feature enabled.
3. View the frontend and try the lightbox.
4. See error.

### Expected behavior:

Correct displaying of the image :)

### Isolating the problem:

- [x] This bug happens with no other plugin (except Lazy Loader) activated

### WordPress version:
5.4 RC2-47460

AnthonyLedesma:
### Description
Closes #1421

The Lazy Loading plugin alters `img` element src attribute during load. Moved image preloader to an onClick event ensuring that correct `src` attribute is set.

### Types of changes

JavaScript changes to the Lightbox utility.

### How has this been tested?

Tested with and without the Lazy Loading plugin active. 

### Checklist:
- [x] My code is tested


src/js/coblocks-lightbox.js
```diff
@@ -71,8 +71,16 @@
 		const imagePreloader = {};
 
 		Array.from(images).forEach( function( img, imgIndex ) {
+			// Enables support for lazy loading base64 URLs by using the data-src attribute.
```
florianbrinkmann, line 74: Hi, wouldn’t it be better/working for more cases to change the code so that it gets the image source on click, instead of for all images on page load? Because when the image is lazy loaded, the `src` attribute holds the correct image link. And with that change (I guess) it would also work for other lazy load solutions that use other attributes.

florianbrinkmann: Thanks for working on this ❤️

AnthonyLedesma: Great feedback! Changes made. 👍

florianbrinkmann:
Thanks @AnthonyLedesma, tried the PR and it works! One thing I noticed by accident: When an offset gallery block has the link setting active to link the images to the media file, and after that the lightbox features is turned on, the lightbox just shows up for a short time and then the image URL is opened in the browser. Should I add another issue for that?

AnthonyLedesma:
Hi @florianbrinkmann, 

We would appreciate you writing up the issue for the bug you describe. Especially if you re-iterate the steps needed to replicate the bug. Reports like this helps to make CoBlocks better so we certainly encourage participation.

AnthonyLedesma:
This PR also resolves the bug reported on WP.org regarding Lightbox keyboard navigation:  https://wordpress.org/support/topic/coblock-masonry-lightbox-keyboard-arrows/

richtabor:
@AnthonyLedesma Does this resolve lazy-load issues with any plugin/JS that performs lazy loading? Or is it specific to this plugin?

AnthonyLedesma:
@richtabor Good call. I tested with 5 different lazy load plugins and was able to successfully use the lightbox in all instances. Made a minor change to the logic used to pre-load captions.

PR #1502, opened 2020-05-25
Update typography hooks docs.

juanfra:
The description was saying `PHP` when it is actually a `JavaScript` filter, and can be misleading.

### Description
Only a small change in the docs.


AnthonyLedesma:
@juanfra Nice catch. Made a few minor tweaks to the docs and specified that the bottom two filters are indeed PHP. I also made a consistency edit to another doc.

PR #1511, opened 2020-05-27
Improve conditional loading of scripts and styles

### LINKED ISSUE #1490: Coblock styles not loading on archive pages.
### Describe the bug:
Showing content that uses CoBlocks in a archive page doesn't shows the content correctly cause styles are not loaded. 

Looks like this would be linked to #1454 

### To reproduce:
1. Have content created with Gutenberg and CoBlocks in a post.
2. Editing your theme, load that content on at archive.php using some programming.
`$content = get_the_content(null, false, $postId);
if(!empty($content)){
    echo apply_filters( 'the_content', $content );
}`
3. The content appears, but CSS is not loaded.

### Expected behavior:
CoBlocks CSS to be loaded on pages where it is present, having in mind that it could be loaded in different circumstances than default behaviour. Or giving developers a way to override CSS optimisations. 

### Why
While Gutenberg is not available yet to create content to archive pages, I'm doing a custom linking to create content as post and show it as archive description.

### Isolating the problem:
- [x] This bug doesn't happen with 1.25.0 version.
- [x] This bug happens with 1.26.0 and 1.26.1 version.

### WordPress version:
5.3.2

### LINKED ISSUE #1491: Masonry: No scripts in front-end when made reusable
### Describe the bug:
It looks like the Masonry block doesn't enqueue scripts and styles on the front-end when converted into a reusable block.

### To reproduce:

1. Insert a Masonry block
1. Select two images and insert the "gallery"
1. Convert the block to a Reusable Block
1. Preview the post in the front-end
1. Notice the block is being displayed in a bulleted list.
-->

### Screenshots
See https://github.com/Automattic/wp-calypso/issues/42235

### Isolating the problem:

### WordPress version:
5.4.1, 5.5-alpha-47800

### Gutenberg version:
Whatever ships with the above WP versions, 8.0

### LINKED ISSUE #1496: Image block's block styles are not loading appropriate styles
### Describe the bug:
Due to how we're checking for `<-- wp:coblocks` to render our style assets, certain applications of styles do not properly get added. For instance, on the Image block we have a few style variations (waves, etc). It's not one of our blocks, so our styles do not load on that page.

Introduced by #1455.

### To reproduce:

1. Add an image block
2. Add a waves style
3. View the front-end

### Expected behavior:
The image block's (and any other style variations) should still be applied to the front-end, even if no CoBlocks blocks are on the same page.

### Error:
https://wpnux.godaddy.com/v2/?template=chai&lang=en_US

### Isolating the problem:

### WordPress version:
5.4.1

### Gutenberg version:
With and without 8.1

jrtashjian:
### Description
This PR builds upon #1454 which attempted to conditionally load CoBlocks scripts and styles only when needed. This attempt was great but in the end broke some other conditions we didn't test for.

The `has_block` method in WordPress Core as well as the custom check for blocks in the `coblocks/` namespace only work on singular (`is_singular()`) pages. Two conditions we missed related to any non-singular page with CoBlocks rendered (#1490 & #1496) and CoBlocks used within a reusable block that is inserted into a page (#1491).

Currently it's impossible to easily determine if a specific block is loaded anywhere on the page (in any WP_Query loop) or if they exist in a reusable block (because it's saved as a `core/block` type). To solve those two problems will take some thought to take care of server-side performance, specifically around database lookups.

The behavior this PR presents is the following:
1. Editor scripts and styles are verified to only load in the block editor.
2. Frontend scripts and styles are verified to load only on the frontend and block editor.
3. Frontend scripts and styles are loaded by default on all non `is_singular()` pages.
3. Frontend scripts and styles are loaded by default on all all `is_singular()` pages that include a reusable block (`core/block`).
4. Frontend scripts are conditionally loaded based on the blocks that exists on an `is_singular()` pages.

**EDIT**
Because we apply typography controls to certain core blocks and style treatments to the `core/image` block, we need to add these supported blocks to our conditions. In the future we will need to split up our stylesheets to minimize the payload rather than loading the entire stylesheet.

These are the currently supported blocks: `core/block`, `core/button`, `core/cover`, `core/heading`, `core/image`, `core/list`, `core/paragraph`, `core/pullquote`, and `core/quote`.

Loading all frontend scripts and styles on non-`is_singular()` pages or pages with a reusable block (`core/block`) isn't ideal but I think it's an acceptable trade-off right now until we can better detect blocks on a full rendered page.

### Types of changes
- Resolves #1490 
- Resolves #1496 
- Resolves #1491 

### How has this been tested?
Unit tests have been provided for the new conditions.

### Checklist:
- [x] My code is tested

### DIFF
--- a/.dev/tests/phpunit/includes/test-coblocks-block-assets.php
+++ b/.dev/tests/phpunit/includes/test-coblocks-block-assets.php
@@ -16,6 +16,10 @@ class CoBlocks_Block_Assets_Tests extends WP_UnitTestCase {
 
 		set_current_screen( 'dashboard' );
 
+		// Reset queued scripts and styles.
+		global $wp_scripts, $wp_styles;
+		$wp_scripts = new WP_Scripts();
+		$wp_styles = new WP_Styles();
 	}
 
 	public function tearDown() {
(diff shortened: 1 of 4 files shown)


.dev/tests/phpunit/includes/test-coblocks-block-assets.php
```diff
@@ -16,6 +16,10 @@ public function setUp() {
 
 		set_current_screen( 'dashboard' );
 
+		// Reset queued scripts and styles.
+		global $wp_scripts, $wp_styles;
+		$wp_scripts = new WP_Scripts();
+		$wp_styles = new WP_Styles();
```
EvanHerman, line 22: Should we add a blank line after `global $wp_scripts, $wp_styles;`?

The `=` signs should also be aligned here.

jrtashjian:
_I've updated the description above but here is the latest update:_

Because we apply typography controls to certain core blocks and style treatments to the core/image block, we need to add these supported blocks to our conditions. In the future we will need to split up our stylesheets to minimize the payload rather than loading the entire stylesheet.

These are the currently supported blocks: core/block, core/button, core/cover, core/heading, core/image, core/list, core/paragraph, core/pullquote, and core/quote.

PR #1515, opened 2020-05-29
Ensure extensions are applied properly to blocks

AnthonyLedesma:
### Description

Running Gutenberg 8.2.1 and the latest version of CoBlocks we are getting block validation error when inserting a Hero block to the page. Stack traces point to errors originating from HOC CoBlocks extensions. Changed three extensions which now allows for the hero block to be correctly inserted. There is a remaining outstanding issue as can be seen in the screenshot. 

CoBlocks extensions use a logic of `select( 'core/block-editor' ).getSelectedBlock()` to identify an inserted block. Without Gutenberg running, this function returns the parent hero block as expected. With Gutenberg active, a child block of the Hero block is returned. 

Looking into code changes I have not been able to determine the source of this bug. 

### Screenshots

Before this fix.
heroBlockBeforeFix

After this fix.
heroBlockAfterFix

### Types of changes

Logic to ensure that values exist.

### How has this been tested?

Tested manually with Gutenberg and without.

### Checklist:
- [x] My code is tested


jrtashjian:
Was the before shot a fresh compile from the master branch? I wasn't able to experience any issues initially. Are you able to post the original error and/or stack trace?

jrtashjian:
Hmm, so I just did a fresh build again on master and ran into the issue. It only seems to happen when inserting the block from the block inserter button in the toolbar and in the editor view.

Using the slash command `/hero` to insert does not trigger the error...

AnthonyLedesma:
@jrtashjian Thanks for testing. Strange we are seeing different behavior. This was a fresh compile at the time of the report. I'll get another look here. Notice within the provided screenshot the error does indeed occur for me when using `/hero` command.

Issue #1360, opened 2020-02-14
ISBAT have links from the Social Profiles Block open in a new window


benchilcote:
ISBAT have the option for social profile links to open in a new window or tab.  This is currently not an option in the block settings.

AnthonyLedesma:
This request also came in on WordPress.org
https://wordpress.org/support/topic/request-social-block-open-in-new-window/

webbpickersgill:
Thanks Anthony for re-posting my request here.  

Question for the dev team on this: Am I allowed to code this enhancement and then push it up?  I do coding, but I've never contributed to a project before through Git.  Are there any guidelines or tips for doing this correctly?  Thank you!

AnthonyLedesma:
Hi @webbpickersgill,

> Am I allowed to code this enhancement and then push it up?

Yes! We welcome all contributions to CoBlocks. Get a look at the [how you can contribute](https://github.com/godaddy-wordpress/coblocks#how-you-can-contribute) section on our README. There we link to our Contributing Guidelines and have other general information. 

If you are just getting started I would advise getting a look at the [development section](https://github.com/godaddy-wordpress/coblocks#development) of the README which covers setup.

webbpickersgill:
Excellent.  Thank you!  Those links will be very helpful.

richtabor:
Let us know if you need a hand @webbpickersgill.

I've put together a mockup of what we should do here. It'd be best to imitate how core does this. The following is based on the UI/X of adding a link to text within the core Paragraph block. 

<img width="866" alt="Screen Shot 2020-03-05 at 12 23 13 PM" src="https://user-images.githubusercontent.com/1813435/76007612-52038580-5edc-11ea-8cf7-05a7eb405b36.png">

And [here's the draft](https://www.figma.com/file/ub8cmDQQt3m6Iidi7lyHYd/CoBlocks-Social-Open-in-new-tab?node-id=0%3A1) in Figma.

richtabor:
Alternate approach (using a control within the sidebar): 
<img width="486" alt="Screen Shot 2020-04-16 at 2 04 41 PM" src="https://user-images.githubusercontent.com/1813435/79776748-2c9fcf00-8304-11ea-8b92-206019ae7b00.png">

I'm wondering if using the first option (commented above) would work better - but have the control labels "Open links in new tabs", which is a singular control for all the links.

PR #1520, opened 2020-06-03
Social profiles: Open links in new tab

juanfra:
### Description

Adding a general option for the social profiles block to open links in new tab. It should partially fix https://github.com/godaddy-wordpress/coblocks/issues/1360

I thought that experience wise, either you choose all links to open in a new tab, or not. That's why I took this approach. And because it doesn't lock "us" to implement on a link basis in the future.

*Note*: I'm building a new site using coblocks and I found it more useful this way.
*Note 2*: I want to add GitHub to this block. Maybe this is more a note to myself. I'll probably open a new issue.

### Types of changes

I added a new block setting for the *Social Profiles* block. Used the same naming as WP does for this, to keep it consistent.

Finally, edited the block template to reflect the changes in the setting.

### How has this been tested?

- Added the social profiles block.
- Configured the "Open links in new tab" setting (ON/OFF).
- Tested the results in the FE.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/social-profiles/block.json
+++ b/src/blocks/social-profiles/block.json
@@ -13,6 +13,10 @@
 			"type": "boolean",
 			"default": true
 		},
+		"opensInNewTab": {
+			"type": "boolean",
+			"default": false
+		},
 		"borderRadius": {
 			"type": "number",
 			"default": 40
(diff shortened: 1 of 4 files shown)


juanfra:
Thanks @AnthonyLedesma! 🎉

PR #1526, opened 2020-06-09
Fix overwriting existing line height style when none has been set in CoBlocks

### LINKED ISSUE #1500: Gutenberg's line-height setting doesn't work if CoBlocks is activated
### Describe the bug:
Gutenberg plugin introduced line-height setting, but in a site with CoBlocks, the setting is not reflected on the front of the site.   

### To reproduce:

1. Go to the editor.
2. Add a heading or a paragraph.
3. Alter the line-height in the block setting in the sidebar.
4. Notice the line-height change in the editor.
5. See the front of the site.
6. Notice the line-height setting is not respected on the front of the site.

Or
1. Deactivate CoBlocks.
2. Go to the editor and add a heading or a paragraph.
3. Change the line-height of the blocks and save.
4. Activate CoBlocks.
5. Open the post/page that you've created in 3.
6. See the invalid block error. 

### Expected behavior:
* The line-height setting should be reflected on the front of the site.
* The content made without CoBlock should not break with CoBlocks activated. 

### Screenshots:
An extreme example to illustrate the issue.

**In the editor:**
Screenshot 2020-05-23 at 03 03 53

**Front of the site**
Screenshot 2020-05-23 at 03 05 02

**Content that altered the line-height setting with CoBlock activated**
Screenshot 2020-05-23 at 03 18 41

### Isolating the problem:

### WordPress version:
5.4.1

### Gutenberg version:
8.1.0

p-jackson:
### Description

The latest version of Gutenberg includes typography settings which already exist in CoBlocks. This PR doesn't fully harmonise those settings, but it stops CoBlock from disabling the core settings.

Using `line-height` as an example: if the line height was unset in CoBlocks, then `applyStyle()` would set the inline style to `null`. This means that if core Gutenberg had set it's own line height then it would get nulled out. As shown in #1500 this caused the line height to be removed on the front end and could cause invalid block errors.

This PR only sets an inline typography style if one has been set in CoBlocks, otherwise it leaves it alone. This means the CoBlocks setting takes precedent over the core Gutenberg setting on the front end. Maybe it should be the other way round, but this at least fixes the invalid block errors.

It's possible to do this without the `_.pickBy` function but it's more verbose and I saw this function was already used elsewhere in the plugin. From a code size point of view.

Fixes #1500 

### Types of changes

### How has this been tested?

- Enable latest Gutenberg and disable CoBlocks
- Add a paragraph block to a post
- Set a custom line height in the block settings
- Enable CoBlocks
- Open post, should have no block errors
- Preview post, line height should be applied

And checked the other way round just to be sure
- Enable CoBlocks and disable latest Gutenberg
- Add a paragraph block to a post
- Set a custom line height using the CoBlocks settings in the toolbar
- Enable latest Gutenberg
- Open post, should have no block errors
- Add a different line height using core block settings
- Preview post, line height from CoBlocks settings should be applied

Also tried the above using [the other blocks that have custom typograph settings](https://github.com/godaddy-wordpress/coblocks/blob/5b28202bc0bbe6235d7906314146006a2c829acf/src/extensions/typography/index.js#L21).

### Checklist:
- [x] My code is tested

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -30,7 +30,7 @@ This list is manually curated to include valuable contributions by volunteers th
 | @mahdiyazdani     | @mahdiyazdani          |
 | @nicolad          | @vadimnicolai          |
 | @obenland         | @obenland              |
-| @p-jackson        |                        |
+| @p-jackson        | @philipmjackson        |
 | @passatgt         | @passatgt              |
 | @ramiy            | @ramiy                 |
 | @Reinpal          |                        |
--- a/src/extensions/typography/apply-style.js
+++ b/src/extensions/typography/apply-style.js
@@ -1,19 +1,12 @@
-function applyStyle( attributes, name ) {
-	const {
-		fontFamily,
-		lineHeight,
-		letterSpacing,
-		fontWeight,
-		textTransform,
-	} = attributes;
+/**
+ * External dependencies
+ */
+import pickBy from 'lodash/pickby';
 
-	const style = {
-		lineHeight: lineHeight || null,
-		fontFamily: fontFamily || null,
-		fontWeight: fontWeight || null,
-		textTransform: textTransform || null,
-		letterSpacing: letterSpacing || null,
-	};
+const styleAttributes = [ 'fontFamily', 'lineHeight', 'letterSpacing', 'fontWeight', 'textTransform' ];
+
+function applyStyle( attributes, name ) {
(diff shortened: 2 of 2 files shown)


AnthonyLedesma:
Though I approve this PR the work needed for typography controls is not yet completed. A new properly scoped issue details the outstanding required fixes with typography controls. 
https://github.com/godaddy-wordpress/coblocks/issues/1528

PR #1545, opened 2020-06-17
Fix Accordion children block appender

AnthonyLedesma:
### Description

Accordion block appender had been using `onClick` listener to handle the creation of new children. I found that changes within the DOM upon `mouseDown` event were causing the `onClick` event to cease bubbling thus resulting in a failure to execute the callback. 

By using `onMouseUp` listener we can circumvent the bubbling issue and react to user interaction.   

### Screenshots

accordionAddButton

### Types of changes

Minor changes to event listener attached to the block appender.

### How has this been tested?

Tested manually with and without the Gutenberg plugin active. Tested with Go and Twenty Twenty. 

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/accordion/edit.js
+++ b/src/blocks/accordion/edit.js
@@ -56,6 +56,43 @@ class AccordionEdit extends Component {
 
 		const items = this.props.getBlocksByClientId( clientId );
 
+		const handleEvent = ( ) => {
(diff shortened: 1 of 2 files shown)


richtabor:
I'm getting some odd behavior here? 

acc

PR #1571, opened 2020-07-14
Detect modified code in extensions when setting up test specs

jrtashjian:
This updates our `.dev/bin/setup-test-specs.sh` script to also check for modified files within the `src/extensions` directory when setting up specs for CI.


jrtashjian:
Looks like we have some tests to fix!

PR #1575, opened 2020-07-17
Fix Click to Tweet inline toolbar via SVG

richtabor:
This PR is a quick style fix to adjust the SVG within the "via" toolbar component in the Click to Tweet block. Tested with core WP and Gutenberg 8.5.1

Before: 
Screen Shot 2020-07-17 at 12 49 18 PM

After: 
Screen Shot 2020-07-17 at 12 48 05 PM
Screen Shot 2020-07-17 at 12 48 42 PM


AnthonyLedesma:
Pushed a minor fix to handle header toolbar styles.
image

PR #1577, opened 2020-07-17
Layout selector performance improvements

jrtashjian:
### Description
The way we were using the BlockPreview component from Gutenberg was very slow when loading more than 5 previews in a list. With the release of Block Patterns in Gutenberg I've recognized performance improvements that were made in core where they BlockPreview component was being used inth block inserter. This PR takes those findings and implements them here.

The result is a more snappy interaction when interacting with the previews and no delay in interacting with the layout selector while previews load.

### Types of changes
Enhancement

### How has this been tested?
This has been tested manually with 1, 5, 10, 20, 50, and 100 layouts being added through the available filters in PHP.

### Checklist:
- [x] My code is tested


includes/class-coblocks-block-assets.php
```diff
@@ -295,7 +296,7 @@ public function frontend_scripts() {
 		$vendors_dir = CoBlocks()->asset_source( 'js', 'vendors' );
 
 		// Masonry block.
-		if ( has_block( 'coblocks/gallery-masonry' ) || has_block( 'core/block' ) ) {
+		if ( is_admin() || has_block( 'coblocks/gallery-masonry' ) || has_block( 'core/block' ) ) {
```
richtabor, line 299: Is there a check for a Gutenberg admin page instead?

src/extensions/layout-selector/editor.scss
```diff
@@ -4,36 +4,6 @@ $color-highlight: #00a4a6;
 	opacity: 1 !important;
 }
 
-.coblocks-layout-selector .coblocks-layout-selector__layout .block-editor-block-preview__container.editor-styles-wrapper {
```
richtabor, line 7: Why did you remove these styles?

src/extensions/layout-selector/index.php
```diff
@@ -46,6 +50,128 @@ function coblocks_layout_selector_categories() {
 function coblocks_layout_selector_layouts() {
 	$layouts = array();
 
+	for ( $i=0; $i <= 100; $i++ ) {
```
richtabor, line 53: Wonder if we should have one for testing, if WP_DEBUG is on, or a test var we can state?

jrtashjian: This was mistakenly committed. I've removed it.

includes/class-coblocks-block-assets.php
jrtashjian: Yes, we actually have `is_page_gutenberg()` we can use. Switching to it in the next push.

src/extensions/layout-selector/editor.scss
jrtashjian: Probably just did it to strip every possible override during experimentation. This will continue to exist.

jonathanbardo:
@jrtashjian Can you take a look at the conflict? Pretty sure it's only because we took out image fetch.

src/extensions/layout-selector/layout-selector-results.js
```diff
@@ -0,0 +1,117 @@
...
+				 *
+				 * @param {Array} blocks The block objects of the layout.
+				 */
+				blocks: applyFilters( 'coblocks.layoutPreviewBlocks', layout.blocks ),
+			} ) ),
+		[ layouts, category ]
+	);
+
+	// Needed to render our list of previews asynchronously for better performance.
+	const currentShowLayouts = useAsyncList( filteredLayouts );
```
jonathanbardo, line 35: Oh that's pretty cool. Nice find

src/extensions/layout-selector/layout-selector-results.js
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Renders the layout's block preview.
+ */
+export const LayoutPreview = ( { layout, onClick } ) => {
+	const [ overlay, setOverlay ] = useState( false );
+	const isSelected = false;
+
+	return (
+		<a href="#!"
```
jonathanbardo, line 67: @jrtashjian liked we chat about, I would remove the anchor here and move to a button. Replace isSelected with overlay and remove the button below.

jonathanbardo (changes requested):
This LGTM. Small changes on the LayoutPreview markup only.

src/extensions/layout-selector/layout-selector-results.js
jrtashjian: Yeah, it was buried in the patterns api code. Works well :)

src/extensions/layout-selector/layout-selector-results.js
jrtashjian: Resolved.

PR #1581, opened 2020-07-21
Layouts as Custom Post Type

jrtashjian:
### Description
This PR introduces a new custom post type and related taxonomies to allow saving custom block patterns and full page layouts. These new block patterns will be loaded into the "Patterns" tab of the block inserter and the full page layouts are loaded into the Layout Selector provided by CoBlocks.

UI is currently shown but will be hidden before merging. UI coming later but #1659 has a UI for saving Block Patterns here.

### Screenshots

Creating a new layout in the custom post type
creating-layout-in-cpt

Viewing the quick-edit for the created layout in the CPT
viewing-layout-quick-edit-in-cpt

The created layout listed within the Layout Selector modal when creating a new page.
layout-selector-create-new-page

### Types of changes
New feature

### How has this been tested?
Manually. E2E/Unit tests still need to be added.

### Checklist:
- [ ] My code is tested


class-coblocks.php
```diff
@@ -115,17 +116,7 @@ private function includes() {
...
 
-			// Require the Gutenberg plugin for specific components.
-			include_once ABSPATH . 'wp-admin/includes/plugin.php';
-			$installed_plugins = get_plugins();
-
-			$gutenberg_plugin_file    = 'gutenberg/gutenberg.php';
-			$gutenberg_plugin_version = empty( $installed_plugins[ $gutenberg_plugin_file ] ) ? null : $installed_plugins[ $gutenberg_plugin_file ]['Version'];
-
-			// Require CoBlocks custom filters
-			require_once COBLOCKS_PLUGIN_DIR . 'src/components/gutter-control/gutter-wrapper.php';
```
jonathanbardo, line 126: @jrtashjian did that move somewhere else?

jrtashjian: Mistakenly removed. Fixed in 3aee61c

includes/class-block-patterns.php
```diff
@@ -0,0 +1,144 @@
...
+	private static $instance;
+
+	/**
+	 * Registers the plugin.
+	 *
+	 * @return CoBlocks_Block_Patterns
+	 */
+	public static function register() {
+		if ( null === self::$instance ) {
+			self::$instance = new CoBlocks_Block_Patterns();
```
jonathanbardo, line 30: I feel like this is prone to error doing copy paste. I would suggest using `new self()` instead. Also since we are over php 5.4 now I feel like we could introduce a singleton trait that we can reuse everywhere.

```php
trait Singleton {
	/**
	 * The plugin instance.
	 *
	 * @var Plugin
	 */
	private static $instance = null;
	/**
	 * Return the plugin instance.
	 *
	 * @return Plugin
	 */
	public static function load() {
		if ( ! self::$instance ) {
			self::$instance = new self();
		}
		return self::$instance;
	}
	/**
	 * Reset the plugin instance.
	 */
	public static function reset() {
		self::$instance = null;
	}
}
```

jrtashjian: Good call. This will be added in the next push.

richtabor (approved):
If @jonathanbardo is good with this, I am too. 💪

jonathanbardo (approved):

PR #1587, opened 2020-07-24
Fix Lightbox use with Go theme and Carousel Gallery block

AnthonyLedesma:
### Description

The CoBlocks Lightbox works by adding `click` event listeners to document elements. Previous iterations of Lightbox worked by adding a click listener to the `img` elements which were arguably at the bottom of the document hierarchy. Because the event listener was near the bottom of hierarchy we made it possible for events to fire on elements that reside above our selected target. As a result of this bug, the Lightbox for Go theme when using the Carousel Gallery block would not function. 

This PR changes the listener target from the lowest `img` element to the parent `figure` element of each respective image. This allows the event bubbling to function as expected even if theme styles result in click events firing on the Figure or the Img elements. 

### Screenshots

CarouselGoFixed

### Types of changes

Minor JavaScript change to use `.closest( 'figure' )` function. 

### How has this been tested?

Tested manually in Chrome, and Firefox. 
Tested with and without the Gutenberg plugin active.
Tested with Go, TwentyTwenty, Barebones theme (underscores).

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/js/coblocks-lightbox.js
+++ b/src/js/coblocks-lightbox.js
@@ -3,7 +3,7 @@
 
 	const lightboxModals = document.getElementsByClassName( 'has-lightbox' );
 
-	Array.from( lightboxModals ).forEach( ( lightbox, index ) => {
+	Array.from( lightboxModals ).forEach( function( lightbox, index ) {
 		lightbox.className += ' lightbox-' + index + ' ';
 		renderLightboxModal( index );
 	} );
@@ -69,7 +69,7 @@
 		}
 
 		Array.from( images ).forEach( function( img, imgIndex ) {
-			img.addEventListener( 'click', function() {
+			img.closest( 'figure' ).addEventListener( 'click', function() {
 				changeImage( imgIndex );
 			} );
 		} );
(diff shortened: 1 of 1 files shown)


src/js/coblocks-lightbox.js
```diff
@@ -116,8 +116,8 @@
 						imagePreloader[ `img-${ imgIndex }` ] = new window.Image();
 						imagePreloader[ `img-${ imgIndex }` ].src = img.attributes.src.value;
 						imagePreloader[ `img-${ imgIndex }` ][ 'data-caption' ] =
-							( images[ imgIndex ] && images[ imgIndex ].nextElementSibling )
-								? getImageCaption( images[ imgIndex ] ) : '';
+								( images[ imgIndex ] && images[ imgIndex ].nextElementSibling )
```
jasonlemay, line 119: why are we changing the spacing here? It looks like the previous version (one space) was good

src/js/coblocks-lightbox.js
```diff
@@ -3,7 +3,7 @@
 
 	const lightboxModals = document.getElementsByClassName( 'has-lightbox' );
 
-	Array.from( lightboxModals ).forEach( ( lightbox, index ) => {
+	Array.from( lightboxModals ).forEach( function( lightbox, index ) {
```
jasonlemay, line 6: Why are we changing fat arrows to function? Is this because of a problem with `this` or this is to unify the codebase?

src/js/coblocks-lightbox.js
AnthonyLedesma: Hey Jason. This one was actually made by the ESLint rules that I have loaded within my editor Visual Studio Code. We use the same rules that are defined from Core with a few minor modifications, so occasionally when those rules change, we see code shifts like this. 

https://github.com/godaddy-wordpress/coblocks/blob/master/.eslintrc.js

src/js/coblocks-lightbox.js
AnthonyLedesma: This was to unify the codebase specifically. I realized that this was the only `forEach` method using arrow syntax, so I made that change to unify to use of `forEach`.

PR #1590, opened 2020-07-25
Consider adding a BlockBook to the repository

youknowriad:
If you missed my announcement, I released a tool to showcase and develop blocks in isolation https://riad.blog/2020/07/22/introducing-blockbook-for-wordpress/ and I've been using coblocks as one of the test use-cases. I figure that you might  be interested in adding it to the repository.

Capture d’écran 2020-07-25 à 4 23 11 PM

This PR adds a blockbook in this repository by just registering all the blocks that I've found.

 - I didn't register any custom stories for the blocks, but I think you should consider adding different stories for the different blocks (See the "stories" tab here as an example https://youknowriad.github.io/blockbook/block/core---buttons )
 - I added a Github action to deploy it as a Github Pages for this repository.

 - As you'll see there are still some JS errors triggered here and there (relying on globals, specific setup and things like that) but I'd be happy to help see if these can be resolved in blockbook or here (depending on the error).
 - Also right now BlockBook builds its output in `dist` folder (can't be overriden yet) which overwrites some of your files temporarily.

You can try locally `npm run blockbook:start` 

Let me know what you think.

Thanks.


jrtashjian:
Fantastic! Thanks @youknowriad! As soon as we saw the announcement we talked about integrating it because previously we were experimenting with Frontenberg to accomplish a similar thing.

I'm excited to review this.

youknowriad:
Oh, great updates here. Awesome, I was using `npm link` and I forgot to add the dependency 🤕

jrtashjian:
@youknowriad this works well and looks pretty good already. There are still some issues with certain blocks though that I haven't dug deeper into yet. I did figure out a way to override the output path though by overriding specific webpack config properties. Is there another reason I'm overlooking that you mentioned it wasn't possible yet?

@richtabor could you determine what kind of "stories" we can add to this? Stories look to be a way to present different examples of how our blocks can be used. We'll need to figure out the image assets we want to use for this as well.

TODO:
- [ ] Add TwentyTwenty theme preview
- [ ] Add Go theme preview
- [ ] Add stories for each block
- [ ] Identify any remaining block rendering issues

Not sure if there is anything else remaining after those items. Maybe we can restyle the page to better resemble the GoDaddy brand.

youknowriad:
>  I did figure out a way to override the output path though by overriding specific webpack config properties. Is there another reason I'm overlooking that you mentioned it wasn't possible yet?

No, actually your fix is great but I wanted to provide that a simple CLI argument maybe.

youknowriad:
> Maybe we can restyle the page to better resemble the GoDaddy brand.

While you can already do that by just loading CSS, I do think we'd need an official API for this to make it more  "future proof" as overrides can break with updates.

youknowriad:
Going to close this as it went stale.

PR #1607, opened 2020-08-03
Download remote images used in layouts

jrtashjian:
### Description
When a layout is inserted using our Layout Selector, the example images we provide are now downloaded to the WordPress website so that they are not being pulled by the server remotely. So long as each image/gallery block handles it's own upload state (blob detection), the effect should be visible with loading indicaters properly shown. Any blocks missing this functionality will need to be updated.

The only blocks enabled right now are `core/image`, `core/cover`, and `core/gallery`. We'll need to test additional blocks including our own. I believe we are using a similar attribute format that lines up with core which should make things easier. Additional blocks can be added for testing here: https://github.com/godaddy-wordpress/coblocks/compare/layout-selector-download-remote-images?expand=1#diff-b4042632c815428a1b269a87da50fa1cR149-R150

#### Layout Selector

Using a test layout with the image URL linking directly to my [personal website](https://jrtashjian.com/).

Screen Shot 2020-08-03 at 12 01 46 AM

#### Image uploading
Screen Shot 2020-08-03 at 12 01 50 AM

#### Image uploaded to site's media library
Screen Shot 2020-08-03 at 12 02 04 AM

### Types of changes
Enhancement to the layout selector. Pulls remote images and uploads them to the site's media library automatically.

### How has this been tested?
Manually, we need to figure out how to add automated testing here.

### Checklist:
- [ ] My code is tested


richtabor:
I was able to get the Image block to work, but the core/gallery didnt seem to be working on my end.

src/extensions/layout-selector/index.js
```diff
@@ -92,6 +93,28 @@ class LayoutSelector extends Component {
 		this.useTemplateLayout = this.useTemplateLayout.bind( this );
 		this.useEmptyTemplateLayout = this.useEmptyTemplateLayout.bind( this );
 		this.renderContent = this.renderContent.bind( this );
+		this.detectImageBlocks = this.detectImageBlocks.bind( this );
+		this.uploadExternalImages = this.uploadExternalImages.bind( this );
+	}
+
+	componentDidUpdate( prevProps ) {
+		if ( prevProps.clientIds.length === 0 && this.props.clientIds.length !== 0 ) {
```
jonathanbardo, line 101: @jrtashjian This is a big If statement. I would find the reverse condition and return early instead if possible.

src/extensions/layout-selector/index.js
```diff
@@ -92,6 +93,28 @@ class LayoutSelector extends Component {
...
+			this.detectImageBlocks( this.props.clientIds )
+				.filter(
+					( block ) => {
+						if ( typeof block === 'undefined' || isEmpty( Object.values( block )[ 0 ] ) ) {
```
jonathanbardo, line 105: @jrtashjian I think this could easily by a one liner and still be readable. Just a simple return statement with the condition.

src/extensions/layout-selector/index.js
```diff
@@ -117,6 +140,80 @@ class LayoutSelector extends Component {
...
+								} );
+							} else {
+								updateBlockAttributes( clientId, {
+									id: media.id,
+									url: media.url,
+								} );
+							}
+						},
+						onError( message ) {
+							console.log( [ 'onError', message ] );
```
jonathanbardo, line 210: @jrtashjian Can we find a way to display an error to the end user directly in that case insteaf of burrying it inside the console?

src/extensions/layout-selector/index.js
```diff
@@ -117,6 +140,80 @@ class LayoutSelector extends Component {
...
+								updateBlockAttributes( clientId, {
+									ids: media.id !== null ? [ ...currentAttributes.ids, media.id ] : currentAttributes.ids,
+									images: newImages,
+								} );
+							} else {
+								updateBlockAttributes( clientId, {
+									id: media.id,
+									url: media.url,
+								} );
+							}
```
jonathanbardo, line 207: @jrtashjian I would reverse the condition here and get rid of the else entirely since it's such a simple condition.

src/extensions/layout-selector/index.js
AnthonyLedesma: Agree that the errors need to be surfaced for the end-user. I attempted a number of URLs with images but am getting error `post.php?post=1985&action=edit:1 Access to fetch at 'https://www.cloudflare.com/img/learning/ddos/what-is-a-ddos-attack/syn-flood-ddos-attack.png' from origin 'https://wordpress.test' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.` followed by `Failed to load resource: net::ERR_FAILED`. The Cors issues may be unavoidable but if we have failure to upload, the user should be aware.

src/extensions/layout-selector/index.js
AnthonyLedesma: Something like this @jonathanbardo?
```javascript
.filter( ( block ) => ( typeof block === 'undefined' || isEmpty( Object.values( block )[ 0 ] ) ) ? false : true )
```

jonathanbardo: Yes exactly

src/extensions/layout-selector/index.js
```diff
@@ -117,6 +140,80 @@ class LayoutSelector extends Component {
...
+		const urls = [];
+
+		if ( blockAttributes.hasOwnProperty( 'images' ) ) {
+			blockAttributes.images.forEach( ( image ) => urls.push( image.url ) );
+		} else {
+			urls.push( blockAttributes.url );
+		}
+
+		urls.forEach( ( imageUrl, index ) => {
+			window.fetch( imageUrl )
```
jonathanbardo, line 182: I would be curious if we could use `globalThis` instead of window. It would make the code a little more platform agnostic. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

src/extensions/layout-selector/index.js
jonathanbardo: I was afraid of this... not everything will be downloadable from the browser itself. Wondering if we need to come up with a way to download the image server side and then send it down to the front end to circumvent CORS issues.

jonathanbardo: If remote files are fetched from a server we can control this is a non issue.

src/extensions/layout-selector/index.js
```diff
@@ -117,6 +140,80 @@ class LayoutSelector extends Component {
...
+					mediaUpload( {
+						filesList: [ blob ],
+						allowedTypes: [ 'image' ],
+						onFileChange( [ media ] ) {
+							if ( blockAttributes.hasOwnProperty( 'images' ) ) {
+								const currentAttributes = getBlockAttributes( clientId );
+
+								const newImages = currentAttributes.images.map( ( oldImage, oldIndex ) => {
+									return oldIndex === index
+										? Object.assign( {}, oldImage, { url: media.url } )
```
jonathanbardo, line 194: I would use the slightly less opaque syntax here.
```js
{
 ...oldImage,
url: media.url
}
```

src/extensions/layout-selector/index.js
jrtashjian: I believe we can use globalThis. Although IE 11 doesn't have support it looks like babel handels it.

jonathanbardo: Do we officially support IE though?

src/extensions/layout-selector/index.js
jrtashjian: Fixed in 4d05075

src/extensions/layout-selector/index.js
jrtashjian: Fixed in 7c959cb

src/extensions/layout-selector/index.js
jrtashjian: Fixed in 30ff716. I used a warning notice over an error notice because the issue is non-blocking for the user, we only want to let the user know we were unable to pull down the remote image.

src/extensions/layout-selector/index.js
jrtashjian: Yes, we support IE >= 11 just like the [Gutenberg project](https://github.com/WordPress/gutenberg/blob/master/packages/browserslist-config/index.js). @richtabor would know if this has changed.

src/extensions/layout-selector/index.js
jrtashjian: Fixed in 07f1e16

src/extensions/layout-selector/index.js
jrtashjian: Changed in 382e0e6

src/extensions/layout-selector/index.js
jonathanbardo: Okay, if babels handles it let's use it then

AnthonyLedesma:
Looks like we have some failures with the Layout Selector in this branch/PR. This does **not** happen on master branch.
layoutSelector

richtabor (commented):
@jrtashjian I'm running into an issue trying to get it to work. Tried a few different image sources. Ideas? 
Screen Shot 2020-09-14 at 1 27 25 PM

jrtashjian:
This error will happen (you should see a notice in the editor) on servers we don't have control of. Remote images on my website work because of the way I've configured it, and we will have to make sure GoDaddy's CDN is configured properly as well.

richtabor:
> This error will happen (you should see a notice in the editor) on servers we don't have control of. Remote images on my website work because of the way I've configured it, and we will have to make sure GoDaddy's CDN is configured properly as well.

Ok cool, what requirements are there (to reference in future stories)?

richtabor:
> This error will happen (you should see a notice in the editor) on servers we don't have control of. Remote images on my website work because of the way I've configured it, and we will have to make sure GoDaddy's CDN is configured properly as well.

@jrtashjian Here's with trying an image from your site. Do you have a different image URL I could test with?
Screen Shot 2020-09-14 at 1 35 52 PM

AnthonyLedesma (approved):
re-tested changes and am getting the behavior that I expected for the downloader on the core/image, core/cover, and core/gallery blocks. I have added [new tests to the e2e PR here](https://github.com/godaddy-wordpress/coblocks/pull/1642/commits/f1bf2417b7912aa2be0d7ef40bffaa4e3967f111) which will properly tests the downloader function. 

LGTM

jrtashjian:
There is one test failing but it seems to be an errorcoming from a core component. The odd bit is that these tests pass locally without any issue on Chrome 85 (what Cypress is using). Here's a screenshot as proof.

image

I fixed a couple issue I ran across with how ID's were being inserted into the block attributes, causing some oddities. I also added a proper check to make sure *only* remote images are being downloaded to the library and not the ones included in our Go theme.

richtabor (approved):

PR #1610, opened 2020-08-04
Prevent empty Advanced panel from showing on blocks

### LINKED ISSUE #1364: ISNBAT see Gif block sidebar panels lacking controls
### Describe the bug:
When adding a Gif block to a page and Gutenberg plugin is active a useless sidebar panel option is added to the block. 

### To reproduce:
1. Add post and edit post.
2. Add Gif block and view the sidebar.
3. See bug.

### Expected behavior:

Expect advanced panel not to exist when no advanced options are present.

### Screenshots:

image

### Isolating the problem:

### WordPress version:

5.3.2

### Gutenberg version:

This occurs only with the Gutenberg plugin active. Tested with GB 7.5.

AnthonyLedesma:
### Description

Closes #1364
The `advanced-controls` extension for CoBlocks had been inadvertently adding an `InspectorAdvancedControls` where it should not be. This PR adds logic to exclude the `InspectorAdvancedControls` component when necessary.

### Screenshots

image

### Types of changes

Minor JavaScript changes.

### How has this been tested?

Tested manually within the editor.
Tested with and without the Gutenberg plugin active.

### Checklist:
- [x] My code is tested

### DIFF
--- a/.dev/bin/setup-test-specs.sh
+++ b/.dev/bin/setup-test-specs.sh
@@ -27,9 +27,13 @@ do
   if [[ $FILE == *"src/extensions/"* ]]; then
     testSpec=$(echo $FILE | cut -d'/' -f3)
 		foundwords=$(echo ${SPECS[@]} | grep -o "${testSpec}" | wc -w)
+		# Catch cases where no cypress spec file exists. 
+			if [[ $(ls -l src/extensions/${testSpec}/**/*.cypress.js | wc -l) -eq 0 ]]; then
+				continue
+			fi
 		# The test spec does not yet exist in the SPECS array
 		if [[ "${foundwords}" -eq 0 ]]; then
-			# Spec file string is empty, do not start string with a ,
+			# Spec file string is empty, do not start string with a comma
 			if [[ ${#SPECSTRING} -eq 0 ]]; then
 				SPECSTRING="src/extensions/${testSpec}/**/*.cypress.js"
 			else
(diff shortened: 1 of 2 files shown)


AnthonyLedesma:
@EvanHerman This PR has failures with message 
```bash
Can't run because no spec files were found.

We searched for any files matching this glob pattern:

src/extensions/advanced-controls/**/*.cypress.js


Exited with code exit status 1
```

This extension does not have any E2E tests thus the error reported above. It seems that the script should return 0 if there are no specs. In your opinion, should the logic within the script be adjusted in this PR? Or in another issue/PR?

EvanHerman:
@AnthonyLedesma I would say let's update it here in this PR. You are right, since no blocks were updated, no .cypress.test files were added to the run. Technically it shouldn't try and run at all if no test files are altered.

.dev/bin/setup-test-specs.sh
```diff
@@ -27,9 +27,13 @@ do
   if [[ $FILE == *"src/extensions/"* ]]; then
     testSpec=$(echo $FILE | cut -d'/' -f3)
 		foundwords=$(echo ${SPECS[@]} | grep -o "${testSpec}" | wc -w)
+		# Catch cases where no cypress spec file exists. 
+			if [[ $(ls -l src/extensions/${testSpec}/**/*.cypress.js | wc -l) -eq 0 ]]; then
```
EvanHerman, line 31: @AnthonyLedesma These lines can probably be brought back one tab, so it's inline with `foundwords` on the line above.

AnthonyLedesma: Good call! Fixed.

PR #1619, opened 2020-08-07
Introduce support for custom gutter controls

jonathanbardo:
### Description
Add supports for custom gutter controls for all blocks that supports them. 

### Screenshots
custom-gutter

### Types of changes

Refactor of how blocks used to implement gutters. Added new HOC and new Generic Wrappers for inspectors.

Fix an issue where selecting layered style in collage block and then selecting the default style would break the gutter function.
Fix var declaring when scope collision was possible

New addition of coblocks-icons
New babel rules to enable support for class properties

### Checklist:
- [x] My code is tested


src/components/gutter-control/gutter-wrapper.js
```diff
@@ -0,0 +1,52 @@
...
+	if ( children.props.style ) {
+		style = {
+			...children.props.style,
+		};
+	}
+
+	if ( 'custom' === gutter ) {
+		style = {
+			...style,
+			'--coblocks-custom-gutter': `${ gutterCustom }em`,
```
jrtashjian, line 33: Block validation is failing when a custom gutter has been applied. You can see this by adding a block which supports gutters, setting a custom gutter value, saving and reloading the screen.

The style prop is always rendering the default value when it expects the custom value. This is the error message:

```
Content generated by `save` function:

<div class="wp-block-coblocks-services"><div class="has-columns has-2-columns has-responsive-columns has-custom-gutter" style="--coblocks-custom-gutter:1.6em"></div></div>

Content retrieved from post body:

<div class="wp-block-coblocks-services"><div class="has-columns has-2-columns has-responsive-columns has-custom-gutter" style="--coblocks-custom-gutter:5em">

</div></div>
```

The block validation process takes the block html in the post_content and runs the detectected attributes through the save function. If both the retrieved output and the save funciton's output are the same the block will pass validation.

package.json
```diff
@@ -13,6 +13,7 @@
 	},
 	"scripts": {
 		"setup": "composer install && yarn install && yarn run build && wp-scripts env install",
+		"start:dev": "NODE_ENV=development yarn start",
```
jrtashjian, line 16: This is not needed as the package we use - [`@wordpress/scripts`](https://github.com/WordPress/gutenberg/tree/wp/5.4/packages/scripts) - which defaults to `NODE_ENV=development` when `yarn start` is used and `NODE_ENV=production` when `yarn build`.

We need to upgrade this setup to the newest version in the future but the purpose is to stick to the core Gutenberg project standards as much as we can.

src/components/gutter-control/gutter-wrapper.js
```diff
@@ -0,0 +1,52 @@
...
+
+	let style = {};
+
+	if ( children.props.style ) {
+		style = {
+			...children.props.style,
+		};
+	}
+
+	if ( 'custom' === gutter ) {
```
jrtashjian, line 30: The `condition` prop should also determine whether we apply the style prop to the element like it decides the addition of our className below.

jonathanbardo: @jrtashjian are you talking about not having empty style attribute?

jrtashjian: yeah, we don't want to render an attribute we're not using.

src/components/gutter-control/gutter-wrapper.js
jonathanbardo: @jrtashjian turns out I was saying the attribute is a string (because eventually I want it to store a number and the unit) and I was saving a number because that's the output of <RangeControl/>. All fixed in latest commit.

src/components/gutter-control/gutter-wrapper.js
jonathanbardo: Fixed in 7e66cfb

richtabor:
Updated styling to: 
Screen Shot 2020-08-24 at 1 49 11 PM

richtabor:
Minor tweaks to the styling. Decided to keep it blue for now, as that what core is doing. I did add a check if the custom option is active, to properly space the range control below it. 

Screen Shot 2020-08-24 at 2 05 55 PM

PR #1622, opened 2020-08-11
Fix Icon block invalid block content within the editor

AnthonyLedesma:
### Description
Icon block may face conditions where SVG attributes exist that have not been accounted for in save or deprecated save functions leading to invalid block errors within the editor. 

This PR introduces a deprecated function to the Icon block which will catch missing SVG attributes and apply them where necessary. This change allows for successful block deprecation in the case described above.

### Types of changes

Bug Fix.
JavaScript changes.

### How has this been tested?

Tested manually using content containing invalid block content.

### Checklist:
- [x] My code is tested


src/blocks/icon/deprecated/svgsfocusable.js
```diff
@@ -0,0 +1,1246 @@
...
+			__( 'software', 'coblocks' ),
+			/* translators: icon keyword */
+			__( 'upload', 'coblocks' ),
+		],
+		icon: <svg height="20" focusable="false" role="img" aria-hidden="true" viewbox="0 0 20 20" width="20" xmlns="http://www.w3.org/2000/svg"><path d="m19.9986483 8.20489366h-7.5333334l3.0444445-3.13333333c-3.0333334-3-7.94444447-3.11111111-10.97777781-.11111111-3.03333333 3.01111111-3.03333333 7.86666668 0 10.87777778 3.03333334 3.0111111 7.94444441 3.0111111 10.97777781 0 1.5111111-1.4888889 2.2666666-3.2333333 2.2666666-5.4333333h2.2222223c0 2.2-.9777778 5.0555555-2.9333334 6.9888889-3.9 3.8666666-10.23333331 3.8666666-14.13333331 0-3.88888889-3.8555556-3.92222222-10.12222227-.02222222-13.97777783 3.9-3.85555555 10.15555553-3.85555555 14.05555553 0l3.0333334-3.12222222zm-9.4444445-2.35555555v4.72222219l3.8888889 2.3111111-.8 1.3444445-4.75555555-2.8222222v-5.55555559z" transform="translate(0 -.293783)" /></svg>,
+	},
+};
+
+Object.entries( icons ).filter( function( item ) {
+	svgs.outlined[ item[ 0 ] ] = {
```
jonathanbardo, line 1233: @AnthonyLedesma It seems we are always accessing the item[0] here and it makes the code a bit hard to read. I would assign it to a const right before `cons elem = item[0]` and `icons[ item[0] ]` seems to be checked in lots of place I would do the same.

src/blocks/icon/deprecated/svgsfocusable.js
```diff
@@ -0,0 +1,1246 @@
...
+			/* translators: icon keyword */
+			__( 'software', 'coblocks' ),
+			/* translators: icon keyword */
+			__( 'upload', 'coblocks' ),
+		],
+		icon: <svg height="20" focusable="false" role="img" aria-hidden="true" viewbox="0 0 20 20" width="20" xmlns="http://www.w3.org/2000/svg"><path d="m19.9986483 8.20489366h-7.5333334l3.0444445-3.13333333c-3.0333334-3-7.94444447-3.11111111-10.97777781-.11111111-3.03333333 3.01111111-3.03333333 7.86666668 0 10.87777778 3.03333334 3.0111111 7.94444441 3.0111111 10.97777781 0 1.5111111-1.4888889 2.2666666-3.2333333 2.2666666-5.4333333h2.2222223c0 2.2-.9777778 5.0555555-2.9333334 6.9888889-3.9 3.8666666-10.23333331 3.8666666-14.13333331 0-3.88888889-3.8555556-3.92222222-10.12222227-.02222222-13.97777783 3.9-3.85555555 10.15555553-3.85555555 14.05555553 0l3.0333334-3.12222222zm-9.4444445-2.35555555v4.72222219l3.8888889 2.3111111-.8 1.3444445-4.75555555-2.8222222v-5.55555559z" transform="translate(0 -.293783)" /></svg>,
+	},
+};
+
+Object.entries( icons ).filter( function( item ) {
```
jonathanbardo, line 1232:
```suggestion
Object.entries( icons ).filter( ( item ) => {
```

jonathanbardo, line 1232: Whenever I see loops over object like that I would always wonder if it would have been cleaner to just use a Set/ Map instead. It has cleaner API and it seems like the usecase here would be appropriate.

AnthonyLedesma: I think the Set/Map API is interesting and it likely should be implemented in cases where we are using `Object.entries` instead. I went ahead and cleaned up that code to help the readability.

src/blocks/icon/deprecated/svgsfocusable.js
AnthonyLedesma: Good call here @jonathanbardo. I have gone ahead and made a few changes for readability. Now we are using optional chaining to reduce the ternary operators into `logical or` operators. Also setup `const elem = item[ 0 ]`to reference the looped element and switched to use `.forEach` iteration method because `.filter` expects return values whereas `.forEach` does not.

PR #1638, opened 2020-08-21
Introduce Posts and Post Carousel block deprecations

### LINKED ISSUE #1632: Post blog category filter not working
### Describe the bug:
I should be able to filter the posts by category using "Posts" block

### To reproduce:

1. Edit the page 
2. Add Posts block
3. Choose a category in the Feed settings 
4. See error on the fronted - category filter is not applied - all posts are displayed instead. However you can see the correct posts in the backend. 

### Expected behavior:
Posts from selected category should be displayed 

### Screenshots:
http://prntscr.com/u23e6v
http://prntscr.com/u23f9b
by the way, featured image is displayed over the post title - http://prntscr.com/u23ew9

### Isolating the problem:

- [ x] This bug happens with no other plugins activated
- [x ] This bug happens with a default WordPress theme active
- [ x] This bug happens **without** the Gutenberg plugin active
- [x ] I can reproduce this bug consistently using the steps above

### WordPress version:
WordPress 5.5

AnthonyLedesma:
### Description

Closes #1632 
Following the lead of the Gutenberg team, this PR introduces measures that will allow proper migration/deprecation from previously configured and published blocks into a working state using modern interface and features. 

In this case, the `core/latest-posts` block had a number of breaking changes related to the implementation of a tokenized category selector. The new selector requires new logic and a new attribute schema.

This change is inspired by this [PR in Gutenberg](https://github.com/WordPress/gutenberg/blob/d0ca2551fdaa460d1ce4dc0a3bf0125ea4dbdba2/packages/block-library/src/latest-posts/index.php#L50). This PR resolves a bug with migration from old-style `coblocks/posts` and `coblocks/post-carousel` attributes.

This can be replicated using old markup:
```
<!-- wp:coblocks/post-carousel {“columns”:3,“categories”:“1”} /-->
OR
<!-- wp:coblocks/post-carousel {"categories":"4"} /-->

<!-- wp:coblocks/posts {“className”:“is-style-stacked”,“categories”:“3”} /-->
```
**Please note** This fix will successfully migrate the chosen category from a previously configured block into a category token within the new interface. While the token exists and works as expected as a filter, the token is missing contextual language. This is a [known issue with migration](https://github.com/WordPress/gutenberg/pull/21359#issuecomment-615273139).
image

### Types of changes

JavaScript changes and PHP changes.

### How has this been tested?

Tested manually migrating from 5.4.2 -> 5.5
Tested with and without the Gutenberg plugin.
Tested migration from CoBlocks 1.23.0 -> 2.2.2

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/post-carousel/index.php
+++ b/src/blocks/post-carousel/index.php
@@ -341,3 +341,35 @@ function coblocks_register_post_carousel_block() {
 	);
 }
 add_action( 'init', 'coblocks_register_post_carousel_block' );
+
+
+
+/**
+ * Handles outdated versions of the `coblocks/post-carousel` block by converting
+ * attribute `categories` from a numeric string to an array with key `id`.
+ *
+ * This is done to accommodate the changes introduced in https://github.com/WordPress/gutenberg/pull/20781 that sought to
+ * add support for multiple categories to the block. However, given that this
+ * block is dynamic, the usual provisions for block migration are insufficient,
+ * as they only act when a block is loaded in the editor.
+ *
+ * TODO: Remove when and if the bottom client-side deprecation for this block
+ * is removed.
+ *
+ * @param array $block A single parsed block object.
+ *
+ * @return array The migrated block object.
+ */
+function block_coblocks_post_carousel_migrate_categories( $block ) {
+	if (
+		'coblocks/post-carousel' === $block['blockName'] &&
+		! empty( $block['attrs']['categories'] ) &&
+		is_string( $block['attrs']['categories'] )
+	) {
+		$block['attrs']['categories'] = array(
+			array( 'id' => absint( $block['attrs']['categories'] ) ),
+		);
+	}
+	return $block;
+}
+add_filter( 'render_block_data', 'block_coblocks_post_carousel_migrate_categories' );
--- a/src/blocks/post-carousel/index.php
+++ b/src/blocks/post-carousel/index.php
@@ -27,7 +27,7 @@ function coblocks_render_post_carousel_block( $attributes ) {
 
 	if ( isset( $attributes['categories'] ) ) {
 
-		$args['category'] = $attributes['categories'];
+		$args['category__in'] = array_column( $attributes['categories'], 'id' );
 
 	}
 
--- a/src/blocks/posts/deprecated.js
+++ b/src/blocks/posts/deprecated.js
@@ -1,0 +1,36 @@
+/**
+ * Internal dependencies
+ */
+import metadata from './block.json';
+
+const { attributes } = metadata;
+
+export default [
+	{
+		attributes: {
+			...attributes,
+			categories: {
+				type: 'string',
+			},
+		},
+		supports: {
+			align: true,
+			html: false,
+		},
+		migrate: ( oldAttributes ) => {
+			// This needs the full category object, not just the ID.
+			console.log( oldAttributes );
+			console.log( {
+				...oldAttributes,
+				categories: [ { id: Number( oldAttributes.categories ) } ],
+			} );
+			return {
+				...oldAttributes,
+				categories: [ { id: Number( oldAttributes.categories ) } ],
+			};
+		},
+		isEligible: ( { categories } ) =>
+			categories && 'string' === typeof categories,
+		save: () => null,
+	},
+];
--- a/src/blocks/posts/deprecated.js
+++ b/src/blocks/posts/deprecated.js
@@ -19,11 +19,6 @@ export default [
 		},
 		migrate: ( oldAttributes ) => {
 			// This needs the full category object, not just the ID.
-			console.log( oldAttributes );
-			console.log( {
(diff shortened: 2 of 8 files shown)


src/blocks/post-carousel/index.php
```diff
@@ -341,3 +341,35 @@ function coblocks_register_post_carousel_block() {
...
+		! empty( $block['attrs']['categories'] ) &&
+		is_string( $block['attrs']['categories'] )
+	) {
+		$block['attrs']['categories'] = array(
+			array( 'id' => absint( $block['attrs']['categories'] ) ),
+		);
+	}
+	return $block;
+}
+add_filter( 'render_block_data', 'block_coblocks_post_carousel_migrate_categories' );
```
jrtashjian, line 375: I don't think we actually need to do this. Instead we should handle the existence of both keys and manage it within `coblocks_render_post_carousel_block` which I believe you already are.

src/blocks/posts/deprecated.js
```diff
@@ -0,0 +1,31 @@
...
+				type: 'string',
+			},
+		},
+		supports: {
+			align: true,
+			html: false,
+		},
+		migrate: ( oldAttributes ) => {
+			// This needs the full category object, not just the ID.
+			return {
```
jrtashjian, line 22: Let's query the REST API for this additional information (left a code snippet in another comment).

src/blocks/posts/index.php
```diff
@@ -329,3 +329,33 @@ function coblocks_register_posts_block() {
...
+		! empty( $block['attrs']['categories'] ) &&
+		is_string( $block['attrs']['categories'] )
+	) {
+		$block['attrs']['categories'] = array(
+			array( 'id' => absint( $block['attrs']['categories'] ) ),
+		);
+	}
+	return $block;
+}
+add_filter( 'render_block_data', 'block_coblocks_posts_migrate_categories' );
```
jrtashjian, line 361: I don't think we actually need to do this. Instead we should handle the existence of both keys and manage it within coblocks_render_posts_block which I believe you already are.

src/blocks/post-carousel/index.php
AnthonyLedesma: You are saying that I should not use the filter and instead place this logic elsewhere? Within the function `coblocks_render_post_carousel_block`? 

The logic to migrate from string type attribute to array type attribute does not presently exist meaning that any configured categories do **not** function correctly after migration.

AnthonyLedesma: @jrtashjian  This filter will handle the migration on published pages without the need for the user to enter the editor and update markup. As a result, without the filter, previously configured and saved blocks that once properly filtered by category on the front-end no longer function as expected. This filter is for that specific case. 

Otherwise, you are correct that the deprecated migration function handles the migration type transform from string to an array of objects. This **does** require a user to enter the editor, however.

AnthonyLedesma: With Filter Works
image

Without Filter Broken
withoutPHPFilterbroken

src/blocks/posts/deprecated.js
AnthonyLedesma: Having issues getting the fetch results saved properly. There seems to be a race condition of sorts where this migration function will not await the fetch results.

src/blocks/post-carousel/index.php
AnthonyLedesma: A bit more about the need for the attribute migration that is not handled in https://github.com/godaddy-wordpress/coblocks/blob/67f84b3e1911fddcfcba23c184b62dc3bfd4c40b/src/blocks/post-carousel/index.php#L28-L32

https://github.com/godaddy-wordpress/coblocks/blob/67f84b3e1911fddcfcba23c184b62dc3bfd4c40b/src/blocks/post-carousel/block.json#L63-L68
```
<!-- wp:coblocks/post-carousel {"categories":"4"} /-->
```
I suspect that due to changes within the block.json attribute schema the classic attribute syntax saved into post content is not properly parsed into the new format. In this case above the `$attributes['categories']` is unset and treated as undefined. The filter accomplishes the goal of correctly parsing the attribute from post content thus allowing migration from categories attribute typed string to categories attribute typed as an array of objects.

src/blocks/posts/deprecated.js
jrtashjian: Marking this resolved since we've decided this missing data is a minor issue which does not break the actual functionality of the block.

src/blocks/post-carousel/index.php
jrtashjian: Marking this resolved since the original solution is sufficient.

src/blocks/posts/index.php
jrtashjian: Marking this resolved since the original solution is sufficient.

jrtashjian (approved):

PR #1636, opened 2020-08-20
Introduce improved E2E testing for layouts and image downloader

AnthonyLedesma:
### Description

This PR introduces layouts specific for testing and logic introduced to the layout selector which allows proper functionality when running tests.

### Types of changes

JavaScript changes to logic and added sample testing layouts that are used only for e2e testing.

### How has this been tested?

Tested manually within the browser.

### Checklist:
- [x] My code is tested
- [x] My code follows accessibility standards 
- [x] My code has proper inline documentation 
- [x] I've included any necessary tests 
- [x] I've included developer documentation 
- [x] I've added proper labels to this pull request

PR #1642, opened 2020-08-24
Introduce improved E2E testing for layouts and image downloader

AnthonyLedesma:
### Description

This is round two attempt at implementing proper e2e tests for image downloader functionality built into the layout selector. The original attempt was made in [PR 1636](https://github.com/godaddy-wordpress/coblocks/pull/1636) which was subsequently closed due to the drawbacks with the proposed solution. This PR managed to get around the issues that existed in #1636

This PR also changes .eslintignore to properly exclude/include the appropriate test files.
Upgrade to Cypress 5.2.0.
Configure native test retries.

### Types of changes

JavaScript changes. 

### How has this been tested?

Tested manually using the e2e tests.

### Checklist:
- [x] My code is tested


AnthonyLedesma:
Looking at the failing tests, it appears that the failures can be attributed to changes in the DOM as a result of WordPress updates from the core, particularly for 5.5. My opinion is that these test failures are out of the scope of this specific PR due to the fact that the failures will occur on the master branch and are not related to code changes here. 

I propose these issues with failing tests be resolved in a separate pull request.

jrtashjian (dismissed):
This looks great but the `does not open modal when disabled via the "Editor Settings" panel` test seems to be failing. By manually going through it, they layout selector modal is immediately triggered when re-enabling the setting in the "Editor Settings" modal. We need to make it so that doesn't happen in this case.

Maybe we can add some actual content to this page and publish it so that doesn't happen. We'll need to fix this bug though in the case of a brand new page.

jrtashjian:
I also updated this branch with the latest from the base branch.

AnthonyLedesma:
Base branch and PR have bugs with layout selector. Test failures are related to base branch changes. Cannot move forward on this until we have progress on https://github.com/godaddy-wordpress/coblocks/pull/1607.

jrtashjian:
Fixed the issues from the base branch and pulled those changes into here. I also made a few updates to how the layouts are loaded because I wasn't getting consistent results.

All of the tests pass locally using Go and TwentyTwenty but they seem to be failing in Cypress. I'm not sure why but it looks like the layouts are still not being loaded into the `window.coblocksLayoutSelector` property in time. I just pushed some additional updates so we'll see if they still fail.

AnthonyLedesma:
Concerning your review regarding 

> layout selector modal is immediately triggered when re-enabling the setting in the "Editor Settings" modal.

If this is a bug with the behavior of the Layout Selector, I propose that both the layout selector behavior bug and tests for the editor settings controls should be updated separately. That way we can create a request scoped to that change. Thoughts @jrtashjian?

AnthonyLedesma:
Should we try to remove the recently added step to install Go for e2e testing? It is effecting tests on the [Social Profiles and Share blocks for example](https://dashboard.cypress.io/projects/sovnn2/runs/868/failures).

jrtashjian (approved):

PR #1656, opened 2020-09-01
Remove arrows from lightbox in single image galleries

### LINKED ISSUE #1601: Hide lightbox arrows when gallery contains only one image
Hi,

Pretty clear idea from the title, I think? Currently the next/previous arrows show up even with just one image, but there is probably the use case where you'd like either the fullscreen functionality of lightbox for just one image, or would like to have the same sort of design across several galleries (e.g. 'stacked', or 'carousel'), but one of them has only one image... 

Nothing huge, I wonder if people have thought of that case or might be interested in that?

Cheers!

EvanHerman:
### Description
Remove gallery lightbox arrows when only one image is used in a gallery.

Resolves #1601 

### Types of changes
Non-breaking change

### How has this been tested?
Manually. Created one gallery of each type with only one image. Each gallery had the `lightbox` property turned on.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/js/coblocks-lightbox.js
+++ b/src/js/coblocks-lightbox.js
@@ -58,6 +58,10 @@
 
 		if ( images.length > 0 ) {
 			document.getElementsByTagName( 'BODY' )[ 0 ].append( wrapper );
+			if ( images.length === 1 ) {
+				arrowRightContainer.remove();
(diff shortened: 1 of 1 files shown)


jrtashjian (approved):
Looks good!

PR #1657, opened 2020-09-01
Fix cursor hover state on offset/collage blocks

### LINKED ISSUE #1644: ISBAT see correct cursor when hovering Offset/Collage image
### Describe the bug:

In the editor, when hovering an image in the Offset/Collage gallery block that has the Lightbox setting on, the cursor is set to a magnifying glass (`zoom-in`). This can be misleading as nothing will happen when clicked. Note that it doesn't happen for Masonry block.

### To reproduce:

1. Add Offset/Collage blocks and add some photos
2. Set the Lightbox setting on
3. Hover over one of the images
4. Notice the cursor changing to a magnifying glass

### Expected behavior:
Zooming is possible on preview/published site only so the cursor shouldn't change in the editor.

### Screenshots:

Screenshot 2020-08-25 at 14 47 20

### WordPress version:
5.5

### Gutenberg version:
8.7.1

EvanHerman:
### Description
Fix the cursor hover state on offset and collage blocks.

Resolves #1644 

### Types of changes
New style (non-breaking change which adds functionality)

### How has this been tested?
Manually tested in the editor.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/gallery-masonry/styles/editor.scss
+++ b/src/blocks/gallery-masonry/styles/editor.scss
@@ -93,12 +93,8 @@
 	}
 }
 
-.has-lightbox {
-
-	.coblocks-gallery--figure:hover,
-	.coblocks-gallery--item figure:hover {
-		cursor: default !important;
-	}
+.has-lightbox figure {
+	cursor: default !important;
 }
 
 // Needed until GB 6.9 is merged into core or backward support.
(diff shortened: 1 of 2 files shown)


jrtashjian (approved):
Looks good. I just moved these styles into the extensions dir where the lightbox controls live. I also changed the selected to force all `figures` within a `.has-lightbox` to use the default cursor (editor only).

PR #1660, opened 2020-09-03
Fix variation titles for Row and Form blocks

richtabor:
### Description
Variations in our Row and Form blocks are not displaying the labels associated with each variation. Upon investigation, I uncovered that each variation's label should be defined with a `title` key, instead of label. 

#### Before
Screen Shot 2020-09-03 at 1 30 36 PM

#### After
Screen Shot 2020-09-03 at 1 27 47 PM

### Types of changes


AnthonyLedesma (approved):
LGTM

PR #1671, opened 2020-09-09
Pricing Table: Use ToolbarGroup to render controls correctly

### LINKED ISSUE #1670: ISBAT use Pricing Table columns control in Gutenberg v8.8.0=<
### Describe the bug:
Pricing Table's columns control is unavailable for Gutenberg releases since v8.8.0.

Since v8.8.0, Toolbar's `controls` prop is ignored if the `label` prop is passed along. This change happened [here](https://github.com/WordPress/gutenberg/pull/23316/files#diff-2a931797916e4fd724b30446c67121acR28-R49). I could confirm it locally by simply removing the `label` prop which made the option columns option appear. While this seems like a quick-fix for this bug, not passing `label` is now deprecated. ~Converting controls to `ToolbarButtons` would probably be a better idea here?~ Edit: Looks like switching Toolbar with ToolbarGroup [does the trick](https://github.com/godaddy-wordpress/coblocks/pull/1671).

### To reproduce:

1. Start local env or go to your site
2. Install Gutenberg >=v8.8.0
3. Go to editor
4. Add Pricing Table block
5. Select the block and see that there is no columns control:
   image
6. You can check that this option is available for releases <=v8.7.1, or by [replacing Toolbar with ToolbarGroup](https://github.com/godaddy-wordpress/coblocks/blob/master/src/blocks/pricing-table/controls.js#L60) if you're debugging in local env: 
   image

### Expected behavior:

Pricing Table columns option to be available.

### Isolating the problem:

### WordPress version:
`5.5.1`

### Gutenberg version:
`>=8.8.0`

WunderBart:
Fixes #1670

### Description
Use ToolbarGroup instead of Toolbar to render Pricing Table controls correctly. This fixes a bug that's present when CoBlocks are used with Gutenberg >=8.8.0. This solution is based on how the [Gutenberg's alignment toolbar is implemented](https://github.com/WordPress/gutenberg/blob/v8.9.3/packages/block-editor/src/components/block-alignment-toolbar/index.js#L69-L86).

### Screenshots
| Before | After |
| ------------- | ------------- |
| image | image |

### Types of changes

### How has this been tested?
Confirmed happening with activated Gutenberg plugin v8.8.0 and later.

### DIFF
--- a/src/blocks/pricing-table/controls.js
+++ b/src/blocks/pricing-table/controls.js
@@ -9,7 +9,7 @@ import { find } from 'lodash';
 import { __ } from '@wordpress/i18n';
 import { Component, Fragment } from '@wordpress/element';
 import { AlignmentToolbar, BlockControls } from '@wordpress/block-editor';
-import { Toolbar, Path, SVG } from '@wordpress/components';
+import { ToolbarGroup, Path, SVG } from '@wordpress/components';
 
 const DEFAULT_ALIGNMENT_CONTROLS = [
 	{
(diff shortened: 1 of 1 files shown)


richtabor (approved):

richtabor:
Good stuff! I found a few more that are doing the same thing. We'll get those cleaned up.

PR #1672, opened 2020-09-09
Remove unnecessary SCSS variables and use @wordpress/base-styles instead

richtabor:
I've combed through, removed unused variables, updated variables to use the proper ones from @wordpress/base-styles and cleaned up some of the SCSS.


jrtashjian (approved):
This looks good other than the failing tests which I'm not sure why these are failing. I've just pulled in the latest from the master branch since we have made some updates to tests there.

Approving since these failing tests have nothing to do with the changes. However, **before merging** let's see if we can resolve the tests or at least get some more information about why they are failing.

EvanHerman:
For some reason the layout selectors aren't showing in the e2e tests 🤔

EvanHerman (approved):
Looks great to me. I'm not sure why the layout selector isn't showing in the e2e test though. I will try and do some digging before we merge this, and see what I can find. Otherwise we can just find a fix in another PR.

richtabor:
I resolved the merge conflicts with master

PR #1673, opened 2020-09-09
HeadingToolbar: Use ToolbarGroup to render controls correctly

richtabor:
### Description
Use ToolbarGroup instead of Toolbar to render HeadingToolbar controls correctly. This fixes a bug that's present when CoBlocks are used with Gutenberg >=8.8.0. This solution is based on how the Gutenberg's alignment toolbar is implemented.

### Screenshots
Before: 
Screen Shot 2020-09-09 at 1 35 35 PM

After:
Screen Shot 2020-09-09 at 1 35 16 PM

### Types of changes

### How has this been tested?
Confirmed happening with activated Gutenberg plugin v8.8.0 and later - and with core WordPress.

### DIFF
--- a/src/components/heading-toolbar/index.js
+++ b/src/components/heading-toolbar/index.js
@@ -13,7 +13,7 @@ import HeadingLevelIcon from './icon';
  */
 import { __, sprintf } from '@wordpress/i18n';
 import { Component } from '@wordpress/element';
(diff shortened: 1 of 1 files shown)


richtabor:
Same issue/solution as https://github.com/godaddy-wordpress/coblocks/pull/1671.

PR #1677, opened 2020-09-09
Fix required checkbox/radio hidden message escaped markup

EvanHerman:
### Description
There was a bug with the output of the required text, which should be hidden on initial load. Because we were passing the HTML through `esc_html()` the markup was not properly parsed on the front of site. This PR switches things to use `wp_kses_post()`  which parses the HTML as expected.

#### Before
image

#### With Patch
image

### Types of changes

### How has this been tested?
Manually tested.

### Checklist:
- [x] My code is tested


AnthonyLedesma (approved):
Confirmed to resolve the issue. LGTM.

PR #1676, opened 2020-09-09
Add form field transformations

EvanHerman:
### Description
Add the ability to transform between form fields/blocks.
- Name, Date, Textarea, Phone, Text, Website, Hidden
- Select, Checklist, Radio

### Screenshots
image

### Types of changes
New feature (non-breaking change which adds functionality)

### How has this been tested?
Manually tested each form field.

### Checklist:
- [x] My code is tested
- [x] My code follows accessibility standards 
- [x] My code has proper inline documentation 
- [x] I've added proper labels to this pull request

PR #1680, opened 2020-09-14
Transform Form fields with with proper labels

richtabor:
### Description
This PR builds on #1676 by adapting the transformed field label to it's original label. For example, if transforming a Date block into a Website block, the label "Date" becomes "Website".

### DIFF
--- a/src/blocks/form/fields/checkbox/transforms.js
+++ b/src/blocks/form/fields/checkbox/transforms.js
@@ -1,6 +1,7 @@
 /**
  * WordPress dependencies
  */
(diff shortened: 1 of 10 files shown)


EvanHerman:
@richtabor Looks fine and works well on my end 👍. I had considered this in the original PR but just chose to preserve the labels after a transform.

EvanHerman (approved):
@richtabor Looks fine and works well on my end 👍. I had considered this in the original PR but just chose to preserve the labels after a transform.

PR #1683, opened 2020-09-15
Tweak Form label color controls

AnthonyLedesma:
### Description
I had a realization that I was using a superfluous block exclusion here where it is not required. This behavior is already handled here - except for cases where the same attribute schema exists regardless of defined block supports.
https://github.com/godaddy-wordpress/coblocks/blob/4db325197106746692240a29d64818adcb9b4fe6/src/components/form-label-colors/index.js#L7-L11

Because we might end up altering attributes of children blocks that do not have explicit form label support, this component has been modified to allow attribute propagation to only blocks with defined supports. (form child blocks)

This change prevents colors from being set on a child paragraph block within the Form block.

### Screenshots
Exclude blocks without defined support.

formLabelColorsHidden

### Types of changes

Remove and replace JavaScript logic. This change will use defined block supports to determine attribute propagation.

### How has this been tested?

Tested manually. E2E.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/components/form-label-colors/label-color-control.js
+++ b/src/components/form-label-colors/label-color-control.js
@@ -58,15 +58,12 @@ export default compose( [
 
 		const updateInnerAttributes = ( newAttributes ) => {
 			const innerItems = getBlocksByClientId(	props.clientId	)[ 0 ].innerBlocks;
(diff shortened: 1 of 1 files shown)


richtabor (approved):
LGTM

PR #1691, opened 2020-09-24
Update/gallery innerblocks arrow orientation

EvanHerman:
### Description
Set the gallery block to display the correct move arrow orientation depending on if things are set to 1 or more columns.

### Screenshots
image
image

### Types of changes

### How has this been tested?
Manually tested. We should probably add tests.

### Checklist:
- [x] My code is tested


EvanHerman:
For some reason the `Test services block saves with heading level set.` test is failing. The header is not being set to the correct element in the test, but in my manual test it does look like it's set properly.

jasonlemay (approved):

PR #1715, opened 2020-10-05
Add e2e test utils for Cypress

jrtashjian:
### Description
Adding a library to standardize and organize our repeatable actions interacting with the Block Editor within our Cypress E2E tests. This library will mostly duplicate the API of the [`@wordpress/e2e-test-utils`](https://github.com/WordPress/gutenberg/tree/master/packages/e2e-test-utils) package which was built specifically for Puppeteer. The plan would be to get near 100% parity with the existing API and to extract this into it's own package for public use (once ready).

The only utilities I've added so far are those that interact with the global block inserter to search for and insert a block. I've tested this specifically with the existing Alert block tests and all are passing as if nothing changed. Here is [the current API](https://github.com/godaddy-wordpress/coblocks/blob/6bfcf13062a4c9c0918727e51c5356a3fb56ab8e/.dev/tests/cypress/e2e-test-utils/README.md) docs (auto generated using [`@wordpress/docgen`](https://github.com/WordPress/gutenberg/tree/master/packages/docgen)).

Why am I doing this? Cypress has been frustratingly inconsistent and we need to do a better job of not writing duplicate code or targeting selectors specific to the block editor in multiple places. These can and have changed in other releases of the block editor and has been time consuming to update. Basically, I want to stop testing Gutenberg and focus our tests on CoBlocks.

### Types of changes
New testing tools for Cypress


jrtashjian:
Closing this as we may be moving away from Cypress.

PR #1723, opened 2020-10-06
Fix image preview regressions

EvanHerman:
### Description
- Fix regressions introduced in https://github.com/godaddy-wordpress/coblocks/pull/1721
- Fix bug with image preview image opacity being set to `0.5`.

### Types of changes

### How has this been tested?
Manually tested in the editor with and without Gutenberg active.

### Checklist:
- [x] My code is tested


src/components/media-filter-control/styles/editor.scss
```diff
@@ -0,0 +1,26 @@
+.components-coblocks-media-filter {
+
+	&.has-filter {
+		position: relative;
+
+		::after {
```
jasonlemay, line 6: Is this working? I was sure you needed `&::after {`

richtabor: It adds the little triangle that we use for the animations control as well - when an image filter is added.

jasonlemay: But the `&` isn't required? Good to know!

PR #1724, opened 2020-10-07
Remove search keywords from Media Card block

AnthonyLedesma:
### Description

Remove keyword search terms from the Media Card block. Reducing searchability for the Media Card block but not yet making the block un-insertable.

Note: Since the name of the block is a searchable token, I would suggest registering Media Card block without the ability to insert if that behavior is desired. 
https://github.com/godaddy-wordpress/coblocks/blob/31938b07cf68b0910f408c5972ac794b5163e795/src/js/deprecations/deprecate-coblocks-buttons.js#L9-L25

### Types of changes

Minor JavaScript change.

### How has this been tested?

Tested manually

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/media-card/index.js
+++ b/src/blocks/media-card/index.js
@@ -36,13 +36,6 @@ const settings = {
 	/* translators: block description */
 	description: __( 'Add an image or video with an offset card side-by-side.', 'coblocks' ),
 	icon: <Icon icon={ icon } />,
-	keywords: [
-		'coblocks',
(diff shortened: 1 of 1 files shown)


EvanHerman (approved):
Looks fine to me 👌 
Simple change 👍

EvanHerman:
I think it's ok, for the time being, that it still shows up if you search `media card`. We'll be removing it soon enough.

PR #1732, opened 2020-10-14
Fix Features block column count controls

AnthonyLedesma:
### Description

The columns count controls had a few situations which could result in either `undefined` or a number > 4 being set for the columns attribute. This caused a number of strange behaviors most notably broken classnames. Another bug is that the inner-block count is not tied to the column count as it has been in the past. This is due to changes with core InnerBlocks component over time. 

This PR includes fixes for both of the bugs described above.

### Screenshots

featuresBlockColumnsCount

### Types of changes

JavaScript changes. Should not require any deprecations.

### How has this been tested?

E2E. Manually. With Gutenberg. Tested with 5.4.2 as well as 5.5.1.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/features/edit.js
+++ b/src/blocks/features/edit.js
@@ -23,23 +23,18 @@ import { compose } from '@wordpress/compose';
 import { InnerBlocks } from '@wordpress/block-editor';
 import { isBlobURL } from '@wordpress/blob';
 import { Spinner } from '@wordpress/components';
-import { dispatch, select } from '@wordpress/data';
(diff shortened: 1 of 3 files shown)


AnthonyLedesma:
These tests are passing locally in both Firefox and Chrome.
While this PR does not resolve the test failures completely, it does resolve bugged behavior and for that reason I sugest that we merge with failures.

richtabor (approved):

PR #1738, opened 2020-10-26
Enhance Layout Selector to allow Image Category selector

AnthonyLedesma:
### Description

Change markup and structure of Layout Selector.


src/extensions/layout-selector/index.js
```diff
@@ -309,20 +315,19 @@ class LayoutSelector extends Component {
...
-				title={ (
-					<Fragment>
-						{ __( 'Add new page', 'coblocks' ) }
-						<span>{ __( 'Pick one of these layouts or start with a blank page', 'coblocks' ) }</span>
-					</Fragment>
-				) }
-				onRequestClose={ () => {
-					this.useEmptyTemplateLayout();
-					closeTemplateSelector();
-				} }
```
jrtashjian, line 321: This title needs to stay for accesibility reasons and the `onRequestClose` is required for the click outside of the modal to default to an empty page.

For the change in sidebar and header treatment, I would use a CSS class to apply similar styles to each while still maintaining them as individual pieces. Think of the title and modal content as two distinct areas each with a sidebar.

Splitting them out into those different areas might make it more difficult for mobile styles (which aren't in a great shape right now).

src/extensions/layout-selector/index.js
```diff
@@ -309,20 +315,19 @@ class LayoutSelector extends Component {
...
 				className="coblocks-layout-selector-modal">
-
 				<div className="coblocks-layout-selector">
 					<aside className="coblocks-layout-selector__sidebar">
+						<h1 className="layout-selector-title">
+							{ __( 'Add new page', 'coblocks' ) }
+						</h1>
```
jrtashjian, line 323: Remove this. See comment about the modal title.

src/extensions/layout-selector/index.js
```diff
@@ -383,6 +388,18 @@ class LayoutSelector extends Component {
...
+							<Button
+								className="coblocks-layout-selector__close-button"
+								icon={ close }
+								onClick={ () => {
+									this.useEmptyTemplateLayout();
+									this.props.closeTemplateSelector();
+								} }
+								aria-label="Close dialog">
+							</Button>
+						</span>
```
jrtashjian, line 402: Remove this. See comment about the modal title.

jrtashjian (changes requested):
Looks pretty good. I've left a few comments about the modal changes and there are some styling issues that we may need @richtabor input on.

Styling is a bit off with the dropdown. It's offset when not opened and pulled backwhen it is open.

image

image

And if we just look at CoBlocks without NextGen the sidebar is compressed.

image

We need to make sure styles look as we expect in CoBlocks as an isolated piece.

jrtashjian:
An aside after talking with Rich: This category control addition should live in NextGen which means we need to make available a filter or slot/fill which allows NextGen to insert itself here.

AnthonyLedesma:
@jrtashjian Any idea why we have this test failure? Seems unrelated.

richtabor:
UI/X approved.

src/extensions/layout-selector/editor.scss
```diff
@@ -33,29 +33,54 @@
...
 		height: auto;
+		padding: 0;
 	}
 
 	.components-modal__header {
 		border-bottom: none;
 		height: 96px;
-		padding: 0 24px;
+		margin: 0;
+		padding: 0;
```
jrtashjian, line 50: I would only force `padding-left: 0` here otherwise the close button is hugging the side of the modal window.

src/extensions/layout-selector/editor.scss
```diff
@@ -33,29 +33,54 @@
 }
 
 .coblocks-layout-selector-modal {
+	background-color: #f5f5f5;
```
jrtashjian, line 36: Use the proper base-styles variable here.

src/extensions/layout-selector/editor.scss
```diff
@@ -33,29 +33,54 @@
...
-				margin-bottom: 13px;
-				margin-top: 0.8em;
+			.components-modal__header-heading {
+				background-color: #f5f5f5;
+				display: flex;
+				flex-direction: row;
+				font-size: 1rem;
+				height: 100%;
+				justify-content: start;
+				margin: 0;
```
jrtashjian, line 68: Use the proper base-styles variable here.

jrtashjian (changes requested):
Looks good. It would be great to add in the `@wordpress/library-export-default-webpack-plugin` and export these slot-fills as a global var accessible outside of CoBlocks so we don't need this filter. But that's something we can do another time.

jrtashjian:
> @jrtashjian Any idea why we have this test failure? Seems unrelated.

Yeah, this is unrelated but we can fix it here. It looks like it's not selecting the correct instance of the block in the post, it's pulling it from the style preview. Try altering the selector to this to fix the social block failure:

```
.edit-post-visual-editor .wp-block[data-type="coblocks/social"]
```

Not sure about the layout one, but I'd double check the test data we're asserting against.

src/extensions/layout-selector/editor.scss
AnthonyLedesma: @richtabor, Is there a proper base-style variable to use instead? Looking at `node_modules/@wordpress/base-styles/_colors.scss` I do not see that hex value. I suppose the color would be altered otherwise. Thoughts?

src/extensions/layout-selector/editor.scss
AnthonyLedesma: A slight difference here with the placement of the close button. I personally prefer the right aligned button. Any thoughts @richtabor?

**with `padding: 0`**
image

**with `padding-left: 0`**
image

src/extensions/layout-selector/editor.scss
richtabor: Use $gray-100 for now

src/extensions/layout-selector/editor.scss
richtabor: Let's keep it as is.

src/extensions/layout-selector/editor.scss
AnthonyLedesma: $gray-100

jrtashjian (approved):

AnthonyLedesma:
For posterity, we will be fixing failing tests in a separte pull request.

PR #1752, opened 2020-11-12
Add icons for animations UI

richtabor:
Screen Shot 2020-11-12 at 10 28 51 AM


richtabor:
We should add these icons to the coblocks-icons package and follow-up with an update to reference those icons.

jasonlemay (approved):

PR #1760, opened 2020-11-17
Add logic to prevent insertion of Media Card block

AnthonyLedesma:
### Description

Add logic to deprecate CoBlocks media-card block by preventing common block insertion. Since the block is still registered a user is allowed to duplicate or copy existing media-card blocks without issue. Another thing to consider here is that the media-card does not automatically transform to a media-text block as does happen from the `coblocks/buttons` -> `core/buttons` deprecation. 

### Screenshots

mediaCardDeprecated

### Types of changes

JavaScript changes to allow for block deprecation of the media-card block.

### How has this been tested?

Tested manually

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/js/deprecations/deprecate-coblocks-media-card.js
+++ b/src/js/deprecations/deprecate-coblocks-media-card.js
@@ -31,14 +31,7 @@ function deprecateCoBlocksMediaCardSettings() {
  * @return {boolean} True or false if block is deprecated.
  */
 export function mediaCardBlockDeprecated( coreMediaText, coBlocksMediaCard ) {
-	if ( ! coreMediaText ) {
-		return false;
-	}
-
-	if ( ! coBlocksMediaCard ) {
-		return false;
-	}
-	return true;
+	return ( ! coreMediaText || ! coBlocksMediaCard ) ? false : true;
 }
 
 domReady( deprecateCoBlocksMediaCardSettings );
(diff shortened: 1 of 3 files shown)


src/js/deprecations/deprecate-coblocks-media-card.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+export function mediaCardBlockDeprecated( coreMediaText, coBlocksMediaCard ) {
+	if ( ! coreMediaText ) {
+		return false;
+	}
+
+	if ( ! coBlocksMediaCard ) {
+		return false;
+	}
+	return true;
```
EvanHerman, line 41: @AnthonyLedesma We can probably combine the two checks here.

```js
if ( ! coreMediaText || ! coBlocksMediaCard ) {

   return false;

}
```

or maybe even:

```js
return ( ! coreMediaText || ! coBlocksMediaCard ) ? false : true;

EvanHerman (changes requested):
Looks good to me. Just one small change in `mediaCardBlockDeprecated()` and this should be good to merge 👍

src/js/deprecations/deprecate-coblocks-media-card.js
AnthonyLedesma: Good call. Keep it simple 👍

PR #1786, opened 2021-01-04
Add "most used" section to the layout selector.

jrtashjian:
### Description
A "most used" section has been added to the sidebar of the layout selector. This section works similar to how the most used blocks section of the block inserter

### Screenshots
image

### Types of changes

### How has this been tested?
Unit tests.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/extensions/layout-selector/store.js
+++ b/src/extensions/layout-selector/store.js
@@ -1,5 +1,10 @@
 /* global coblocksLayoutSelector */
 /**
+ * External dependencies
+ */
+import { kebabCase } from 'lodash';
+
+/**
  * WordPress dependencies
  */
 import { registerStore } from '@wordpress/data';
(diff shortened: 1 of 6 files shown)


src/extensions/layout-selector/store.js
```diff
@@ -51,9 +77,40 @@ const store = registerStore( 'coblocks/template-selector', {
...
+					default:
+						return count / 4;
+				}
+			};
+
+			return layouts.map( ( layout ) => {
+				const { time, count = 0 } = state.layoutUsage[ kebabCase( layout.label ) ] || {};
+				return {
+					...layout,
+					frequency: calculateFrequency( time, count ),
```
AnthonyLedesma, line 107: It seems that due to a number of reasons, this function runs a significant number of times within a single load. Since it's using `Date.now()` each subsequent run features a different duration time. If we compare to the Core implementation which features a cached response and thus a `Data.now()` value that can go back in time before the present execution context. 
https://github.com/WordPress/gutenberg/blob/e41e4f62074fac964d5c92e8836e826e90b289f7/packages/block-editor/src/store/selectors.js#L1439-L1440

calculatedDuration


Since we are receiving a recent timestamp there are no circumstances where the page execution time should venture beyond `MILLISECONDS_PER_HOUR`. Since we are not accounting for historical usage with this time selector, it makes much of this code unreachable. 

I am of the opinion that this logic should be simplified to apply static incremental updates to the layout selection frequency prop.

AnthonyLedesma (dismissed):

src/extensions/layout-selector/store.js
jrtashjian: I agree that this shouldn't be running at this kind of interval. I pulled the `calculateFrequency` function from core to keep the behavior consistent. The PER_HOUR/DAY/WEEK will eventually be hit because the algorithm is based on the current time and the last time a layout was used. So this code is rechable.

I will look into that subsequent run issue.

AnthonyLedesma: Cool. I see now where the time value is persisted into local storage. 👍

jrtashjian: After investigating this issue, pulling this function outside of the selector function block and memoizing it seemed to work best. I'm not really sure exactly why it was running so many times but maybe it's related to how the stores work.

Regardless, it only runs once per layout now which is what I expect.

image

AnthonyLedesma (approved):
LGTM

PR #1788, opened 2021-01-05
Bulk ESLint fixes

AnthonyLedesma:
### Description

ESLint rules have been updated over time but our project has remained untouched in many respects. This PR makes bulk recommended changes from ESLint to satisfy warnings.

### Types of changes

JavaScript ESLint

### How has this been tested?

Manually, E2E, Unit.

### Checklist:
- [x] My code is tested


.dev/tests/jest/setup-globals.js
```diff
@@ -10,7 +10,7 @@ global.wp = {};
 // Prevent console messages when running tests.
 console = {
 	...console,
-	log: () => { },
+	// log: () => { },
```
jrtashjian, line 13: This was enabled to prevent an abundance of messages specifically around successful block deprecations when running our unit tests. It would be better to determine if we still need this or not and remove it entirely if we don't.

jrtashjian (changes requested):
These changes look great and seem to resolve a number of test failures on the master branch which is great. It looks like there are still some test failures that we should look at before merging this in.

.dev/tests/jest/setup-globals.js
AnthonyLedesma: Good call JR. Pushed up some changes here. Most of the superfluous console messages are typed `info` and `warn` where we see block deprecation success messages as well as deprecated component usage notices. I suggest that we allow logging for dev debug logs when writing tests.

AnthonyLedesma:
Given a large number of fixes included within this PR, I propose that we merge into the main branch for the next release. Outstanding bugs will exist in main branch regardless of this pull request and can be handled in a scoped pull request.

.dev/tests/jest/setup-globals.js
```diff
@@ -12,7 +12,8 @@ console = {
 	...console,
 	// log: () => { },
 	info: () => { },
-	error: () => { },
+	warn: () => { },
+	// error: () => { },
```
jrtashjian, line 16: No need to leave error commented out here, just remove it.

jrtashjian (approved):
Minor change, approved otherwise.

PR #1790, opened 2021-01-11
Introduce custom padding controls to core/group block

AnthonyLedesma:
### Description

Modified Padding Controls extension to add custom controls.
Added styles and apply style variables to group elements.

### Screenshots

groupBlockCustomPadding

### Types of changes

JavaScript and styles changes.
No Deprecations needed.

### How has this been tested?

Tested manually in the editor.
E2E. 

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/extensions/padding-controls/index.js
+++ b/src/extensions/padding-controls/index.js
@@ -9,7 +9,7 @@ import { startCase } from 'lodash';
  */
 import { __, sprintf } from '@wordpress/i18n';
 import { addFilter } from '@wordpress/hooks';
-import { PanelBody } from '@wordpress/components';
+import { PanelBody, RangeControl } from '@wordpress/components';
 import { InspectorControls } from '@wordpress/block-editor';
 import { createHigherOrderComponent } from '@wordpress/compose';
 
@@ -17,6 +17,7 @@ import { createHigherOrderComponent } from '@wordpress/compose';
  * Internal dependencies.
  */
 import OptionSelectorControl from '../../components/option-selector-control';
+import PaddingWrapper from './padding-wrapper';
 const blocksWithPaddingSupport = [ 'core/group' ];
 
 const paddingOptions = [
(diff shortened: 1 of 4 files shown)


src/extensions/padding-controls/index.js
```diff
@@ -60,13 +61,23 @@ const paddingOptions = [
 const withPaddingControls = createHigherOrderComponent( ( BlockEdit ) => {
 	return ( props ) => {
 		const { name, attributes, setAttributes, isSelected } = props;
-		const { padding } = attributes;
+		const { padding, paddingCustom } = attributes;
+
+		const getMaxValue = () => {
+			const { maxValue = null } = props;
```
jrtashjian, line 67: The default `maxValue` should be passed through as a parameter instead of destructuring the props within here. The function should only have knowledge of the data passed into it.

jrtashjian (changes requested):
Looks good, only one minor change and it can be merged.

src/extensions/padding-controls/index.js
AnthonyLedesma: Good call JR. This was based on work done for the Gutter Controls component. That works differently however since its a class component as opposed to the functional component we am using for Padding Controls.
https://github.com/godaddy-wordpress/coblocks/blob/b3fae2ea8cf4fc4779d4c605bafbf877d815ec8e/src/components/gutter-control/gutter-control.js#L35-L43

This has been simplified and pushed back up.

src/extensions/padding-controls/index.js
```diff
@@ -63,15 +63,7 @@ const withPaddingControls = createHigherOrderComponent( ( BlockEdit ) => {
...
-		const getMaxValue = () => {
-			const { maxValue = null } = props;
-
-			if ( ! maxValue ) {
-				return attributes.align === 'full' ? 10 : 5;
-			}
-
-			return maxValue;
-		};
+		const maxValue = attributes.align === 'full' ? 10 : 5;
```
jrtashjian, line 66: It looks like you're removing the possibility to pass a maxValue through props entirely. I would have instead changed it like this:
```
// From this:
const getMaxValue = () => {
	const { maxValue = null } = props;

// To this
const getMaxValue = ( maxValue = null ) => {
```

and then just pass the prop into the function where it's called:

```
max={ getMaxValue( props.maxValue ) }
```

If we don't need it that's fine, I just wanted to call it out.

AnthonyLedesma: Good call. You are right that the ability to use the maxValue as a prop has been removed. I made this decision due to the use-case for this extension.

This was originally based on work done for the GutterControl component 
https://github.com/godaddy-wordpress/coblocks/blob/7d4333c2f9c1d7691cf8d9d1f8df144d019db5d8/src/components/gutter-control/gutter-control.js#L35-L43

In that instance, a getter is used to set a class property. Given the re-usability of the component, it makes more sense to allow an override to this dynamic property.

jrtashjian: I see now, thanks!

jrtashjian (approved):

Issue #1773, opened 2020-12-09
Carousel block - first gallery image not showing in Gutenberg


bradley2083:
Hey, guys.  Seeing this issue in Chrome. Happens with default WP and no plugins other than Coblocks.  [Quick video showing](https://soapbox.wistia.com/videos/eZIy12Q5pQ) the first image in the carousel is missing in Gutenberg.  It appears if you click the arrow and also in the front-end, but not on the initial view when editing the page in Gutenberg. 

Thanks!
Brad

PR #1798, opened 2021-01-19
Fix initial image hidden on carousel initialization

EvanHerman:
Resolves https://github.com/godaddy-wordpress/coblocks/issues/1773

### Description
The first carousel image is hidden on initial load. Related:

It looks like the final image is being placed at the end of the carousel, and not in the proper location before the first. When a new image is added, or the developer console is brought up, the carousel reinitializes and the issue fixes itself.

Without re-initializing the carousel again, the easiest fix is to set the `initialIndex` in `flickityOptions` resolves the issue. However, this means that the second selected image will now be the first image shown in the carousel on initialization.

#### Before
image

#### After
image

### Types of changes
Non-breaking change

### How has this been tested?
Manually checked.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/gallery-carousel/edit.js
+++ b/src/blocks/gallery-carousel/edit.js
@@ -212,6 +212,7 @@ class GalleryCarouselEdit extends Component {
 		);
 
 		const flickityOptions = {
+			initialIndex: 1,
 			draggable: false,
 			pageDots: true,
 			prevNextButtons: true,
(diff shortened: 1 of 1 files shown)


AnthonyLedesma (approved):
I found that the `imagesLoaded` script which may work alongside the Flickity Carousel, in this case, was causing superfluous loads. 
The initial image missing issue is solved by preventing `imagesLoaded` script. This also preserves the image index number.

PR #1799, opened 2021-01-20
Disable animations when IntersectionObserver is not detected

### LINKED ISSUE #1737: IntersectionObserver reference error in older browsers
### Describe the bug:

Coblocks animation [uses the IntersectionObserver object](https://github.com/godaddy-wordpress/coblocks/blob/d94c88d3f829e389f386cf757be24c3e9db0f125/src/js/coblocks-animation.js#L3) which is [not supported by older browsers like IE11 or iOS Safari prior to iOS 12.1](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#Browser_compatibility). This leads to an error when executing coblocks scripts. These are older browsers, but it seems like it might be easy to avoid this error even if animation is unsupported in older browsers.

As a bonus, if a site is using a script concat plugin, the error can prevent scripts concatenated _after_ coblocks scripts from running since the error interrupts execution.

### To reproduce:

1. Install and activate the coblocks plugin
2. Open iOS Simulator with iOS 12 and iPhone X
3. Open Safari and load your test site's homepage
4. Connect desktop safari to iOS simulator for debugging
5. Observe there is an error like "ReferenceError: Can't find variable: IntersectionObserver" related to coblocks-animation.js

### Expected behavior:

There is no script error from using a nonexistent `IntersectionObserver` global on older browsers.

### Screenshots:
Screen Shot 2020-10-20 at 4 15 53 PM

### WordPress version:
WordPress 5.5

### Gutenberg version:
9.1.1

jrtashjian:
### Description
Resolves #1737 where animations failed if IntersectionObserver was not available in older browsers.

### Types of changes
Bug fix

### How has this been tested?
Manually with iOS Simulator based on reproduction steps provided in the bug report.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/js/coblocks-animation.js
+++ b/src/js/coblocks-animation.js
@@ -1,18 +1,26 @@
 const elems = document.querySelectorAll( '.coblocks-animate' );
 
-const observer = new IntersectionObserver( ( entries ) => {
-	entries.forEach( ( entry ) => {
-		if ( ! entry.isIntersecting ) {
-			return;
-		}
+if ( 'IntersectionObserver' in window ) {
+	const observer = new IntersectionObserver( ( entries ) => {
+		entries.forEach( ( entry ) => {
+			if ( ! entry.isIntersecting ) {
+				return;
+			}
 
-		entry.target.classList.add( entry.target.dataset.coblocksAnimation );
-		observer.unobserve( entry.target );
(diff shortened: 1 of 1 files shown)


jonathanbardo (commented):
Tested and working 👍

PR #1807, opened 2021-01-27
Add Link to images within Service Item blocks

jasonlemay:
### Description
Add Link to images within Service Item blocks

### Types of changes

### How has this been tested?
Manually tested

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/services/service/edit.js
+++ b/src/blocks/services/service/edit.js
@@ -142,7 +142,7 @@ class Edit extends Component {
 			allowedTypes: [ 'image' ],
 			filesList: files,
 			onFileChange: ( [ media ] ) =>
-				this.props.setAttributes( { imageUrl: media.url, imageAttachmentPage: media.link, imageAlt: media.alt } ),
+				this.props.setAttributes( { imageUrl: media.url, imageAlt: media.alt, imageId: media.id } ),
 		} );
 	}
 
@@ -168,7 +168,7 @@ class Edit extends Component {
 						<ButtonGroup className="block-library-gallery-item__inline-menu is-right is-visible">
 							<Button
 								icon={ closeSmall }
-								onClick={ () => setAttributes( { imageUrl: '', imageAttachmentPage: '' } ) }
+								onClick={ () => setAttributes( { imageUrl: '', imageAlt: '', imageId: null } ) }
 								className="coblocks-gallery-item__button"
 								label={ __( 'Remove image', 'coblocks' ) }
 								disabled={ ! isSelected }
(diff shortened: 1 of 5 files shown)


src/blocks/services/service/edit.js
```diff
@@ -217,19 +235,33 @@ class Edit extends Component {
...
 		return (
 			<Fragment>
+				<BlockControls>
+					{ imageUrl && (
+						<ImageURLInputUI
+							url={ href || '' }
+							onChangeUrl={ this.onSetHref }
+							linkDestination={ linkDestination }
+							mediaUrl={ imageUrl }
+							mediaLink={ imageUrl }
```
jasonlemay, line 251: @jrtashjian I need to pass the right value here so we can link the image to the actual page.

This is what's done on the core/image block : 

```
mediaUrl={ ( image && image.source_url ) || url }
mediaLink={ image && image.link }
```

AnthonyLedesma:
Two more bugs around the Services block and default WordPress themes. 
https://github.com/godaddy-wordpress/coblocks/issues/1812
https://github.com/godaddy-wordpress/coblocks/issues/1813

AnthonyLedesma:
A few items that should be considered before moving forward here. Existing setup blocks do not have any reference nor any method to obtain the `imageAttachmentPage` uri. Core utilizes a differing method where the image ID is used to perform a lookup with `getMedia`. With our current architecture, I do not see how we can reliably obtain the ID or do a lookup for the correct URI. 

That being said, the ability to link to the `imageAttachmentPage` does not exist for blocks that have previously been configured with an image prior to these changes.

Similarly, the `imageAttachmentPage` attribute is not being correctly parsed from post markup using the current definition when a page is refreshed after initial configuration.

@jrtashjian Would like to request some eyes on this enhancement. I am under the impression we should limit the ability to link to the `imageAttachmentPage` since it cannot be obtained consistently. What are your thoughts?

jasonlemay:
@AnthonyLedesma @jrtashjian What should we do with this task? Is it good to go?

jrtashjian:
> Existing setup blocks do not have any reference nor any method to obtain the imageAttachmentPage uri.

Those images were "inserted by url" because of the way we originally built it. WordPress only cares if it sees an associated ID and so should we. I don't think adding a condition which checks the adjacent service blocks is worth the effort.

> I am under the impression we should limit the ability to link to the imageAttachmentPage since it cannot be obtained consistently.

You would run into a problem where new inner blocks can't link to the attachment page

1. New inner `coblocks/service` blocks can't link to the attachment page.
2. Old inner `coblocks/service` blocks can't link to the attachment page.
3. Some parent `coblocks/services` blocks allow linking to attachment pages and others don't.

The current behavior is acceptable where the individual `coblocks/service` blocks handle it's abilities based on if it has an image id or not irregardless of surrounding blocks.

jrtashjian (approved):

PR #1833, opened 2021-02-19
Fix outstanding E2E test failures

AnthonyLedesma:
### Description

This PR refactors use and implementation with the `win` object and its properties. New logic checks that the object is safe to use or throws an error when it's missing properties. 

These changes should improve the reliability and reduce the flakiness of the tests.

### Types of changes

JavaScript

### How has this been tested?

Tested manually and in the CI Pipeline

### Checklist:
- [x] My code is tested


jrtashjian (dismissed):
Although the code is fine I think the approach of just "throw an error" isn't the solution we're looking for. I'm not seeing any core differences to the original method outside of the thrown errors. We will still run into a race condition at times where the global variables we are relying on are not defined and the tests will fail.

We need to find a solution that waits and confirms that certain globals are ready before proceeding with the test to avoid race conditions. I did see [this in the Cypress docs](https://docs.cypress.io/api/commands/window.html#Start-tests-when-app-is-ready) and I wonder if this is a simple solution as opposed to our `.should( 'have.property', 'wp' )`.

If we can effectively wait for the required globals before running our tests that would bring the stability we're looking for.

AnthonyLedesma:
Additional contextual messaging from this PR has been removed. Changes are around tests remain. This PR is still needed for its alternate changes.

jrtashjian (approved):

PR #1835, opened 2021-02-24
Fix Gallery Carousel navigation with multiple blocks

### LINKED ISSUE #1549: ISBAT create two thumbnail carousels on the same page
### Describe the bug:
I should be able to create a carousel. I’ve got an issue with a carousel block which is a part of your plugin. I’m using the latest WordPress version, Twenty Twenty WP theme, and Coblocks plugin (no custom themes or any other plugins.

### To reproduce:
Here is a problem – when I add 2 carousels with the thumbnails, a weird problem appears – clicking the thumbnails (to scroll through the gallery) of the second carousel affects the first carousel instead.

### Expected behavior:
Clicking the thumbnails of the second carousel should be applied to the second carousel only. 

### Screenshots:
You may check this video record to see the problem – https://www.screencast.com/t/JjYAYNNZ2

### Isolating the problem:

- [x ] This bug happens with no other plugins activated
- [x ] This bug happens with a default WordPress theme active
- [ x] This bug happens **without** the Gutenberg plugin active
- [x ] I can reproduce this bug consistently using the steps above

### WordPress version:

the latest WP version - 5.4.2

AnthonyLedesma:
### Description

Closes #1549 
The Gallery Carousel thumbnail navigation had been configured to work with a unified class name. This caused conflicts when users added multiple Carousel galleries that have thumbnail navigation enabled. This PR introduces a new attribute that will allow thumbnail navigation when multiple galleries are used.

No deprecation function needed.

### Screenshots

galleryCarouselThumbnailNav

### Types of changes

New attribute. Does not require deprecation but will require users to enter into the editor and save the new attribute before changes will take effect.

### How has this been tested?

Tested manually
Modified the Jest test to account for new behavior.

### Checklist:
- [x] My code is tested


jrtashjian (approved):
Works well!

PR #1840, opened 2021-03-02
Add a from: header to the form emails.

EvanHerman:
### Description
Add a `From:` header to the form emails. This value is set to the name field, and if one is not set then it falls back to the site title.

### Types of changes
New feature (non-breaking change which adds functionality).

### How has this been tested?
Manually tested locally running MailHog. I was having some issues locally, so I will need to double check this is working as intended on a live site.

### Checklist:
- [x] My code is tested


AnthonyLedesma:
@EvanHerman Is there any reason not to use the mandatory `Email` field as the From header when the `Name` field is undefined? I am concerned that Site Titles are not unique enough to consistently prevent spam flagging by hosting email relay servers.

EvanHerman:
@AnthonyLedesma No, no reason other than that was the suggestion and AC. Although, I do see your point. Maybe we use the name field, and if that is not set we fall back to the email field?

EvanHerman:
@AnthonyLedesma Done in https://github.com/godaddy-wordpress/coblocks/pull/1840/commits/3bfd6405256a87956d56590c357823ab4885f13a

richtabor (approved):

PR #1845, opened 2021-03-04
Introduce WordPress 5.7 changes

AnthonyLedesma:
### Description

Changes here have been tested on both 5.7 as well as 5.6.2 versions with full E2E test suites having passed.

WordPress 5.7 Introduced API changes resulting in a necessary refactor for the CoBlocks Settings extension where old methodologies have been deprecated.
https://developer.wordpress.org/block-editor/packages/packages-data/#register

Backward compatible test selector updates.
Improved cross browser support by `focus()` on input instead of `click()` prior to `type()` event.
Cypress update to 6.6.0.

New method to force CI to run all tests. The requirement will be to use a branch name that includes the string **`run-all-tests`**. This new feature has been included with a `.md` document file.

### Types of changes

Test updates, Refactor to CoBlocks Settings, and new test override feature.

### How has this been tested?

Tested manually and in CI. WP 5.7 + WP 5.6.2
### Checklist:
- [x] My code is tested


AnthonyLedesma:
Tests Passing WP 5.7-RC2 [695c8c9](https://github.com/godaddy-wordpress/coblocks/pull/1845/commits/695c8c995fe4fe6832b71010bbef6d31950100a8)

Tests Passing WP 5.6.2 [a40f114](https://github.com/godaddy-wordpress/coblocks/pull/1845/commits/a40f1149dee4410fe7d332b3422c8c656f0a2252)

src/extensions/coblocks-settings/coblocks-settings-store.js
```diff
@@ -1,17 +1,75 @@
...
+ *
+ * @function getStoreType
+ * @param {string} storeName - Store identifier for CoBlocks Setting store.
+ * @param {Object} storeOptions - Object with Redux-like configuration.
+ * @return {Function} createReduxStore(storeName, storeOptions) || registerStore(storeName, storeOptions)
+ */
+const getStoreType = ( storeName, storeOptions ) => {
+	return registerStoreDeprecated
+		? createReduxStore( storeName, storeOptions ) : registerStore( storeName, storeOptions );
+};
```
jrtashjian, line 32: This comment applies to the entire file. Honestly nothing needs to be changed here for the 5.7 release since the functions are deprecated not removed. Instead we should look at refactoring this store in the future while also bumping the minimum WP version to 5.7 (because of the new APIs).

I think the best thing we could do here is remove this store entirely and instead use the core data api for interacting with the settings endpoint. This only requires we register our settings with the REST API on the PHP side.

jrtashjian (changes requested):
The test updates are good but I think the entire store could be left alone for now and in fact could be removed entirely in the future by preferring the existing core-data apis.

src/extensions/coblocks-settings/coblocks-settings-store.js
AnthonyLedesma: I'd like a bit more information here on what you are suggesting. Are you referring to making API PUT requests to handle settings changes and exclude the use of a store altogether? I would agree with that solution if so.

One of the reasons that I had to change the store completely was that WordPress 5.7 had not just deprecated the APIs but had also made enough sweeping changes to break backward compatibility for the deprecated APIs. The store, as we were using it, was no longer functional.

AnthonyLedesma: Here is the behavior of CoBlocks settings store prior to refactors: https://dashboard.cypress.io/projects/sovnn2/runs/1172/test-results/d3772cfb-af9f-4d09-b0bf-c088fdf8d6b7

Essentially UI would not reflect actually saved settings.

EvanHerman (approved):
This looks good to me. Tests are passing on my en, locally. 👍 Not sure if @jrtashjian also wants to take a look at this, but I'm good with merging this.

src/extensions/coblocks-settings/coblocks-settings-store.js
jrtashjian: Logically the old function still exists and works with no changes from my tests and that test failure seems to be caused by the selector not existing. I'm not seeing that selector actually being rendered although it does exist in our code. Are we using the component properly?

Regardless, we can still merge this I'm just noting this is a deprecation of the core API not a breaking change and that we should remove this store in the future.

jrtashjian (approved):

PR #1850, opened 2021-03-12
Added lot of Google Fonts

mhimon:
Hi,
I am also a Gutenberg Block Developer. I had a few fonts like coblocks so I was thinking to extend them! I have tested these fonts with my block plugin 

[Powerful Blocks for Gutenberg](https://wordpress.org/plugins/powerful-blocks) 
I have tested on localhost, didn't push the updates yet!

https://profiles.wordpress.org/mhimon/


mhimon:
Hi,
I think we shouldn't use localize script. But yeah it is good to use unminified objects. You can unminify or tell me, I will unminify and create another PR.

Thanks

jrtashjian:
No need to make another PR since you can make the adjustments here.

What are the reasons you feel localize script is not a good solution and what would be your suggestion instead? I suggest we use `wp_localize_script` because it allows developers to filter the list of available fonts before they are passed into the React app which in the WordPress ecosystem is much more accessible and allows us to optimize if necessary before the site is transmitted to a device.

mhimon:
Hi,
I added almost 1k+ fonts! Its huge so I think if we add in js as js files
will be minified later it will be better.

jrtashjian:
Have you looked into the performance impact of adding 1k+ fonts when editing a page in the block editor? How does the UI where these fonts are loaded handle it? What is the increase in resources loaded on the page like now?

src/components/font-family/fonts.js
```diff
@@ -3,34 +3,1011 @@
...
-	'Playfair Display': { weight: [ '400', '700', '900' ] },
-	'PT Serif': { weight: [ '400', '700' ] },
-	Roboto: { weight: [ '100', '300', '400', '500', '700', '900' ] },
-	Rubik: { weight: [ '300', '400', '500', '700', '900' ] },
-	Tajawal: { weight: [ '200', '300', '400', '500', '700', '800', '900' ] },
-	Ubuntu: { weight: [ '300', '400', '500', '700' ] },
-	Yrsa: { weight: [ '300', '400', '500', '600', '700' ] },
+    "ABeeZee": { "weight": ["400"] },
+    "Abel": { "weight": ["400"] },
+    "Abhaya Libre": { "weight": { "0": "400", "3": "500", "4": "600", "5": "700", "6": "800" } },
```
jasonlemay, line 8: Formating is wrong. Weights should be in array.

`'Abhaya Libre': { weight: [ '400', '500', '600', '700', '800' ] }`

Otherwise we get this :

image

image

jasonlemay: Also, `yarn lint:js` is throwing 31551 errors

mhimon:
Check now

Issue #1768, opened 2020-12-01
'Add accordion item' creates 2 accordion items instead of one


davidLpt:
Hi, when you click in the 'Add content' field of an accordion item and then click on 'Add accordion Item' button right after, it creates 2 new accordion item blocks instead of one.

richtabor:
I was able to repeat this behavior, but it was only happening if I had more than one accordion item added. Here's a screen recording: 
![ScreenFlow](https://user-images.githubusercontent.com/1813435/103812922-40739880-502d-11eb-82bb-073d9c28fa00.gif)

olafleur:
@AnthonyLedesma I think that issue can be closed. :)

PR #1851, opened 2021-03-15
Fix for the fact that sometimes too much accordion items were added

olafleur:
### Description
This is a proposed fix for #1768 

### Types of changes
Bug fix

### How has this been tested?
The automated test was only triggering the onMouseUp event. When you change it to a full click event in the test, it reproduce the error in the issue.
I have changed the onMouseUp event in the component to on MouseDown and it fixes the test and the behavior, from what I see.

### Checklist:
- [x] My code is tested


jonathanbardo (approved):
@olafleur 👏 Thank you! I tested your CyPress fix and it does indeed highlight the bug quite nicely.

AnthonyLedesma (approved):
Works well. LGTM!

PR #1852, opened 2021-03-15
Fix Logos & Badges block sizing behavior

### LINKED ISSUE #1811: ISBAT use consistently rendered Logos & Badges block
### Describe the bug:

When first adding the Logos & Badges block we are presented with a different render from what is expected on the Front-end as well as what can be expected within the editor on page refresh. 

### To reproduce:

1. Add Logos & Badges block and add several images to the block.
2. Take note of the visual appearance of the Logos & Badges.
3. Publish/save the page and refresh the editor page.
4. Take note of the visual changes of the Logos & Badges block.

### Expected behavior:

Expect that initial width attributes either persist upon page save.
OR
Expect more accurate initial attributes.

### Screenshots:

logosDifferentOnRefresh

### Isolating the problem:

### WordPress version:

5.6

### Gutenberg version:

N/A

AnthonyLedesma:
### Description

When the Logos & Badges block is used and the image width is not configured by the user, the existing block logic will automatically determine the widths for the images. When the save function fires, there was an improper logic that did not properly determine the widths of images based on row count. Fixes here allow for proper render of Logos & Badges after saving and editor reload.

This change does **not** require a deprecated save function. The issue was around where attributes had been improperly parsed upon _initial_ save for unconfigured images. This condition cannot be met again after the initial save, thus no block validation errors will occur.

Closes #1811 

### Screenshots

logosBadgesSizingFixed

### Types of changes

Minor save function change.

### How has this been tested?

Tested manually.

### Checklist:
- [ ] My code is tested

### DIFF
--- a/src/blocks/logos/save.js
+++ b/src/blocks/logos/save.js
@@ -4,9 +4,9 @@
 import { chunk } from 'lodash';
 
 export default function save( { attributes, className } ) {
-	const { images, align } = attributes;
+	const { align } = attributes;
 
-	const hasImages = !! images.length;
+	const hasImages = !! attributes.images.length;
 
 	if ( ! hasImages ) {
 		return null;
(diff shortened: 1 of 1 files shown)


jrtashjian (approved):
Looks good!

Issue #1842, opened 2021-03-03
ISBAT Gists are not displayed in editor


mvlaisavljevic:
### Describe the bug:
There's an issue with the Gist block. If we have more than one Gist block on a post/page, only one of them can be previewed once we come back to edit the post/page (the preview works fine the first time the Gists are added).

### To reproduce:

1. Create a couple of Gist blocks
2. Save the page
3. Go to the page editor again
4. See error > https://prnt.sc/10c5s0p

### Expected behavior:
All Gist blocks to be visible in the editor

### Isolating the problem:

- [x] This bug happens with no other plugins activated
- [x] This bug happens with a default WordPress theme active
- [x] This bug happens **without** the Gutenberg plugin active
- [x] I can reproduce this bug consistently using the steps above

### WordPress version:
tested on the latest WP 5.6.2

PR #1856, opened 2021-03-16
When two or more gists on an edit page, fix so that they render properly

olafleur:
### Description
This is a proposed fix for #1842 

Each time we request a new Gist, we have to provide a new global function name to serve as the JSONP callback.
The static callback id to remember wasn't properly incremented.

### Types of changes
Bug fix

### How has this been tested?
I have written a test to reproduce the bug and then fixed it.

### Checklist:
- [x] My code is tested


AnthonyLedesma (approved):
Pulled it down and tested it. Great work. LGTM.

Issue #1331, opened 2020-02-10
ISBAT receive consistent error notices using Gist block


AnthonyLedesma:
**Describe the bug**
When a request returns as a 404 the Gist block becomes unusable. The block should be improved by adding proper error handling using the existing error notice mechanism.

The Gist block is generating an error when applying a URL with a Gist file.
![image](https://user-images.githubusercontent.com/30462574/74161676-8edda480-4bdc-11ea-9f9c-4720aa68b5e4.png)

Original URL
`https://gist.github.com/AnthonyLedesma/33ad1a8cd86da3b6bddbdefa432cb51d#file-gistblocktest-text` (Returns 200)

Transformed request
`https://gist.github.com/AnthonyLedesma/33ad1a8cd86da3b6bddbdefa432cb51d.json?callback=embed_gist_callback_NaN&file=gistblocktest.text` (Returns 404)

**To Reproduce**

1. Copy a Gist URL with a file.
  * EG( https://gist.github.com/AnthonyLedesma/33ad1a8cd86da3b6bddbdefa432cb51d#file-gistblocktest-text )
2. Add Gist block to a page.
3. Paste the Gist URL with a file into the Gist block.
4. See error.

**Screenshots**

**Gist URL without file**
![gistBlockUrlWithoutFile](https://user-images.githubusercontent.com/30462574/74162085-512d4b80-4bdd-11ea-831f-96c3b5940175.gif)

**Gist URL with file and error returned from Github**
![gistBlockError](https://user-images.githubusercontent.com/30462574/74161858-eda31e00-4bdc-11ea-94dc-8aaf79920084.gif)

**Expected behavior**
Expect Gist block to properly load Gist files when a correct URL is provided. Expect Gist block to provide useful error notices and recover from invalid attributes.

**Isolating the problem:**

- This bug happens with no other plugins activated
- This bug happens with a default WordPress theme active
- This bug happens **without** the Gutenberg plugin active
- I can reproduce this bug consistently using the steps above

**WordPress Version**
What version of WordPress are you using?
5.3.2

**Gutenberg Version**
What version of the Gutenberg plugin are you using? (If any)
N/A (Also happens with 7.4) active.

**Side note**
I found it interesting that raw transforms properly handles both cases of URL **with** and **without** a file present. See Gif below.
![gistBlockRaw](https://user-images.githubusercontent.com/30462574/74162378-e4668100-4bdd-11ea-9382-d315bf08edca.gif)

[...]

PR #1859, opened 2021-03-20
Refactor Gist component to hooks

olafleur:
### Description
This is a suggestion for a refactor of the Gist component to hooks. I think that it will allow to make a cleaner fix for #1331 after that.

Open to comments/improvements.

### Types of changes
Refactoring

### How has this been tested?
All the tests still pass.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/gist/edit.js
+++ b/src/blocks/gist/edit.js
@@ -37,6 +37,14 @@ class Edit extends Component {
 		}
 	}
 
+	static __gistCallbackId = 0;
+
+	// Each time we request a new Gist, we have to provide a new
+	// global function name to serve as the JSONP callback.
+	static __nextGist() {
+		return 'embed_gist_callback_' + this.__gistCallbackId++;
+	}
+
 	updateURL( newURL ) {
 		this.props.setAttributes( { url: newURL, file: '' } );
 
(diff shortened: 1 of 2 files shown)


src/blocks/gist/edit.js
```diff
@@ -37,6 +37,14 @@ class Edit extends Component {
 		}
 	}
 
+	static __gistCallbackId = 0;
```
jonathanbardo, line 40: @olafleur Have you noticed this being used anywhere else? Wondering if it has any purpose of being static or not.

olafleur: Yes, it is in fact useful to make sure every JSONP callback as a unique id (see #1856).

However, it could easily be rewritten as a useState hook if the Edit component was also rewritten with hooks.

I will put back the comment that was explaining it before ;)

jrtashjian: FWIW, this can be rewritten to use hooks if you'd like. We have been slowly transitioning parts of code to utilize hooks.

olafleur: Yes! I was trying with this PR to see if it was perceived as a good idea. If this get merged I'll gladly convert also the Edit component to Hooks :)

snovosel (approved):

AnthonyLedesma (approved):
Great work here @olafleur. LGTM!

PR #1869, opened 2021-03-26
Refactor the settings modal to simplify and isolate each control's implementation

jrtashjian:
### Description
The settings modal has been refactored to remove the `coblocks-settings-modal` store that is registered and implements a new slot-fill which allows us to keep controls with their associated files.

I've abstracted out a new component called `CoBlocksSettingsToggleControl` that we can use anywhere in our code to provide a new toggle control that is inserted into this settings modal. Custom controls can be added by simply using the `CoBlocksSettingsModalControl` slot-fill. I've added this example within an inline comment:

```
import { registerPlugin } from '@wordpress/plugins';
import { CoBlocksSettingsModalControl } from './coblocks-settings-slot';
import { CoBlocksSettingsModalControl } from './coblocks-settings-toggle-control';

registerPlugin( 'your-settings-modal-control', {
    render: () => (
        <CoBlocksSettingsModalControl>
            <CoBlocksSettingsToggleControl
                settingsKey="name_of_wordpress_setting"
                label={ __( 'The toggle control label', 'coblocks' ) }
                help={ __( 'The toggle control help text', 'coblocks' ) }
            />
        </CoBlocksSettingsModalControl>
    ),
} );
```

All existing controls have been migrated out of this settings-modal file and are now registered from their own extension directories.

### Screenshots

Slightly modified styles because of updates in core. I've also remove the horizontal rule separating each control.

Before:

Screen Shot 2021-03-26 at 17 05 43

After:

Screen Shot 2021-03-26 at 21 16 54

### Types of changes
Bug fix

### How has this been tested?
E2E tests

### Checklist:
- [x] My code is tested


snovosel (approved):

src/extensions/block-editor-settings/index.js
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Internal Dependencies
+ */
+import CoBlocksSettingsModalControl from '../coblocks-settings/coblocks-settings-slot';
+
+export const COLOR_PANEL_SETTINGS_KEY = 'coblocks_color_panel_controls_enabled';
+export const CUSTOM_COLORS_SETTINGS_KEY = 'coblocks_custom_colors_controls_enabled';
+export const GRADIENT_PRESETS_SETTINGS_KEY = 'coblocks_gradient_presets_enabled';
+
+function CoBlocksEditorSettingsControls() {
```
jonathanbardo, line 20: @jrtashjian I wonder if the color related settings belong here at all? Maybe seperate then into their own concern?

src/extensions/coblocks-settings/index.js
```diff
@@ -9,52 +9,85 @@ import './styles/style.scss';
...
+					closeLabel={ __( 'Close', 'coblocks' ) }
+					onRequestClose={ closeModal }
+					shouldCloseOnClickOutside={ false }
+				>
+					<section className="edit-post-preferences-modal__section">
+						<h2 className="edit-post-preferences-modal__section-title">
+							{ __( 'General' ) }
+						</h2>
+
+						<CoBlocksSettingsModalControl.Slot>
```
jonathanbardo, line 55: ❤️ this

src/extensions/block-editor-settings/index.js
jrtashjian: You're right, I didn't think enough on where to put this one. I can move it into `src/extensions/colors` instead.

jonathanbardo:
@jrtashjian do we still need this since we got rid of the store? https://github.com/godaddy-wordpress/coblocks/blob/master/includes/class-coblocks-settings.php#L62

jrtashjian:
We can definitely get rid of that nonce.

src/extensions/block-editor-settings/index.js
jonathanbardo: Yeah I feel it might be more appropriate

AnthonyLedesma (commented):
There seems to be a condition preventing the implementation of one of these settings within the CI environment. Notice there are 5 settings registered in CI and 6 settings registered locally.

**CI Env**
image

**Locally**
image

jrtashjian:
@AnthonyLedesma good catch! It's actually missing the Layout Selector toggle because the layout selector isn't being setup for this test. I'll get that setup.

AnthonyLedesma (approved):
LGTM

PR #1871, opened 2021-03-30
Add rtl prop to slick-slider component based on if locale is set to rtl

### LINKED ISSUE #1704: ISBAT use Post Carousel block in RTL language
### Describe the bug:
[This PR#1700](https://github.com/godaddy-wordpress/coblocks/pull/1700) works to address the CSS changes that are needed when using an RTL language. Regardless of the CSS changes that are needed there are also changes needed to correct an improper Slick carousel initialization.

### To reproduce:

1. Add Post Carousel to page and publish.
2. Switch site to RTL language like Arabic.
3. View Post Carousel on front-end and editor to see the issues.

### Expected behavior:

Expect the Post Carousel block to be usable on front-end when using an RTL language.

### Screenshots:

**When LTR**
postCarouselBug

**When RTL**
postCarouselBugRTL

### Isolating the problem:

### WordPress version:

5.5.1

### Gutenberg version:

N/A

snovosel:
### Description
Determine if locale is set to rtl and then set the rtl setting on the react slick slider accordingly.
fixes #1704

### Types of changes
Bug fix

### How has this been tested?
visual inspected

### Checklist:
- [x] My code is tested


AnthonyLedesma (changes requested):
Within the editor and using an RTL mode we are still seeing LTR behavior and setting.
image

From my testing, it looks like the front-end is not properly setting as RTL either. We see the same indexing scrolling behavior on the front-end as we do in the editor.

According to the docs here: 
https://kenwheeler.github.io/slick/
It might be related to this:
```Note: the HTML tag or the parent of the slider must have the attribute "dir" set to "rtl".```

AnthonyLedesma:
The Jest tests related to this change do not seem to accurately reflect the markup. For example, when we pass in the RTL prop as true, the markup reflects RTL as false. We should either fix the test if possible and reasonable or remove the test.

AnthonyLedesma (approved):
LGTM

PR #1874, opened 2021-04-01
Fix layouts jumping around within the Layout Selector in Safari 

jasonlemay:
### Description
Fix layouts jumping around within the Layout Selector in Safari. I had to remove the spinner in Safari, there's a known bug with `column-count` that we can't fix.

### Types of changes

### How has this been tested?
Manually tested

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/extensions/layout-selector/editor.scss
+++ b/src/extensions/layout-selector/editor.scss
@@ -53,6 +53,7 @@
 		height: 96px;
 		margin: 0;
 		padding: 0;
+		z-index: 15;
 
 		@include break-medium() {
 			background-color: $white;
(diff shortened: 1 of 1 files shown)


AnthonyLedesma (changes requested):
I am seeing some jumping going on in Chrome and FF as well. 

layoutSelectorJumping

jasonlemay:
@AnthonyLedesma It looks like the layouts are loaded twice. But this is not the same bug (see movie recording in WPEX-1182). Do you have this same bug on `master` too? If so, we should create a new bug story for this.

AnthonyLedesma (approved):
Good call @jasonlemay! Definitely unrelated.

PR #1875, opened 2021-04-01
Fix posts block margin in Safari and IE

EvanHerman:
### Description
In Safari and IE, the posts block content height was incorrect. This PR fixes that issue.

#### Chrome
image

#### Safari
image

#### Safari Editor
image

### Types of changes

### How has this been tested?
Manually tested this in Safari. Confirmed no regressions were introduced in other browsers.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/posts/styles/style.scss
+++ b/src/blocks/posts/styles/style.scss
@@ -20,7 +20,6 @@
 
 .wp-block-coblocks-posts__item {
 	align-items: stretch;
-	display: flex;
 	flex: 1 1 auto;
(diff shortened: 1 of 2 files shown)


AnthonyLedesma (changes requested):
I think we need a way to handle this for the Horizontal style as well as the stacked style.

postsStackedHorizontal

EvanHerman:
@AnthonyLedesma Good call. I forgot to double check the horizontal layouts. I've made some tweaks to account for that.

AnthonyLedesma (approved):
LGTM

PR #1876, opened 2021-04-01
Fix width of images in image filter previews

EvanHerman:
### Description
Fix the width of images in the image filter preview window. The preview is still about 3px taller than the image, so there is a small "tail" of background color beneath the image. The height of the parent container, holding the image, is auto generated, so I'm not sure the best way to set that.

### Types of changes

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/extensions/animation/styles/editor.scss
+++ b/src/extensions/animation/styles/editor.scss
@@ -96,6 +96,10 @@
 		max-width: none !important;
 	}
 
+	.components-resizable-box__container {
+		width: 100% !important;
+	}
+
 	.components-tip {
 		margin-top: 12px;
 	}


AnthonyLedesma (commented):
I am not seeing any visual difference with previews when viewing on Master vs on fix/animation-preview-width branch.

EvanHerman:
@AnthonyLedesma I named this branch incorrectly. All image previews should be 100% width. Not just the animation preview. Also, I added an image up there with whats currently happening on master. The image doesn't extend the full width of the preview window.

AnthonyLedesma (approved):
LGTM!

PR #1878, opened 2021-04-06
Try shaking the trees

AnthonyLedesma:
### Description

Need to consider side effects. All tests run here in an attempt to spot issues.

### Checklist:
- [ ] My code is tested


AnthonyLedesma:
Size dropped from 2.4 -> 2.3.

AnthonyLedesma:
I don't think that this will benefit us much for the work it takes to maintain and understand. We should consider delivering our CSS and JS differently in terms of performance optimizations.

PR #1891, opened 2021-04-21
add default align property to gallery collage block

snovosel:
### Description
Sets align attribute to wide as default on gallery collage block

### Types of changes

Enhancement

### How has this been tested?
visually tested

### Checklist:
- [x] My code is tested


AnthonyLedesma (approved):
No deprecation needed. The block is automatically updated to use the new default alignment if it has not previously been configured by a user. This means that existing blocks with images but have not yet been aligned by a user will still be upgraded to the new default width upon loading the editor. 

LGTM.

AnthonyLedesma (changes requested):
Actually, I realize that this change causes default markup for the collage gallery to differ from previous iterations. We need to update Jest test snapshots here I think and possibly modify some tests.

snovosel:
@AnthonyLedesma updating snapshots seem to have fixed the test issues. Was there anything else that should to be updated?

AnthonyLedesma (approved):
LGTM

PR #1905, opened 2021-05-17
Try Carousel block variations

AnthonyLedesma:
### Description

Introduce Carousel block variation picker and update relevant tests.

### Types of changes

Major JavaScript changes.

### How has this been tested?

E2E tests have been updated.

### Checklist:
- [x] My code is tested


richtabor (approved):
Works as expected!

richtabor (changes requested):
Just one clarification: Does the "Skip" go to the first variation by default? Or can we set what that "Skip" goes to? I'd prefer if it went to the standard experience without any attributes defined - other than the default values, instead of the current flow (which goes to the thumbnails carousel).

richtabor:
Also, the media uploader at the footer of the block is missing: 
Screen Shot 2021-05-25 at 10 58 13 AM

AnthonyLedesma:
This is complicated to control where the skip will go due to the way that this component works. The way that the variation picker knows to show is by comparing variation attributes to actual attributes. In other words, default attributes === show variation picker. Indeed this is the way the Columns block behaves as well. Additionally, the logic used to show the Variation title within the placeholder does not have any method to detect when a skip has been processed, meaning that the default variation title would show instead of `Gallery Carousel`.

AnthonyLedesma:
> Also, the media uploader at the footer of the block is missing:
> Screen Shot 2021-05-25 at 10 58 13 AM

This has been fixed in the latest commit.

richtabor (approved):

PR #1907, opened 2021-05-24
Allow inner blocks (e.g. lists) in alert

GenieTim:
### Description

This PR adds the possiblity to add "child-blocks" (inner blocks) to the alert box. The main intent was to add the possibility for lists, but others are possible too. 

See also: https://wordpress.org/support/topic/add-lists-in-alert-block/#post-14476981

### Screenshots

Screenshot 2021-05-24 at 19 13 00

### Types of changes

New feature. 
The only "breaking" change that could be discussed is the reduction of the importance of .wp-block-coblocks-alert__text, as it can be added as a simple paragraph which will not have this class. 

### How has this been tested?

Compiled the plugin & installed in my WordPress installation (5.7.2). Visual confirmation that lists and other blocks could be added and successfully persisted.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/alert/edit.js
+++ b/src/blocks/alert/edit.js
@@ -11,12 +11,32 @@ import Inspector from './inspector';
 import applyWithColors from './colors';
 
 /**
+ * Allowed blocks and template constant is passed to InnerBlocks precisely as specified here.
+ * The contents of the array should never change.
+ * The array should contain the name of each block that is allowed.
+ *
+ * @constant
+ * @type {string[]}
+ */
+ const ALLOWED_BLOCKS = [
+  "core/heading",
+  "core/paragraph",
+  "core/spacer",
+  "core/button",
+  "core/buttons",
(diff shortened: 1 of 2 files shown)


AnthonyLedesma:
Hi @GenieTim, Thank you for taking the time to create this pull request adding innerBlocks functionality to the Alert block. After consideration, I do not think that this will be an enhancement that should be merged into CoBlocks. The Alert block having the intention to emphasize sections of text does not seem to align with the idea of showing content from multiple blocks. It seems that they may be better/existing ways to bring attention to lists without needing to emphasize the entire list. 

For example, I was able to create an aesthetically similar block using the Group block containing a Paragraph and a List block by setting a custom background color. I am also able to change fonts or bold text to emphasize making this enhancement less useful. I will be closing this issue but I encourage further comments or contributions.

Thank you again for your time to contribute.

PR #1909, opened 2021-05-27
Bring the farthest back image to the front when it's selected in the layered gallery block

EvanHerman:
### Description
When `image-2` is selected in the layered gallery, the image now pops to the front so the upload/replace/x buttons are all visible to the user.

### Screenshots
image

image

### Types of changes

### How has this been tested?
Manually tested.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/gallery-collage/styles/style.scss
+++ b/src/blocks/gallery-collage/styles/style.scss
@@ -142,6 +142,10 @@
 			width: calc(198 / 890 * 100%);
 			z-index: 2;
 
+			&.is-selected {
+				z-index: 4;
+			}
+
 			figure,
 			.block-editor-media-placeholder {
 				padding-top: 100%;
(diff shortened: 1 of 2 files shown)


src/blocks/gallery-collage/styles/style.scss
```diff
@@ -142,6 +142,10 @@
 			width: calc(198 / 890 * 100%);
 			z-index: 2;
 
+			&.is-selected {
```
jonathanbardo, line 145: @EvanHerman This works great. I think nth-child(4) would also benefit from this. The replace button is visible in english but it might be hidden in other languages. I tested added the same rule on the 4th element and it works pretty well overall. What do you think?

AnthonyLedesma: I agree that the nth-child(4) should be targeted.

AnthonyLedesma (approved):
LGTM. I went ahead and pushed up a change so that this new z-index is applied to the `nth-child(4)` element.

PR #1917, opened 2021-06-08
Fix image link controls for Services block

### LINKED ISSUE #1901: ISBAT add a link to an image in the Service Block
### Describe the bug:
ISBAT add a link to an image in the Service Block

### To reproduce:

1. Add New Page
2. Add Service Block
3. Add Image to the block
4. Try to add a link to it. 

### Expected behavior:
You should be able to do so. 

### Screenshots: (Video)
https://cloudup.com/cLMgjiNrsa8

As You can see from the video, it is still possible to add a link through the code editor. It just seems that there is a bug with adding the link normally from the visual editor. 

### WordPress version:
Current version: 5.7.1

### Gutenberg version:
Version 10.5.4

snovosel:
### Description
Adds the href attribute to the image within the services block
Fixes #1901 

### How has this been tested?
Visually

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/services/service/edit.js
+++ b/src/blocks/services/service/edit.js
@@ -166,7 +166,7 @@ const Edit = ( props ) => {
 		);
 	};
 
-	const onSetHref = ( ) => setAttributes( props );
+	const onSetHref = ( { href } ) => setAttributes( { href } );
 
 	const { className } = props;
 	const {


olafleur (approved):
It works great! Good job!

In term of UX improvement, I was wondering : should we automatically append a http before an URL that the user enters?

For example, if I enter example.com, the block is looking for a local path (localhost:8888/example.com) instead of https://example.com

AnthonyLedesma:
Are there any side-effects for Service blocks that had already been configured? I am thinking about validation errors.

snovosel:
@AnthonyLedesma I don't believe so. Previous to this PR this function would only have been called when adding a URL link to an image service

AnthonyLedesma (approved):
LGTM

PR #1925, opened 2021-06-23
Update dependencies

EvanHerman:
### Description
Update package dependencies.

### Types of changes
No new code, updated dependencies.

### Checklist:
- [x] My code is tested


package.json
```diff
@@ -131,6 +131,16 @@
 		"style-loader": "^0.19.1",
 		"webpack-fix-style-only-entries": "^0.4.0"
 	},
+	"resolutions": {
```
olafleur, line 134: Nice! I did not know that existed!

olafleur (approved):

package.json
EvanHerman: Sadly, I had to go this route because some nested dependencies were outdated.

jonathanbardo: @olafleur only in yarn :)

olafleur: @EvanHerman : yeah, not always an ideal situation, but it is sometimes a useful tool, like in this case :)

AnthonyLedesma:
I always prefer to execute all E2E tests for an extra level of confidence that cascading issues do not exist within our software due to these dependency updates. I've re-branched this so we can watch tests here:

https://app.circleci.com/pipelines/github/godaddy-wordpress/coblocks?branch=fix%2Fdependencies-run-all-tests

AnthonyLedesma (approved):
LGTM. Tests are good.
image

https://dashboard.cypress.io/projects/sovnn2/runs/1394/overview

PR #1928, opened 2021-06-28
ISBAT add an OpenTable Reservations block to my site

ehamwey:
### Description
Add an OpenTable reservations block to a CoBlocks-enabled site. This PR serves as a point of discussion for the changes being made.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
 		"zh_TW": "繁體中文"
 	},
 	"dependencies": {
-		"@godaddy-wordpress/coblocks-icons": "1.3.0"
+		"@godaddy-wordpress/coblocks-icons": "1.5.3"
 	},
 	"devDependencies": {
 		"@babel/plugin-proposal-class-properties": "7.10.4",
--- a/src/blocks/opentable/edit.js
+++ b/src/blocks/opentable/edit.js
@@ -1,0 +1,116 @@
+/**
+ * External dependencies
+ */
+import { OpentableIcon as icon } from '@godaddy-wordpress/coblocks-icons';
+import { useState, useEffect } from '@wordpress/element';
+/**
+ * Internal dependencies
+ */
+import Controls from './controls';
+import OpenTable from './opentable';
+import InspectorControls from './inspector';
+
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { compose } from '@wordpress/compose';
+import {
+	Placeholder,
+	Button,
+	withNotices,
+	Icon,
+	TextControl,
+} from '@wordpress/components';
+
+const Edit = ( props ) => {
+	// const onSelectImages = ( images ) => {
+	// 	props.setAttributes( { images: images.map( ( image ) => pickRelevantMediaFiles( image ) ) } );
+	// };
+
+	// const onDropImages = ( images ) => {
+	// 	const imagesNormalized = images.map( ( image ) => pickRelevantMediaFiles( image ) );
+	// 	const currentImages = props.attributes.images || [];
+	// 	props.setAttributes( { images: currentImages.concat( imagesNormalized )	} );
+	// };
+
+	// const { className, noticeOperations, attributes, noticeUI, isSelected } = props;
+
+	// const [ restaurantID, setRestaurantID ] = useState();
+	const [ ridField, setRidField ] = useState();
+
+	const { className, isSelected, attributes } = props;
+
+	useEffect( () => {
+		setRidField( attributes.restaurantID );
+	}, [] );
+
+	const renderOpenTable = ( event ) => {
+		if ( event ) {
+			event.preventDefault();
+		}
+
+		// setAttributes( { restaurantID: restaurantID } );
+	};
+
+	// const hasRestaurantID = !! attributes.restaurantID.length;
+
+	return (
+		<>
+			<Controls { ...props } />
+			<InspectorControls
+				attributes={ attributes }
+				setAttributes={ props.setAttributes }
+			/>
+			<div className={ className }>
+
+				{ ( ! attributes.restaurantID ) ? (
+					<Placeholder
+						icon={ <Icon icon={ icon } /> }
+						label={ __( 'OpenTable', 'coblocks' ) }
+						instructions={ __(
+							'Enter your OpenTable Restaurant ID to show the reservations widget.',
+							'coblocks'
+						) }
+					>
+
+						<form onSubmit={ renderOpenTable }>
+							<TextControl
+								value={ ridField }
+								className="components-placeholder__input"
+								placeholder={ __(
+									'Restaurant ID',
+									'coblocks'
+								) }
+								onChange={ ( newRestaurantID ) => {
+									setRidField( newRestaurantID );
+								} }
+							/>
+							<Button
+								isPrimary={ !! ridField }
+								isSecondary={ ! ridField }
+								type="submit"
+								disabled={ ! ridField }
+								onClick={ () => {
+									props.setAttributes( { restaurantID: ridField } );
+								} }
+							>
+								{ __( 'Embed', 'coblocks' ) }
+							</Button>
+						</form>
+						<a target="_blank" rel="noopener noreferrer" href="https://guestcenter.opentable.com/login">{ __( 'Get your Restaurant ID here.', 'coblocks' ) }</a>
+					</Placeholder>
+				)
+					: <>
+						<div
+							style={ { width: '100%', height: '100%', position: 'absolute' } }
+						/>
+						<OpenTable { ...props } />
+					</>
+				}
+			</div>
+		</>
+	);
+};
+
+export default compose( [ withNotices ] )( Edit );
--- a/src/blocks/opentable/index.js
+++ b/src/blocks/opentable/index.js
@@ -1,0 +1,76 @@
+/**
+ * External dependencies
+ */
+import { OpentableIcon as icon } from '@godaddy-wordpress/coblocks-icons';
+
+/**
+ * Internal dependencies
+ */
+import edit from './edit';
+import metadata from './block.json';
+import save from './save';
+
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { Icon } from '@wordpress/components';
+
+/**
+ * Block constants.
+ */
+const { name, category, attributes } = metadata;
+
+const settings = {
+	/* translators: block name */
+	title: __( 'OpenTable', 'coblocks' ),
+	/* translators: block description */
+	description: __( 'Embed an OpenTable Reservations Widget.', 'coblocks' ),
+	icon: <Icon icon={ icon } />,
+	keywords: [
+		'coblocks',
+		/* translators: block keyword */
+		__( 'restaurant', 'coblocks' ),
+		/* translators: block keyword */
+		__( 'reservation', 'coblocks' ),
+		/* translators: block keyword */
+		__( 'open', 'coblocks' ),
+		/* translators: block keyword */
+		__( 'table', 'coblocks' ),
+	],
+	styles: [
+		{
+			name: 'standard',
+			/* translators: block style */
+			label: __( 'Standard', 'coblocks' ),
+			isDefault: true,
+		},
+		{
+			name: 'wide',
+			/* translators: block style */
+			label: __( 'Wide', 'coblocks' ),
+		},
+		{
+			name: 'tall',
+			/* translators: block style */
+			label: __( 'Tall', 'coblocks' ),
+			isDefault: true,
+		},
+		{
+			name: 'button',
+			/* translators: block style */
+			label: __( 'Button', 'coblocks' ),
+		},
+	],
+	example: {
+		attributes: {
+			align: 'wide',
+			restaurantID: '123456',
+		},
+	},
+	attributes,
+	edit,
+	save,
+};
+
+export { name, category, metadata, settings };
--- a/src/blocks/opentable/index.js
+++ b/src/blocks/opentable/index.js
@@ -62,6 +62,9 @@ const settings = {
 			label: __( 'Button', 'coblocks' ),
 		},
 	],
+	supports: {
+		align: [ 'wide', 'full' ],
+	},
 	example: {
 		attributes: {
 			align: 'wide',
--- a/src/blocks/opentable/inspector.js
+++ b/src/blocks/opentable/inspector.js
@@ -1,0 +1,38 @@
+/**
+ * External dependencies.
+ */
+
+/**
+ * Internal dependencies
+ */
+
+/**
+ * WordPress dependencies.
+ */
+import { __ } from '@wordpress/i18n';
+import { PanelBody, TextControl } from '@wordpress/components';
+import { InspectorControls } from '@wordpress/block-editor';
+
+const Inspector = ( props ) => {
+	const {
+		attributes,
+		setAttributes,
+	} = props;
+
+	return (
+		<InspectorControls>
+			<PanelBody title={ __( 'OpenTable settings', 'coblocks' ) }>
+				<TextControl
+					label={ __( 'Restaurant ID', 'coblocks' ) }
+					className="components-text-control--opentable-restaurant-id"
+					value={ attributes.restaurantID }
+					onChange={ ( rID ) => {
+						setAttributes( { restaurantID: rID } );
+					} }
+				/>
+			</PanelBody>
+		</InspectorControls>
+	);
+};
+
+export default Inspector;
--- a/src/blocks/opentable/opentable.js
+++ b/src/blocks/opentable/opentable.js
@@ -1,0 +1,64 @@
+
+/**
+ * Internal dependencies
+ */
+import { Component, Fragment } from '@wordpress/element';
+
+class OpenTable extends Component {
+	constructor() {
+		super( ...arguments );
+
+		this.state = {
+			selectedImage: null,
+		};
+	}
+
+	// componentDidUpdate( prevProps ) {
+	// 	if ( ! this.props.isSelected && prevProps.isSelected ) {
+	// 		this.setState( {
+	// 			selectedImage: null,
+	// 		} );
+	// 	}
+	// }
+
+	render() {
+		return (
+			<Fragment>
+				<div className="iframe__overflow-wrapper">
+					{ { wide: (
+						<iframe
+							frameBorder="0"
+							style={ { width: '840px', height: '200px', display: 'block', margin: 'auto' } }
+							title="open table frame"
+							src={ `//www.opentable.com/widget/reservation/canvas?rid=${ this.props.attributes.restaurantID }$&domain=com&type=standard&theme=wide&overlay=false&insideiframe=true` }
+						/>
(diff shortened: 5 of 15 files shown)


package.json
```diff
@@ -69,7 +69,7 @@
 		"zh_TW": "繁體中文"
 	},
 	"dependencies": {
-		"@godaddy-wordpress/coblocks-icons": "1.3.0"
+		"@godaddy-wordpress/coblocks-icons": "^1.5.3"
```
ehamwey, line 72: First observation - this should be `1.6.0` since that is the latest version and already merged on `master`

AnthonyLedesma:
Impressive stuff here @ehamwey! I have a few comments.


I should not be able to access an edit state before the user confirms an initial selection. The edit button could be hidden until a restaurant id is set.
openTableEditState

AnthonyLedesma:
Looks like the example styles might need some attention.
openTableExampleStyle

AnthonyLedesma:
There seem to be some downsides here with the way in which block styles function. Multiple block instances are initialized which causes a number API fetches to OpenTable. Because of the transient nature of the OpenTable Iframe, there are situations where styles are not rendered consistently.
openTableStyles

I think we should consider non-standard behavior here by adding custom controls for block styles. 
image

This will allow us greater control over the interface such as setting a static image to represent block styles thereby forgoing superfluous API requests.

Having custom controls we could also then hide the button style more easily in the case of multiple restaurant IDs.

It would even be possible to mimic the standard block styles UI to achieve a similar UX.
```
<PanelBody title={ __( 'Styles', 'coblocks' ) } className="block-coblocks__inspector-block-styles">
   ...
</ PanelBody>
```

src/blocks/opentable/edit.js
```diff
@@ -0,0 +1,148 @@
...
+				encodeURIComponent( token )
+		)
+			.then( ( response ) => response.json() )
+			.then( ( json ) => {
+				setQueryResults( [] );
+				for ( const item in json.items ) {
+					const name = decodeURIComponent( json.items[ item ].name ) + ' (' + json.items[ item ].rid + ')';
+					setQueryResults( ( oldQueryResults ) => [ ...oldQueryResults, name ] );
+				}
+			} );
```
AnthonyLedesma, line 58: On line 56 `setQueryResults` setter function is used in a loop. While we preserve old results, there would be excess calls to this function where it could be avoided. I would suggest setting a variable on line 53, add to it on line 56, and finally execute `setQueryResults` outside of the loop on line 58.

src/blocks/opentable/index.js
```diff
@@ -0,0 +1,75 @@
...
+			name: 'tall',
+			/* translators: block style */
+			label: __( 'Tall', 'coblocks' ),
+		},
+		{
+			name: 'button',
+			/* translators: block style */
+			label: __( 'Button', 'coblocks' ),
+		},
+	],
```
AnthonyLedesma, line 58: These might be removed. Related to my comment here:
https://github.com/godaddy-wordpress/coblocks/pull/1928#issuecomment-888370643

src/blocks/opentable/inspector.js
```diff
@@ -0,0 +1,55 @@
...
+		<InspectorControls>
+			{ ( className.substring( className.lastIndexOf( '-' ) + 1 ) === 'button' ) && ( attributes.restaurantIDs.length > 1 ) &&
+			<Notice
+				isDismissible={ false }
+				status="warning">
+				{ __( 'The button style does not support multiple restaurants. Clicking the button navigates to the first provided Restaurant.', 'coblocks' ) }
+			</Notice>
+			}
+			{ /* language selector */ }
+			<PanelBody title={ __( 'OpenTable settings', 'coblocks' ) } className="block-coblocks__inspector-block-settings-panel-body">
```
AnthonyLedesma, line 33: We might add custom style-controls here. Related to my comment:
https://github.com/godaddy-wordpress/coblocks/pull/1928#issuecomment-888370643

src/blocks/opentable/inspector.js
```diff
@@ -0,0 +1,55 @@
...
+					} }
+					options={ [
+						{ value: 'en-US', label: 'English-US' },
+						{ value: 'fr-CA', label: 'Français-CA' },
+						{ value: 'de-DE', label: 'Deutsch-DE' },
+						{ value: 'es-MX', label: 'Español-MX' },
+						{ value: 'ja-JP', label: '日本語-JP' },
+						{ value: 'nl-NL', label: 'Nederlands-NL' },
+						{ value: 'it-IT', label: 'Italiano-IT' },
+					] } >
```
AnthonyLedesma, line 48: Are these all of the languages available? It would be very cool to pass in the currently set WordPress language into the API.

src/blocks/opentable/opentable.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+class OpenTable extends Component {
+	constructor() {
+		super( ...arguments );
+
+		this.state = {
+			showLoading: true,
+		};
+	}
```
AnthonyLedesma, line 26: This should to be a functional component instead of a class component.

src/blocks/opentable/save.js
```diff
@@ -0,0 +1,59 @@
...
+import { transformRIDs } from './opentable';
+
+export default function save( { attributes } ) {
+	const rids = attributes.restaurantIDs;
+	const className = ( attributes.className === undefined ) ? '' : attributes.className.substring( attributes.className.lastIndexOf( '-' ) + 1 );
+	const ridsString = transformRIDs( rids, className );
+	const isMultiple = rids.length > 1;
+
+	const styles = [ 'tall', 'standard', 'button' ];
+	if ( ( attributes.className === undefined || attributes.className === null ) || ( ! styles.includes( className ) ) ) {
```
AnthonyLedesma, line 11: `( attributes.className === undefined || attributes.className === null )`

This could be simplified.
`! attributes?.className`

AnthonyLedesma (changes requested):

src/blocks/opentable/inspector.js
ehamwey: Yes, these are the officially supported languages for the block (I think because of the markets OpenTable is in)

src/blocks/opentable/edit.js
```diff
@@ -0,0 +1,148 @@
...
+	};
+	useEffect( () => {
+		setRidField( attributes.restaurantIDs );
+	}, [] );
```
AnthonyLedesma, line 62: This `useEffect` call would fire on every render thus `setRidField( attributes.restaurantIDs );` would fire on every render. This could be solved by adding `attributes.restaurantIDs` to the dependency array on line 62 and by adding 
```
if ( prevIDs !== attributes.restaurantIDs ) {

}
```
conditional on the setter. 

You can access previous props like this.
https://github.com/godaddy-wordpress/coblocks/blob/aacd25ceb1cf712bbe1f852f197922ca6246de85/src/blocks/gallery-collage/edit.js#L32-L33

ehamwey:
> Looks like the example styles might need some attention.

How did you reproduce this? Example styles working fine on my end.

ehamwey:
> Impressive stuff here @ehamwey! I have a few comments.
> 
> I should not be able to access an edit state before the user confirms an initial selection. The edit button could be hidden until a restaurant id is set.
> openTableEditState

Resolved in 6f24ca5

src/blocks/opentable/edit.js
ehamwey: Question on this - since the useEffect array parameter is an empty array rather than being not provided at all, wouldn't the above code only fire on the first render?

jasonlemay: Yes, it should act as a `componentDidMount()`

AnthonyLedesma: My mistake! 👍

ehamwey:
> There seem to be some downsides here with the way in which block styles function. Multiple block instances are initialized which causes a number API fetches to OpenTable. Because of the transient nature of the OpenTable Iframe, there are situations where styles are not rendered consistently.
> 
> I think we should consider non-standard behavior here by adding custom controls for block styles.
> 
> This will allow us greater control over the interface such as setting a static image to represent block styles thereby forgoing superfluous API requests.
> 
> Having custom controls we could also then hide the button style more easily in the case of multiple restaurant IDs.
> 
> It would even be possible to mimic the standard block styles UI to achieve a similar UX.
> 
> ```
> <PanelBody title={ __( 'Styles', 'coblocks' ) } className="block-coblocks__inspector-block-styles">
>    ...
> </ PanelBody>
> ```
This was resolved in ee11120. Simple change - just made the style pickers static images with background css and hid the iframe enclosing div.

src/blocks/opentable/edit.js
```diff
@@ -0,0 +1,184 @@
...
+			'https://www.opentable.com/widget/reservation/restaurant-search?pageSize=15' +
+				'&query=' +
+				encodeURIComponent( token )
+		)
+			.then( ( response ) => response.json() )
+			.then( ( json ) => {
+				const results = [];
+				for ( const item in json.items ) {
+					const r = json.items[ item ];
+					const name = `${ decodeURIComponent( r.name ) },  ${ decodeURIComponent( r.addressResponse.city ) }, ${ decodeURIComponent( r.addressResponse.country ) } (${ r.rid })`;
```
AnthonyLedesma, line 50: For the sake of readability and posterity, we should try to avoid the use of single-letter variable names `const r =`. Something like `itemProperties`, or `itemProps` perhaps might be more understandable at a glance?

src/blocks/opentable/edit.js
```diff
@@ -0,0 +1,184 @@
...
+			} );
+	};
+	useEffect( () => {
+		if ( prevIDs !== attributes.restaurantIDs ) {
+			setRidField( attributes.restaurantIDs );
+		}
+	}, [ attributes.restaurantIDs ] );
+
+	useEffect( () => {
+		if ( attributes.language === '' && typeof coblocksBlockData.localeCode !== 'undefined' ) {
```
AnthonyLedesma, line 68: In this conditional, if `coblocksBlockDatra.localCode` is undefined then the default switch action does not fire. Is this the intended behavior?
```
default:
	props.setAttributes( { language: 'en-US' } );
	break;
```

I was just thinking that you could create a function 
```javascript
const getLocaleCode = () => {
  if ( typeof coblocksBlockData.localeCode !== 'undefined' ) {
    return coblocksBlockData.localeCode;
  }
  return '';
};
```

Then you could always run the switch on mount against the custom function. If for some reason there is an undefined localeCode then it will still result in the default action.

src/blocks/opentable/edit.js
```diff
@@ -0,0 +1,184 @@
...
+					const r = json.items[ item ];
+					const name = `${ decodeURIComponent( r.name ) },  ${ decodeURIComponent( r.addressResponse.city ) }, ${ decodeURIComponent( r.addressResponse.country ) } (${ r.rid })`;
+					results.push( name );
+				}
+				setQueryResults( results );
+				if ( results.length === 0 ) {
+					setNoResultsFound( true );
+				} else {
+					setNoResultsFound( false );
+				}
```
AnthonyLedesma, line 58: This could be simplified to 
```javascript
setNoResultsFound( !! results?.length === 0 );
```

src/blocks/opentable/edit.js
ehamwey: This is a good suggestion. The original intended behavior was to leave the language as `''` in `attributes`, which still renders the widget correctly, and displays this in the language selector to prompt the user to select a language: 
image

The better solution is to always set a language if unset. I assume this is the safer route, so resolved in 897960c

src/blocks/opentable/edit.js
ehamwey: resolved in 897960c

src/blocks/opentable/edit.js
ehamwey: resolved in 897960c

src/blocks/opentable/save.js
```diff
@@ -0,0 +1,63 @@
...
+			), button: (
+				<iframe
+					id="opentable-iframe"
+					frameBorder="0"
+					scrolling="no"
+					style={ { width: '210px', height: '116px', display: 'block', margin: 'auto' } }
+					title="open table frame"
+					src={ `//www.opentable.com/widget/reservation/canvas?rid=${ ridsString }&domain=com&lang=${ attributes.language }&theme=standard&overlay=false&insideiframe=true` }
+				/>
+			) }[ className ]
```
jrtashjian, line 59: I like this a lot! Smart way to write this type of conditional and it's very clear what is happening.

src/blocks/opentable/edit.js
```diff
@@ -0,0 +1,192 @@
...
+
+	const getLocaleCode = () => {
+		if ( typeof coblocksBlockData.localeCode !== 'undefined' ) {
+			return coblocksBlockData.localeCode;
+		}
+		return '';
+	};
+
+	//searches the opentable reservation network for restaurants with the given name or ID
+	//TODO: 400/500 error handling and show loading spinner
```
jrtashjian, line 48: Let's make sure to handle these before merging.

src/blocks/opentable/edit.js
```diff
@@ -0,0 +1,192 @@
...
+						instructions={ __(
+							'Select your OpenTable restaurant(s) to show the reservations widget.',
+							'coblocks'
+						) }
+						isColumnLayout={ true }
+					>
+
+						{ /* <form onSubmit={ renderOpenTable }> */ }
+						<div className="components-placeholder__flex-fields">
+							<FormTokenField
```
jrtashjian, line 132: Let's add the `maxLength` prop to this component so that a user can't search for and add multiple restaurants.

jrtashjian (changes requested):
This looks great and robust. I just left a few notes that should be resolved before merging.

Well done!

src/blocks/opentable/edit.js
jrtashjian: I missed the dropdown when multiple restaurants are selected. Marking this as resolved.

src/blocks/opentable/edit.js
```diff
@@ -0,0 +1,203 @@
...
+				{ ( isEditing ) ? (
+					<Placeholder
+						icon={ <Icon icon={ icon } /> }
+						label={ __( 'OpenTable', 'coblocks' ) }
+						instructions={ __(
+							'Select your OpenTable restaurant(s) to show the reservations widget.',
+							'coblocks'
+						) }
+						isColumnLayout={ true }
+					>
```
jrtashjian, line 135: The `Placeholder` component has a `notices` prop you can pass the notices to and it will render them at the top with proper styling applied to maintain a consistent UX.

https://github.com/WordPress/gutenberg/blob/wp/5.8/packages/components/src/placeholder/index.js#L16-L30

Example from the `core/file` block:

image

ehamwey: resolved in 9d8be93

ehamwey: image
It seems to have some odd padding behavior. I can confirm this is not being applied from any of the opentable stylesheets.

ehamwey: Both of these styles are coming from core (the extra padding and `display: flex` causing the block not to fully expand to the placeholder size) so going to ignore for now.

jrtashjian: I think you'll need to utilize the `withNotices` HoC which passes a method to create a notice as well as the UI the prop expect.

File block example:
- `withNotices`
https://github.com/WordPress/gutenberg/blob/wp/5.8/packages/block-library/src/file/edit.js#L313

- Props passed
https://github.com/WordPress/gutenberg/blob/wp/5.8/packages/block-library/src/file/edit.js#L66-L67

- Creating the notice
https://github.com/WordPress/gutenberg/blob/wp/5.8/packages/block-library/src/file/edit.js#L104

- Passing `noticesUI` to `Placeholder`
prop https://github.com/WordPress/gutenberg/blob/wp/5.8/packages/block-library/src/file/edit.js#L203

jrtashjian: And for reference this is the `withNotices` HoC source: https://github.com/WordPress/gutenberg/blob/wp/5.8/packages/components/src/higher-order/with-notices/index.js

It looks like it's wrapping Notice components within a `NoticeList` component which might include the additional styles.

ehamwey: https://user-images.githubusercontent.com/9936886/128420681-f2c1abd3-851a-4fbd-b418-307709db1c22.mov

This worked. The off-center X is driving me crazy though, haha.

ehamwey: fixed:
- dismiss button alignment 
- use withNotices HOC instead for UX
in e9bd4e5

src/blocks/opentable/edit.js
```diff
@@ -0,0 +1,204 @@
...
+		}
+		return '';
+	};
+
+	//searches the opentable reservation network for restaurants with the given name or ID
+	const searchRestaurants = useCallback( debounce( ( token ) => {
+		setIsLoading( true );
+		noticeOperations.removeAllNotices();
+		fetch(
+			'https://www.opentable.com/widget/1reservation/restaurant-search?pageSize=15' +
```
jrtashjian, line 54: Is `widget/1reservation` a typo? I think this should be `widget/reservation`.

ehamwey: Oops. That shouldn’t have made it into a commit - that’s how I was testing the error handling.

src/blocks/opentable/edit.js
```diff
@@ -0,0 +1,190 @@
...
+									onChange={ setRidField }
+									onInputChange={ searchRestaurants }
+									__experimentalHowTo={ false }
+									tokenizeOnSpace={ false }
+								/>
+								<Button
+									isPrimary={ !! ridField }
+									isSecondary={ ! ridField }
+									type="submit"
+									disabled={ ridField.length < 1 }
```
AnthonyLedesma, line 151: This logic does not work as expected due to the way in which JavaScript evaluates empty arrays.
Here are some evaluation examples.:
```javascript
isPrimary={ !! ridField }
 > const ridField = [] 
 > !! ridField | true
 
  > const ridField = [ 123, 212 ] 
  > !! ridField | true
  
 isSecondary={ ! ridField }
  > const ridField = [] 
 > ! ridField | false
 
  > const ridField = [ 123, 212 ] 
  > ! ridField | false
```

This is because an empty array is seen as truthy value in JavaScript. In this case, it would be wise to use the `!!` operator with the array.length property. 

```javascript
!! 0 | false
!! 1 | true

! 0 | true
! 1 | false
```

As a result, it looks like the button only has the `isPrimary` prop and not the `isSecondary`. Here is a screenshot of the secondary style applied and it looks like it is missing some margin/padding.
image

AnthonyLedesma (commented):
Comment here on the button logic

src/blocks/opentable/edit.js
AnthonyLedesma: I went ahead and completely removed this logic so that the button has a constant `isPrimary` prop instead. The primary styles have all been handled. If you would like to introduce styles for the `isSecondary` button state and add the logic back I am fine with that too.

richtabor:
UX LGTM

PR #1932, opened 2021-07-08
Fix button size controls on Share and Social Profiles blocks

### LINKED ISSUE #1926: ISBAT use button size controls with Social blocks
### Describe the bug:

Share and Social Profiles block have button size controls available. The E2E tests are written in a way that incorrectly tests the expected output that these attributes should achieve.

Editor markup is appended with size classes.
image

Output markup does not contain size classes and output size is not as expected.
image

Relevant tests are skipped due to this bug and should be fixed when this bug is resolved.

### To reproduce:

1. Add Share/Social Profiles block to content.
2. Open block sidebar and use block size controls.
3. Save page and view published result.
4. See discrepancy between editor and front-end.

### Expected behavior:

Expect markup to match between editor and front-end. 
Changes almost certainly will require block deprecation.

### Isolating the problem:

### WordPress version:

Happens in 5.7.2 as well as 5.8.

### Gutenberg version:

N/A

snovosel:
### Description
This PR applies the block size classnames that are set within the editor on the resulting front end class names.
Fixes #1926 

### Types of changes

Bug Fix

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/share/index.php
+++ b/src/blocks/share/index.php
@@ -219,7 +219,7 @@ function coblocks_render_share_block( $attributes ) {
 		$class .= ' has-colors';
 	}
 
-	if ( isset( $attributes['size'] ) && 'med' !== $attributes['size'] ) {
+	if ( isset( $attributes['size'] ) && 'med' !== $attributes['size'] && strpos( $attributes['className'], 'is-style-mask' ) === false ) {
 		$class .= ' has-button-size-' . $attributes['size'];
 	}
 
--- a/src/blocks/share/index.php
+++ b/src/blocks/share/index.php
@@ -219,7 +219,7 @@ function coblocks_render_share_block( $attributes ) {
 		$class .= ' has-colors';
 	}
 
-	if ( isset( $attributes['size'] ) && 'med' !== $attributes['size'] && ( isset( $attributes['className'] ) && strpos( $attributes['className'], 'is-style-mask' ) === false ) ) {
+	if ( isset( $attributes['size'] ) && 'med' !== $attributes['size'] ) {
 		$class .= ' has-button-size-' . $attributes['size'];
 	}
 
--- a/src/blocks/share/index.php
+++ b/src/blocks/share/index.php
@@ -219,7 +219,7 @@ function coblocks_render_share_block( $attributes ) {
 		$class .= ' has-colors';
 	}
 
-	if ( isset( $attributes['size'] ) && 'med' !== $attributes['size'] && strpos( $attributes['className'], 'is-style-mask' ) === false ) {
+	if ( isset( $attributes['size'] ) && strpos( $attributes['className'], 'is-style-mask' ) === false ) {
 		$class .= ' has-button-size-' . $attributes['size'];
 	}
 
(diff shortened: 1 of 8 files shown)


jrtashjian (changes requested):
Let's remove the `.skip()` from the associated tests to verify this is working as expected.

src/blocks/social-profiles/test/social-profiles.cypress.js
```diff
@@ -249,7 +249,7 @@ describe( 'Test CoBlocks Social Profiles Block', function() {
 		helpers.toggleSettingCheckbox( 'Social colors' );
 
 		cy.get( '.block-editor-writing-flow button[aria-label="Add Facebook profile"]' )
-			.should( 'have.css', 'background-color', 'rgb(40, 48, 61)' );
+			.should( 'have.css', 'background-color', 'rgb(200, 106, 25)' );
```
AnthonyLedesma: I am fine with this change but I do not understand why this was needed.

AnthonyLedesma: Oh, I see now that this was a revert.

src/blocks/social-profiles/index.php
```diff
@@ -160,7 +160,7 @@ function coblocks_render_social_profiles_block( $attributes ) {
 		$class .= ' has-colors';
 	}
 
-	if ( isset( $attributes['size'] ) && 'med' !== $attributes['size'] && ( isset( $attributes['className'] ) && strpos( $attributes['className'], 'is-style-mask' ) === false ) ) {
```
AnthonyLedesma, line 163: For some reason, we were intended for this `has-button-size-` class logic to be excluded when we have the 
`mask` style set. 
```js
strpos( $attributes['className'], 'is-style-mask' ) === false )
```
It would be prudent to understand why this decision was made and to purposefully exclude this specific requirement if we determine the logic is not needed.

src/blocks/share/index.php
```diff
@@ -219,7 +219,7 @@ function coblocks_render_share_block( $attributes ) {
 		$class .= ' has-colors';
 	}
 
-	if ( isset( $attributes['size'] ) && 'med' !== $attributes['size'] && ( isset( $attributes['className'] ) && strpos( $attributes['className'], 'is-style-mask' ) === false ) ) {
```
AnthonyLedesma, line 222: For some reason, we were intended for this `has-button-size-` class logic to be excluded when we have the 
`mask` style set. 
```js
strpos( $attributes['className'], 'is-style-mask' ) === false )
```
It would be prudent to understand why this decision was made and to purposefully exclude this specific requirement if we determine the logic is not needed.

AnthonyLedesma (changes requested):
This PR looks good. Great work. I found a segment of logic that was removed without replacement. Do you have an understanding of what/why this logic existed? Please review and let me know what you find. I wonder if there might be style conflicts or perhaps it was UX decision that was made. It could have also been un-intentional. Whatever the case the understanding would help.

jrtashjian:
If @AnthonyLedesma is ok with the changes I would merge in the latest from the master branch. I think @ehamwey resolved the CSS linting failure.

AnthonyLedesma:
Just tested and things are looking good. I think once we have a few tests this is good to approve and merge.

AnthonyLedesma (approved):

PR #1937, opened 2021-07-15
Fix Social block undefined index

AnthonyLedesma:
### Description
Assertions made on the `className` attribute result in undefined index errors. We can prevent this by first checking `isset` on the object property.


snovosel-godaddy (approved):

ehamwey:
LGTM pending tests. 👍

ehamwey (approved):

Issue #1911, opened 2021-06-01
Coblocks styles are not loaded in FSE templates


janw-me:
### Describe the bug:
Co blocks styles are not loaded for FSE templates.
I know FSE is not released yet, this is more a headsup.

### To reproduce:
1. Clean wp install.
2. Install & activate gutenberg
3. Activate [tt1-blocks](https://wordpress.org/themes/tt1-blocks/) theme. (or another FSE theme)
4. Edit a site template
5. Add a co block like the price table. In the footer or header (not in the main content).
6. Save and preview on frontend.

### Isolating the problem:

- [x] This bug happens with no other plugins activated 
- [x] This bug happens with a default WordPress theme active
- [ ] This bug happens **without** the Gutenberg plugin active
- [x] I can reproduce this bug consistently using the steps above

### WordPress version:
5.7.2

### Gutenberg version:
10.7.1

The styles appear to be loaded when a coblock is in the_content.
They don't appear to be loaded when a coblock is in a template.

PR #1939, opened 2021-07-16
Load CoBlocks assets when wp_template_part post type contains a CoBlocks block

EvanHerman:
### Description
In FSE when a block isn't contained in the post content, the assets do not load. This PR queries the `` psot type and determines if they contain any CoBlocks blocks in their content. If it does, the assets are loaded.

Related issue #1911 

### Types of changes
New functionality (non-breaking change)

### How has this been tested?
Tested against #1911 to ensure the assets load when a CoBlocks block is added to the footer of a page.

### Checklist:
- [x] My code is tested


AnthonyLedesma (approved):
LGTM

PR #1943, opened 2021-07-20
Convert Gist Edit component to React hooks

olafleur:
### Description
Refactor the Gist Edit component from React class to React hooks. It was needed because `withState` is now deprecated, so we need to use `useState`

### How has this been tested?
The unit and Cypress tests for that component still pass.

### Checklist:
- [X] My code is tested
- [X] I've included any necessary tests 
- [X] I've added proper labels to this pull request


AnthonyLedesma (approved):
LGTM. I pushed up a change here that will generate the callback id only a single time on the initial mount. The previous behavior was to generate a new callback ID for each render.

PR #1949, opened 2021-07-22
Fix Font Size controls in WordPress 5.8

### LINKED ISSUE #1945: FontSizePicker not working properly after WordPress 5.8 update
### Describe the bug:
After WordPress update to 5.8, when try to change the font size of your custom blocks it shows that the content is unexpected

### To reproduce:
For example, create the author block and type something, then try to change font size, then update the page and refresh it.

### Expected behavior:
I expect no crashes after font size change. I think the problem appeared because of the new additional unit sizes for Font Size picker.

### Screenshots:
https://prnt.sc/1e2iaph, https://prnt.sc/1e2iokg

### Isolating the problem:

### WordPress version:
WordPress 5.8

### Gutenberg version:
Gutenberg 10.7

AnthonyLedesma:
### Description

Closes #1945. [Due to a bug in 5.8 release](https://github.com/WordPress/gutenberg/issues/33653), the `withFontSizes` HOC is conflicting with the `FontSizePicker` component. Changes here fix logic used with the `FontSizePicker` to allow use in 5.7.2 as well as 5.8.

### Types of changes

New CoBlocks shared component where we utilize `FontSizePicker` with improved logic.

### How has this been tested?

This was tested manually by switching versions 5.7.2 and 5.8.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/highlight/test/highlight.cypress.js
+++ b/src/blocks/highlight/test/highlight.cypress.js
@@ -39,7 +39,9 @@ describe( 'Block: Highlight', function() {
 	/**
 	 * Test the accordion block content font settings
 	 */
-	it( 'Test highlight block font size setting.', function() {
+	// Disabled because fontSize controls are hidden in 5.8. This should be fixed when we can fix the fontSize controls.
+	// eslint-disable-next-line jest/no-disabled-tests
+	it.skip( 'Test highlight block font size setting.', function() {
 		cy.get( 'p.wp-block-coblocks-highlight' ).find( 'mark' )
 			.type( 'highlighted text' );
 
--- a/src/blocks/highlight/test/highlight.cypress.js
+++ b/src/blocks/highlight/test/highlight.cypress.js
@@ -39,9 +39,7 @@ describe( 'Block: Highlight', function() {
 	/**
 	 * Test the accordion block content font settings
 	 */
-	// Disabled because fontSize controls are hidden in 5.8. This should be fixed when we can fix the fontSize controls.
-	// eslint-disable-next-line jest/no-disabled-tests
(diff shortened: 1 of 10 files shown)


Mamaduka:
The problem is with the `FontSizePicker` component. The latest version uses `UnitControl` for the custom font size; it returns numbers with units - `12px,` and I think `withFontSizes` expects just numbers.

Updating `FontSizePicker` to something like this should fix the issue:

```js
<FontSizePicker
	fontSizes={ fontSizes }
	fallbackFontSize={ fallbackFontSize }
	value={ fontSize.size }
	onChange={ ( value ) => {
		const fontSizeValue = value ? parseInt( value, 10 ) : undefined;

		if ( ! Number.isNaN( fontSizeValue ) ) {
			setFontSize( fontSizeValue )
		}
	} }
/>
```

src/blocks/highlight/test/highlight.cypress.js
```diff
@@ -39,7 +39,9 @@ describe( 'Block: Highlight', function() {
 	/**
 	 * Test the accordion block content font settings
 	 */
-	it( 'Test highlight block font size setting.', function() {
+	// Disabled because fontSize controls are hidden in 5.8. This should be fixed when we can fix the fontSize controls.
+	// eslint-disable-next-line jest/no-disabled-tests
+	it.skip( 'Test highlight block font size setting.', function() {
```
olafleur, line 44: should you remove the `skip` here, now that you have a fix?

AnthonyLedesma: Good call. Just pushed up some fixes to that test.

jrtashjian (approved):
This works for me in both WP 5.7.2 and 5.8.

PR #1968, opened 2021-07-27
Fix invisble block states enabled by animation class

### LINKED ISSUE #1941: ISNBAT enable invisible block states with `coblocks-animate` class
### Describe the bug:

Discovered in https://github.com/godaddy-wordpress/coblocks/issues/1879 but not officially reported is a situation where the Media and Text block is able to enter an unusable state. This can only happen when the CoBlocks plugin is disabled and re-enabled later. While perhaps not a typical user interaction, the enabling and disabling of common plugins are steps that many developers take when simply troubleshooting website issues. It is reasonable to believe this situation could happen in the wild and should be handled as a sort of deprecation hopefully.

This behavior occurs with the 'core/media-text' 'core/image' 'core/group' 'core/gallery' 'core/cover' and 'core/columns' blocks.

### To reproduce:

1. Create new post and add the Media and Text block.
2. Edit the Media and Text block to have an image and enable any of the animation control settings.
3. Save post.
4. Using WP-CLI we first disable CoBlocks. `wp plugin deactivate coblocks`
5. Reload post and notice (expected) block error. Attempt block recovery. Save post.
6. Using WP-CLI we activate CoBlocks. `wp plugin active coblocks`
7. Reload post and notice that Media and Text are hidden with `opacity: 0`.
8. Notice that attributes are not parsed due to user interaction to Attempt block recovery.

### Expected behavior:

This interaction is a complicated one but the UX impact is quite large with hidden blocks in the editor.

### Screenshots:

image

### Isolating the problem:

### WordPress version:

5.7.2

### Gutenberg version:

N/A

snovosel:
Fixes #1941 

### Description

If an animation (coblocks extension) is added to a block, and the coblocks plugin is deactivated, block validation will fail and block recovery will remove the animation attribute, however the classname `coblocks-animate` will be kept. If coblocks is re-activated , as the animation is undefined and the className is set, an opacity of 0 is set to the block. This PR attempts to add a block filter to remove this class name in the event of an undefined animation attribute.

### Types of changes

Bug fix

### How has this been tested?

visually

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/extensions/animation/index.js
+++ b/src/extensions/animation/index.js
@@ -136,14 +136,24 @@ const withAnimationSettings = createHigherOrderComponent( ( BlockListBlock ) => 
 	return enhance( ( { select, ...props } ) => {
 		let wrapperProps = props.wrapperProps;
 
-		const block = select( 'core/block-editor' ).getBlock( props.rootClientId || props.clientId );
 		const blockName	= select( 'core/block-editor' ).getBlockName( props.rootClientId || props.clientId );
 
-		if ( ! allowedBlocks.some( ( allowedBlock ) => allowedBlock.blockType === blockName && ! allowedBlock.animateChildren ) || ! block?.attributes?.animation ) {
+		// Block not supported return unmodified.
+		if ( ! allowedBlocks.some( ( allowedBlock ) => allowedBlock.blockType === blockName && ! allowedBlock.animateChildren ) ) {
 			return <BlockListBlock { ...props } />;
 		}
 
-		const { animation } = block.attributes;
+		if ( ! props.attributes?.animation ) {
+			// If animation is unset ensure that `coblocks-animate` class is unset.
+			if ( props.attributes?.className?.includes( 'coblocks-animate' ) ) {
+				const classList = props.attributes?.className.split( ' ' );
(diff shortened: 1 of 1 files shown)


src/extensions/animation/index.js
```diff
@@ -186,3 +186,23 @@ addFilter(
 	'coblocks/withAnimationSettings',
 	withAnimationSettings
 );
+
+const ALLOWED_ANIMATION_BLOCKS = [ 'core/media-text' ];
```
jlemay-godaddy, line 190: What is it doing? If it's checking for blocks with animation enabled, there are a lot more

snovosel: Should we open this up to remove the animate class name if the animation is undefined for all blocks? According to the issue, this is a problem for the following blocks:

```
'core/media-text',
	'core/image',
	'core/group',
	'core/gallery',
	'core/cover',
	'core/columns'
```

jlemay-godaddy: If the problem has a chance to happen, we should ensure that it won't. So yeah, I would target all the blocks that are animation-capable

AnthonyLedesma (approved):
LGTM.
Pushed up some major changes to exclude use of an addtional filter.

PR #1971, opened 2021-07-30
Update to Cypress 8.1.0

AnthonyLedesma:
### Description

Update to Cypress 8.1.0.
Ensure that Firefox 90.0.2 is installed for E2E tests by using Cypress Orb `browser-tools/install-browser-tools`.

### Types of changes

Cypress config changes.

### How has this been tested?

Tested full E2E suite.

### Checklist:
- [x] My code is tested


AnthonyLedesma:
Tests seem fine in chrome.

In Firefox:
```
Opening Cypress...
The `firefoxGcInterval` configuration option was removed in Cypress version `8.0.0`. It was introduced to work around a bug in Firefox 79 and below.

Since Cypress no longer supports Firefox 85 and below in Cypress 8, this option was removed.

You can safely remove this option from your config.
Cypress does not support running Firefox version 82. To use Firefox with Cypress, install a version of Firefox newer than or equal to 86.


Exited with code exit status 1

CircleCI received exit code 1
```

AnthonyLedesma:
Some unhandled exceptions and unhandled rejections may occur on WordPress that are not necessarily related to our components. Occasionally we might see these errors due to rapid interaction in the editor.

We should consider disabling or otherwise logging the now more aggressive behavior of Cypress to fail tests with unhandled exceptions.
https://docs.cypress.io/guides/references/migration-guide#Uncaught-exception-and-unhandled-rejections

In 7.0, Cypress now fails tests in more situations where there is an uncaught exception and also if there is an unhandled promise rejection in the application under test.

You can ignore these situations and not fail the Cypress test with the code below.
Turn off all uncaught exception handling
```javascript
Cypress.on('uncaught:exception', (err, runnable) => {
  // returning false here prevents Cypress from
  // failing the test
  return false
})
```

Turn off uncaught exception handling unhandled promise rejections
```javascript
Cypress.on('uncaught:exception', (err, runnable, promise) => {
  // when the exception originated from an unhandled promise
  // rejection, the promise is provided as a third argument
  // you can turn off failing the test in this case
  if (promise) {
    // returning false here prevents Cypress from
    // failing the test
    return false
  }
})
```

snovosel (approved):
LGTM

AnthonyLedesma:
Closing in favor of https://github.com/godaddy-wordpress/coblocks/pull/2043

Issue #1844, opened 2021-03-04
ISBAT Post Carousel block options - extending Feed settings


mvlaisavljevic:
### Is your feature request related to a problem? Please describe.
No, this would be just an additional option under the Feed settings for the Post Carousel block.

### Describe the solution you'd like
An option for users to filter only posts that are under multiple categories.

### Additional context/screenshots
Feature request initially submitted on WordPress.org:
https://wordpress.org/support/topic/coblock-post-carousel-by-category-and-or/

PR #1973, opened 2021-07-30
Enable post carousel category relation option to show posts with multiple categories only

EvanHerman:
Original feature request #1844 

### Description
Added the ability to filter posts in the post carousel to allow for only posts with 2 or more categories. Users can switch between 'and'/'or' relations. 'and' only shows posts with 2 or more categories. 'or' will show posts with either or, or both, categories.

I also enabled the same functionality for the posts block, as well.

### Types of changes

### How has this been tested?
Manually tested with various post categories and combination of and/or options set.

### Checklist:
- [x] My code is tested


AnthonyLedesma (changes requested):
I like this update. Two changes are requested. We should only show the inspector controls for the `and` relation if there is more than a single category set. Otherwise, default value should just be `or`. If we do not already we could make use of useEffect logic to set back to `or` if the second category is removed. 

The second is to apply these same changes to the Posts block. Both blocks use the same category controls and we should aim for parity between these blocks.

AnthonyLedesma (approved):
I needed to push up a change here because it is possible for a user to enter a state where there is differing output in the editor and the front end because isEqual does not account for incorrect id order.

```javascript
> lodash.isEqual([7,6],[6,7])
false
```
```javascript
> lodash.isEqual([7,6],[7,6])
true
```
To handle this I first `concat` to prevent mutation then `sort` to order the arrays consistently.

PR #1976, opened 2021-08-04
Server side render Click to Tweet block

### LINKED ISSUE #1663: Click to Tweet in new post turns editing state dirty after publishing
### To reproduce:

1. Start a new post.
2. Insert the "Click to Tweet" block
3. Add some text to the block
4. Click "Publish"
5. In the snack bar that appears, try clicking "View Post".
6. You'll see an alert asking you "Leave app? Changes you made may not be saved". (Furthermore, note that the "Update" button is enabled rather than disabled, indicating that there have been changes that need saving -- even though the post was just published.)

### Expected behavior:
No changes needing saving.

### Screenshots:

Bildschirmfoto von 136-2020-09-03T11-58-44 mpg

### Isolating the problem:

The reason seems to be that the block has a `url` attribute that contains to the current post's URL (plus that URL is also found in the block's content). If the block is inserted into a newly created post, that `url` attribute will be set to something like `https:\/\/example.com\/2020\/09\/03\/auto-draft\/`. If the post is saved, it will be updated to use the post's slug (which is based on the post title), e.g. `https:\/\/example.com\/2020\/09\/03\/click-to-tweet-test\/`.

Overall, it seems like rather bad UX to require the user to save a post that they just published. A potential fix would be to make the block dynamic, and render the current post's URL on the server side (in PHP). This would stabilize against a number of other changes (e.g. change of permalink scheme etc).

Originally found here: https://github.com/Automattic/wp-calypso/pull/45345#issuecomment-686533420

### WordPress version:
5.5

### Gutenberg version:
8.7-8.9

EvanHerman:
Resolves #1663 

### Description
Convert the click to tweet block to a server side rendered block.

### Types of changes
Non-breaking change

### How has this been tested?
Manually tested. The jest tests are currently not passing due to the 'buttonText' attribute getting stripped from the serialized block. This needs to be investigated further.

### Checklist:
- [x] My code is tested

### DIFF
--- a/includes/class-coblocks-register-blocks.php
+++ b/includes/class-coblocks-register-blocks.php
@@ -94,14 +94,6 @@ class CoBlocks_Register_Blocks {
 			)
 		);
 		register_block_type(
-			$slug . '/click-to-tweet',
-			array(
-				'editor_script' => $slug . '-editor',
-				'editor_style'  => $slug . '-editor',
-				'style'         => $slug . '-frontend',
-			)
-		);
-		register_block_type(
 			$slug . '/dynamic-separator',
 			array(
 				'editor_script' => $slug . '-editor',
--- a/src/blocks/click-to-tweet/block.json
+++ b/src/blocks/click-to-tweet/block.json
@@ -4,8 +4,6 @@
 	"attributes": {
 		"content": {
 			"type": "string",
-			"source": "html",
-			"selector": "p",
 			"default": ""
 		},
 		"url": {
--- a/src/blocks/click-to-tweet/index.js
+++ b/src/blocks/click-to-tweet/index.js
@@ -4,7 +4,6 @@
 import edit from './edit';
 import icon from './icon';
 import metadata from './block.json';
-import save from './save';
 import transforms from './transforms';
 import deprecated from './deprecated';
 
@@ -39,7 +38,9 @@ const settings = {
 	attributes,
 	transforms,
 	edit,
-	save,
+	save() {
+		return null;
+	},
 	deprecated,
 };
 
--- a/src/blocks/click-to-tweet/index.php
+++ b/src/blocks/click-to-tweet/index.php
@@ -1,0 +1,84 @@
+<?php
+/**
+ * Server-side rendering of the `click to tweet` block.
+ *
+ * @package WordPress
+ */
+
+/**
+ * Renders the block on server.
+ *
+ * @param array $attributes The block attributes.
+ *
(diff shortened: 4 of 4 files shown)


includes/class-coblocks-register-blocks.php
```diff
@@ -93,14 +93,6 @@ public function register_blocks() {
...
 			)
 		);
-		register_block_type(
-			$slug . '/click-to-tweet',
-			array(
-				'editor_script' => $slug . '-editor',
-				'editor_style'  => $slug . '-editor',
-				'style'         => $slug . '-frontend',
-			)
-		);
```
AnthonyLedesma, line 103: This while file surprised me. I somehow forgot that this file exists but honestly, I wonder why it exists at all. There are only 12 blocks registered here some of which are JavaScript blocks that are registered in JS. There are SSR blocks like the Form block that is not registered here at all but separately within its respective index file.

In fact, I do not see this class used at all outside of a test written for it.

AnthonyLedesma: I was mistaken on this. The file is invoked in class-coblocks.php. Either way, we still have a question of why some SSR blocks are registered elsewhere and why JS blocks are registered here at all. Does superfluous registration have performance impacts for instance? Regardless these are things that should be handled in a different story.

AnthonyLedesma (changes requested):
These changes do not allow the block to deprecate as expected. Here is the markup generated from Master.

```<!-- wp:coblocks/click-to-tweet {"url":"https://wordpress.test/test-2/"} -->
<blockquote class="wp-block-coblocks-click-to-tweet"><p class="wp-block-coblocks-click-to-tweet__text">This is a test</p><a class="wp-block-coblocks-click-to-tweet__twitter-btn" href="http://twitter.com/share?&amp;text=This%20is%20a%20test&amp;url=https://wordpress.test/test-2/" target="_blank" rel="noopener noreferrer">Tweet</a></blockquote>
<!-- /wp:coblocks/click-to-tweet -->```

File: src/blocks/click-to-tweet/deprecated.js [hunk 2]
```diff
@@ -87,7 +87,12 @@ const deprecated = [
 					</blockquote>
 				)
 			);
-		} },
+		},
+	},
+	{
+		attributes,
+		save: () => null,
```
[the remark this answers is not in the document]
Mamaduka: I think @AnthonyLedesma is on the right path here.

I would also place new deprecation as the first item of the `deprecated` array.

PR #1982, opened 2021-08-09
refactor Buttons block to use react hooks

snovosel:
### Description

refactor buttons block and related components to use react hooks

### Checklist:
- [ ] My code is tested

### DIFF
--- a/src/blocks/buttons/controls.js
+++ b/src/blocks/buttons/controls.js
@@ -1,29 +1,26 @@
 /**
  * WordPress dependencies
  */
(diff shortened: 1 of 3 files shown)


AnthonyLedesma (changes requested):
This PR should be closed. The buttons block has been deprecated for some time. I'd like to explore a way for us to separate the data we need to maintain backward compatibility with this Buttons block and maybe others. In the meantime, we should avoid any changes in deprecated blocks.

PR #2015, opened 2021-09-03
Refactor gallery masonry block to use react hooks

snovosel:
### Description

This PR refactors the gallery masonry block to use a React hooks architecture.

### Types of changes

code refactor

### How has this been tested?

visually

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/gallery-masonry/edit.js
+++ b/src/blocks/gallery-masonry/edit.js
@@ -58,7 +58,6 @@ const GalleryMasonryEdit = ( props ) => {
 	} = attributes;
 
 	const [ selectedImage, setSelectedImage ] = useState( null );
-	const [ , setCaptionSelected ] = useState( null );
 
 	const prevIsSelected = usePrevious( isSelected );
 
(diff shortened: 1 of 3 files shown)


src/blocks/gallery-masonry/edit.js
```diff
@@ -32,255 +32,233 @@ const masonryOptions = {
...
+		lightbox,
+		images,
+	} = attributes;
 
-		this.state = {
-			selectedImage: null,
-		};
-	}
+	const [ selectedImage, setSelectedImage ] = useState( null );
+	const [ , setCaptionSelected ] = useState( null );
```
olafleur, line 61: It seems weird that there is no variable. Is it normal?

snovosel: The linter wouldn't allow me to commit without removing the state value of this de-structure. It is valid syntax however I agree that it looks a bit odd. We could disable the linter rule for this specific thing as the state value wasn't being used, so with it present the linter was throwing an unused-var error

olafleur: Yeah, but my question is more : if there is no variable, there is no state, right? So, what is the purpose of the function?

PR #2017, opened 2021-09-07
Fix gutter value in the editor always set to 0 on page load

EvanHerman:
### Description
The gutter controls for the Carousel gallery was not showing properly in the editor. The value was saving to the DB and the front of site looked correct, but the editor always had a `gutter` and `gutterMobile` values of 0. The following code is responsible for setting the gutter value back to 0 on page load. This block of code only seemed to be causing a problem and not necessary for the block to function properly.

```js
useEffect( () => {
	if (
		props.attributes.gridSize === 'xlrg' &&
		prevAlign === undefined &&
		( gutter !== 0 || gutterMobile !== 0 )
	) {
		props.setAttributes( { gutter: 0, gutterMobile: 0 } );
	}
}, [ props.attributes.gridSize, prevAlign ] );
```

### Types of changes

### How has this been tested?
Manually tested each block with the older gutter controls.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/gallery-carousel/edit.js
+++ b/src/blocks/gallery-carousel/edit.js
@@ -47,7 +47,6 @@ const GalleryCarouselEdit = ( props ) => {
 	const [ captionFocused, setCaptionFocused ] = useState( false );
 
 	const prevSelected = usePrevious( props.isSelected );
-	const prevAlign = usePrevious( props.attributes.align );
 
 	useEffect( () => {
 		if ( ! props.isSelected && prevSelected ) {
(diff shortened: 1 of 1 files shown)


src/blocks/gallery-carousel/edit.js
```diff
@@ -66,16 +65,6 @@ const GalleryCarouselEdit = ( props ) => {
...
-	useEffect( () => {
-		if (
-			props.attributes.gridSize === 'xlrg' &&
-			prevAlign === undefined &&
-			( gutter !== 0 || gutterMobile !== 0 )
-		) {
-			props.setAttributes( { gutter: 0, gutterMobile: 0 } );
-		}
-	}, [ props.attributes.gridSize, prevAlign ] );
-
```
AnthonyLedesma, line 78: I believe the original intention here was to prevent a gutter value from setting while the block has xlrg grid size. That logic does not seem to be working as expected and a few UX concerns. If we intend to prevent users from using a gutter while the block is set to xlrg size then we should hide the controls altogether as well as include this logic. 

Actually diving in a bit further I see that the prevAlign may not be calculating correctly or something like that because when the alignment is `unset` the controls are indeed hidden. (Master)
AlignSizeControlsHidden


This functionality should persist in this PR.

AnthonyLedesma (changes requested):

src/blocks/gallery-carousel/edit.js
AnthonyLedesma: We Might be able to give this a quick fix by basing the logic on not what the previous value is but what the current calculated value is. The same way in which we hide the controls.

AnthonyLedesma:
Figured I'd link this issue here as well which may be related.
https://github.com/godaddy-wordpress/coblocks/issues/2024

AnthonyLedesma (approved):
Works great! LGTM.

AnthonyLedesma:
CI error for JS test is safe to merge. It has to do with how we track against changes in master.

PR #2018, opened 2021-09-07
WPEX-1459 - do not throw error on custom entered category for post carousel and posts block

### LINKED ISSUE #1974: ISNBAT crash the Posts and Post Carousel block by using an invalid category id
### Describe the bug:

When choosing a category on the Posts or Post Carousel blocks a user might not select a token value and instead enter in a custom value followed by a keyboard return. This results in a block error state.

### To reproduce:

1. Add Posts block.
2. Open block sidebar and feed settings.
3. Type in a bogus category and hit return to see the error.

### Expected behavior:

Expect error notice perhaps or save the value and take no action. Block errors should not occur.

### Screenshots:

postCrashWithCategory

### Isolating the problem:

### WordPress version:

5.8

### Gutenberg version:

N/A

snovosel:
### Description

This PR handles the use case of a custom category entered into the feed settings for both the Posts and Post Carousel block. If an unrecognized, custom category is entered then no action is taken and no block error will be thrown.

Fixes #1974 

### Types of changes

Bug fix

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/post-carousel/edit.js
+++ b/src/blocks/post-carousel/edit.js
@@ -356,12 +356,7 @@ export default compose( [
 
 		const updatedQuery = () => {
 			const catIds = categories && categories.length > 0
-				? categories.map( ( cat ) => {
-					if ( ! cat ) {
-						return null;
-					}
-					return cat.id;
-				} )
+				? categories.map( ( cat ) => cat?.id )
 				: [];
 
 			const latestPostsQuery = pickBy( {
--- a/src/blocks/post-carousel/edit.js
+++ b/src/blocks/post-carousel/edit.js
@@ -356,7 +356,12 @@ export default compose( [
 
 		const updatedQuery = () => {
 			const catIds = categories && categories.length > 0
-				? categories.map( ( cat ) => cat.id )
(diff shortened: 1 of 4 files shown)


src/blocks/post-carousel/edit.js
```diff
@@ -356,7 +356,12 @@ export default compose( [
 
 		const updatedQuery = () => {
 			const catIds = categories && categories.length > 0
-				? categories.map( ( cat ) => cat.id )
+				? categories.map( ( cat ) => {
+					if ( ! cat ) {
+						return null;
+					}
+					return cat.id;
+				} )
```
AnthonyLedesma, line 364: This can be simplified to `? categories.map( ( cat ) => cat?.id )`

AnthonyLedesma: Is this same change needed within `src/blocks/posts/edit.js`?

AnthonyLedesma (dismissed):
Things are working well and the bug seems to be resolved. Question on one of the changes that you have made to an edit function.

AnthonyLedesma (approved):
LGTM

PR #2027, opened 2021-09-14
Convert Buttons block to hooks

olafleur:
We forgot that one ;) 

Probably because there was no Cypress tests on that block. 
Maybe we should add some.

### DIFF
--- a/src/blocks/buttons/controls.js
+++ b/src/blocks/buttons/controls.js
@@ -1,29 +1,26 @@
 /**
  * WordPress dependencies
  */
-import { Component, Fragment } from '@wordpress/element';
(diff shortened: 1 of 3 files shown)


AnthonyLedesma:
Hi @olafleur. Unfortunately this block was left out intentionally and I seemingly forgot to properly notate that fact within the story. Here is the original comment that I left on @snovosel PR

> This PR should be closed. The buttons block has been deprecated for some time. I'd like to explore a way for us to separate the data we need to maintain backward compatibility with this Buttons block and maybe others. In the meantime, we should avoid any changes in deprecated blocks.

https://github.com/godaddy-wordpress/coblocks/pull/1982

PR #2029, opened 2021-09-15
Convert Gallery Stacked to hooks

olafleur:


AnthonyLedesma:
What do you think about refactoring the shared gallery components as well? That could be part of a separate PR because of the size.

`/src/components/block-gallery/`

olafleur:
Yeah @AnthonyLedesma , that would be a great idea!
But I agree, it would be better in another PR, let's do one thing at a time :)

AnthonyLedesma (changes requested):
Captions are not saving properly. Replicate by setting caption, save page, and refresh to see missing caption.

PR #1997, opened 2021-08-19
Introduce performance tests

AnthonyLedesma:
### Description

Major refactors here to allow compatibility with `wp-scripts test-e2e` for performance testing. In order to run the performance tests Docker desktop must be running and a WordPress env must be setup prior.

1. yarn wp-env start
2. yarn test:performance

Running these commands will give you performance testing results in Console.

After this PR is merged into master CI needs to be set up to compare incoming PRs against master and compare performance results. 

### Screenshots

image

### Types of changes

Major changes across config and files. Several changes for ESLint and updated dependencies.

### How has this been tested?

Tested in CI. Tested manually.

### Checklist:
- [x] My code is tested
- [x] My code follows accessibility standards 
- [x] My code has proper inline documentation 
- [x] I've included any necessary tests 
- [x] I've included developer documentation 
- [x] I've added proper labels to this pull request

PR #2031, opened 2021-09-16
Introduce performance tests

AnthonyLedesma:
### Description

Redux of https://github.com/godaddy-wordpress/coblocks/pull/1997.

Major refactors here to allow compatibility with `wp-scripts test-e2e` for performance testing. In order to run the performance tests Docker desktop must be running and a WordPress env must be setup prior.

1. yarn wp-env start
2. yarn test:performance

Running these commands will give you performance testing results in Console.

After this PR is merged into master CI needs to be set up to compare incoming PRs against master and compare performance results. 

Dev Notes:
CI tests by using `.dev/performance/cli.js perf --ci` command translating the `childProcess.exec` function. This function `spawns a shell then executes the command within that shell` where we perform the work to set up local WordPress installations to test against.

Test against branch
After this PR is merged in there will be incoming PRs to master that will change some behavior here. The way that CI is written right now is to use the `try/performance-tests-run-all-tests` branch as opposed to the `master` branch. This is a specific element that must be altered after merged into master.

### Screenshots

image

### Types of changes

Major changes across config and files. Several changes for ESLint and updated dependencies.

Notice spacing changes, comments excluding lint rules, and issue comments that have been placed. These changes are all ESLint related.

### How has this been tested?

Tested in CI. Tested manually.

### Checklist:
- [x] My code is tested


AnthonyLedesma:
I've identified the issue with E2E tests and am deploying a fix to this PR shortly.

AnthonyLedesma:
Please review code at the commit: https://github.com/godaddy-wordpress/coblocks/pull/2031/commits/aebbc78e21b081edb9260b870b5bb03a0c9615a9

Beyond this commit I am testing CI functions which are still likely to change.

.dev/performance/config.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+const config = {
+	slug: 'coblocks',
+	name: 'CoBlocks',
+	team: 'GoDaddy WordPress',
+	versionMilestoneFormat: '%(name)s %(major)s.%(minor)s',
+	githubRepositoryOwner: gitRepoOwner,
+	githubRepositoryName: 'coblocks',
+	pluginEntryPoint: 'class-coblocks.php',
+	// buildZipCommand: '/bin/bash bin/build-plugin-zip.sh',
```
olafleur, line 33: is it needed?

.dev/performance/git.js
```diff
@@ -0,0 +1,185 @@
+// @ts-nocheck
```
olafleur, line 1: why the ts-nocheck on a js file?

package.json
```diff
@@ -79,27 +81,39 @@
...
+		"@wordpress/env": "^4.1.0",
+		"@wordpress/eslint-plugin": "9.0.5",
 		"@wordpress/i18n": "3.6.0",
+		"@wordpress/jest-console": "^4.1.0",
+		"@wordpress/jest-preset-default": "^7.1.0",
+		"@wordpress/jest-puppeteer-axe": "^3.1.0",
 		"@wordpress/plugins": "2.6.0",
 		"@wordpress/postcss-themes": "^2.2.0",
-		"@wordpress/scripts": "7.1.4",
+		"@wordpress/scripts": "16.1.0",
```
olafleur, line 103: Yeah :)

src/blocks/gallery-stacked/edit.js
```diff
@@ -18,7 +18,7 @@ import { GalleryClasses } from '../../components/block-gallery/shared';
  * WordPress dependencies
  */
 import { __, sprintf } from '@wordpress/i18n';
-import { Component, Fragment } from '@wordpress/element';
```
olafleur, line 21: I would suggest reverting this change as #2029 already include them, I think.

olafleur (approved):
This is extraordinary! Good job!

Thank you for being so explicit (and even putting the link of the issue) when you put a hack/workaround :)

.dev/config/utils.js
```diff
@@ -0,0 +1,41 @@
+
+/**
+ * External dependencies
+ */
+const { existsSync } = require( 'fs' );
+const path = require( 'path' );
+const { realpathSync } = require( 'fs' );
+const { sync: readPkgUp } = require( 'read-pkg-up' );
```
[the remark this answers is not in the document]
AnthonyLedesma: Sync should be reading package.json files. It looks like the main reason this method is used here is so that we can gain access to the `{ pkg, path: pkgPath }` variables It looks like we basically just see if we are using a config at that point. This is probably a bit complicated for modularity because originally core had a bit of additional functionality here that we do not need at this time.

src/blocks/accordion/edit.js
```diff
@@ -53,22 +52,17 @@ const AccordionEdit = ( props ) => {
...
-		return {
-			getBlockAttributes: blockEditorSelectors.getBlockAttributes,
-			getBlocksByClientId: blockEditorSelectors.getBlocksByClientId,
-			isSelected: isSelected || rootClientId === selectedRootClientId,
-		};
-	} );
+	// Get clientID of the parent block.
+	const rootClientId = getBlockHierarchyRootClientId( clientId );
+	const selectedRootClientId = getBlockHierarchyRootClientId( getSelectedBlockClientId() );
+	const isBlockSelected = isSelected || rootClientId === selectedRootClientId;
```
AnthonyLedesma, line 65: For posterity: This change was needed because the way in which we had used `useSelect` previously, now errors out. This change allows the selector to function as expected.

src/blocks/food-and-drinks/food-item/edit.js
```diff
@@ -89,13 +89,17 @@ const FoodItem = ( props ) => {
 		if ( showImage !== prevShowPrice ) {
 			setAttributes( { url: showImage ? url : '' } );
 		}
+
+		if ( !! url && url !== attributes.url ) {
+			setAttributes( { url } );
+		}
```
AnthonyLedesma, line 95: For posterity: This change was needed because there were situations where the url attribute had been undefined.

src/blocks/gallery-stacked/edit.js
```diff
@@ -157,6 +157,8 @@ class GalleryStackedEdit extends Component {
 			lightbox,
 		} = attributes;
 
+		const hasImages = !! images.length;
+
```
AnthonyLedesma, line 161: For posterity: This change was needed because of a race condition. It seems that this script is now processed in a `strict mode` perhaps because the error is related to the order of instantiation. This is perhaps due to changes in the way in which WebPack now compiles files. This is unsubstantiated but likely given the refactors with `wp-scripts`. This was the only identifiable issue related to order of instantiation.

src/blocks/map/edit.js
```diff
@@ -31,65 +31,60 @@ import {
...
+	constructor() {
+		super( ...arguments );
+		this.state = {
+			apiKey: '',
+			address: this.props.attributes.address,
+			coords: null,
+			hasError: false,
+			isSavedKey: false,
+			keySaved: false,
+		};
```
AnthonyLedesma, line 47: For posterity: This change was needed because the previous refactor to hooks was incompatible with the use of API Keys for Google Maps. Reverting to the class based structure allows the Maps block to function as expected with or without the API Key.

PR #2039, opened 2021-09-23
Test incoming performance changes

AnthonyLedesma:
### Description

Modifications here allow incoming PRs to compare performance differences with the master branch. CI will detect with another branch is incoming and run the tests in CI. So far CI will report on performance and give us a percentage of change between the two branches.

In the future, we can explore a bot posting in the pull request with results as well as making performance results a required step with hard limits to prevent degradation.


olafleur (approved):
Good job! LGTM!

PR #2044, opened 2021-09-28
Fix the resizable block handles for the Logos and Badges block

AnthonyLedesma:
### Description

Logos block resizable handles can overlap the image when the screen is zoomed in. We do a padding calculation on the resize container in order to allow proper spacing between logos block images and we attempt to compensate for that padding by setting explicit `right` CSS. 

Unfortunately, that solution was not complete due to the edge case of browser zoom. In order to compensate for a slight variation on spacing with various zoom settings, this PR alters the resize handles to be slightly apart from the image edge. This allows for a more consistent UX.

I also modify the way that we utilize the ResizableBox to match core standards with `forwardRef`.

### Screenshots

**Old**
image

**New**
image

### Types of changes

JS. CSS.

### How has this been tested?

Tested manually at various zoom settings. 

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/logos/logos.js
+++ b/src/blocks/logos/logos.js
@@ -5,14 +5,14 @@ import classnames from 'classnames';
 import { chunk, flatten } from 'lodash';
 
 /**
- * Internal dependencies
+ * WordPress dependencies
  */
-import { useState, useEffect } from '@wordpress/element';
-import { usePrevious } from '@wordpress/compose';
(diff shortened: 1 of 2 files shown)


olafleur (approved):
Works great until approximately 150% zoom-in for me. At 175% and over, the handles start to overlap and some handles are unusable (but resize is still possible with the other side usually).

But I guess there has to be a limit to what we support ;)

PR #2049, opened 2021-09-29
Remove Carousel block example

snovosel:
### Description

This PR removes the example preview for the gallery carousel block within the menu inserter. 

Due to a core bug here - https://github.com/WordPress/gutenberg/issues/35229 the block preview is not able to be used. We are waiting to hear how the core team will handle the bug before merging this PR.

### Types of changes

Bug Fix

### How has this been tested?

Visually

### Checklist:
- [x] My code is tested


src/blocks/gallery-carousel/index.js
```diff
@@ -52,17 +52,6 @@ const settings = {
...
-		attributes: {
-			gridSize: 'lrg',
-			gutter: 5,
-			images: [
-				{ url: 'https://s.w.org/images/core/5.3/Biologia_Centrali-Americana_-_Cantorchilus_semibadius_1902.jpg' },
-				{ url: 'https://s.w.org/images/core/5.3/Glacial_lakes,_Bhutan.jpg' },
-				{ url: 'https://s.w.org/images/core/5.3/MtBlanc1.jpg' },
-			],
-		},
-	},
```
AnthonyLedesma, line 65: So I was thinking about this and how we are now going and adding examples back where we did not have them. Here we are removing the example and the reason is made clear within this PR but I think the reason initially looking at the code is not clear. For that reason, I would suggest adding an additional comment here specifying the reason that no example is present. Something like:


```
	attributes,
	variations,
	icon: <Icon icon={ icon } />,
	keywords: [...],
	// Diable reason: link to the issue in GB
        // example,
```

AnthonyLedesma (changes requested):

olafleur-godaddy:
Reopening as it appears that in WP 5.9 (at least the nightly current version), not only the Gallery Carousel block preview generate a console error message, but it also make the editor explode.

olafleur-godaddy (approved):

PR #2055, opened 2021-10-04
Click to tweet

jonathanbardo:
### Description
Augment CSS specificity to make sure we override Go styles that are attributed to Quotes and do not apply in this context.

### Screenshots
Before:
Before
After: 
After

### DIFF
--- a/src/blocks/click-to-tweet/styles/style.scss
+++ b/src/blocks/click-to-tweet/styles/style.scss
@@ -2,14 +2,14 @@
 	border: 1px solid transparent;
 	border-radius: 4px;
 	margin-bottom: 1.9em;
-	position: relative;
-	padding-top: var(--coblocks-spacing--3);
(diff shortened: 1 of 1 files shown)


jlemay-godaddy (approved):

jonathanbardo:
This is related to https://github.com/godaddy-wordpress/go/pull/698

AnthonyLedesma (dismissed):

AnthonyLedesma (approved):

AnthonyLedesma (changes requested):
Sorry for all the emails on this one. I was right about to merge when I realized that these styles would break in themes other than Go. We need another solution that is cross theme compatible.

image

jonathanbardo:
Oh good call there. Will fix the problematic section in Go instead.

jonathanbardo:
This is what I'm seeing in twenty twenty one now.
image

PR #2056, opened 2021-10-04
Fix Hero and Shape Divider resize function

AnthonyLedesma:
### Description

Recent PR has introduced a bug because the incorrect version of getBlockAttributes from core had been used. This change alters the selector so that resize functionality works as expected for both blocks.


olafleur (approved):
Works great on my side!

PR #2107, opened 2021-10-20
Migrate shape divider block to a server-rendered block

### LINKED ISSUE #455: SVG errors with non-admins
When using latest WordPress/CoBlocks, etc....

User is a contributor/author level. Using CoBlocks for the page divider tool, which utilizes SVG causes an error.

You can insert the block and customize it, and even save it, but then upon refresh it says it's incompatible and needs to be converted, and it strips the SVG. This is no longer a problem if you upgrade the user to an admin, so I think it has to do with CoBlocks allowing non-admins or people without SVG permissions as a filetype upload to manipulate those blocks, but then not save them.

On the WordPress side, it'd be nice to receive the correct error that it's an author permission issue, not a formatting issue. But that's likely not for you.

jrtashjian:
### Description
Non-admins are able to use the shape-divider block but the output SVG content is stripped from the content on saving/publishing a post. Migrating to a server rendered block prevents this from happening and allows all users to use the block.

fixes #455

### Types of changes
Bug fix

### How has this been tested?
Automated tests

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/shape-divider/test/shape-divider.cypress.js
+++ b/src/blocks/shape-divider/test/shape-divider.cypress.js
@@ -36,8 +36,28 @@ describe( 'Test CoBlocks Shape Divider Block', function() {
 
 		cy.get( '.edit-post-visual-editor .wp-block-coblocks-shape-divider' ).click();
 
-		helpers.setInputValue( 'divider settings', 'shape height', shapeHeight );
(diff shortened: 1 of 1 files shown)


AnthonyLedesma:
I wonder why these perf tests are failing here 😕

Issue #1890, opened 2021-04-21
Option to sort uploaded images on Masonry Gallery


matthiasballreich:
I did not found a way or option in masony gallery block / the other gallery blocks to order the uploaded images. By default the new uploaded / added image will be appenend to end of the gallery, but i want that the latest added images will be appenend to the top / front of the gallery and not at the end.

So maybe you can add such an order option to your really great and nice plugin.

richtabor:
Hey @matthiasballreich, we do have these left/right arrows on each image that allows you to move them around. Are you seeing those when you select the Masonry block? <img width="1283" alt="Screen Shot 2021-04-21 at 12 13 38 PM" src="https://user-images.githubusercontent.com/1813435/115586869-02784880-a29b-11eb-892d-8b23ec36a4b4.png">

matthiasballreich:
@richtabor yeah i know this and i can see this. but this is only helpful for a few images. When you have 50 or 100 images there this i no usable way. so a general option (on the right side) would be cool where you can say order by oldest / newest or append new images at top / at bottom - so that new added images will be automatically appended to the top or the bottom. is it clear what i mean, or should i provide an graphical images describing this?

AnthonyLedesma:
The new Masonry implementation works quite differently than the last Masonry block and now has an option to sort images from within the media library which was previously not supported. Regardless I think that the option to sort images in different ways like mentioned in this feature request would be of value. 

Re-opening this issue in hopes of landing the feature in the future.

Issue #1756, opened 2020-11-15
Masonry/Gallery add support to select image size


despecial:
The masonry block uses way too big images.
Example: column is a maximum of 400px and the photo itself is 1200px wide
My theme has a gallery image size which fits perfectly

A hook or even better an option to select which images are used would be great

AnthonyLedesma:
Thank you for contributing. I am closing this issue and linking back to a primary issue regarding refactoring of the Masonry gallery block where we plan to address this ask.
https://github.com/godaddy-wordpress/coblocks/issues/2078

Issue #1749, opened 2020-11-09
Masonry Block usability 


eurello:
### Is your feature request related to a problem? Please describe.

I should be able to easily know how to add/change images from the Masonry block, but although I'm a very experienced user, I had a really hard time seeing how to help a customer with this today. The customer asked how to swap out the demo images with their own. We were both expecting to see some option in the toolbar above the Block, and/or on the individual images: 

<img width="1167" alt="Screen Shot 2020-11-09 at 12 32 00 PM" src="https://user-images.githubusercontent.com/8698879/98603923-8370f400-22a8-11eb-8d7e-7f67d1d2593d.png">

It wasn't immediately clear that the Upload option _below_ the block was related to this same block: 

<img width="1158" alt="Screen Shot 2020-11-09 at 12 31 53 PM" src="https://user-images.githubusercontent.com/8698879/98603972-94216a00-22a8-11eb-9f70-e3ad3547cac0.png">

Once we realized the way to edit the Block was to add new images and delete the old ones, the customer wasn't sure if doing that would preserve the exact same sizing and tiling look that she liked. 

### Describe the solution you'd like

Personally, I think it'd be more clear (and consistent with other blocks) to have the upload image icon in this toolbar somewhere: 

<img width="1167" alt="Screen Shot 2020-11-09 at 12 32 00 PM" src="https://user-images.githubusercontent.com/8698879/98604345-1ca00a80-22a9-11eb-9d65-81d811c7907b.png">

richtabor:
Thanks for the feedback @eurello, we will be implementing a "Replace" button within each image once selected. This should make replacing specific images much easier. 

<img width="1115" alt="Screen Shot 2021-03-25 at 8 20 19 AM" src="https://user-images.githubusercontent.com/1813435/112471978-f1dbbd80-8d42-11eb-8336-0151fcfbee14.png">

Issue #1719, opened 2020-10-06
ISBAT transform Gallery blocks that have Gutter controls


AnthonyLedesma:
### Describe the bug:

Custom Gutter controls have been introduced for use with CoBlocks gallery blocks. The changes can be found here: https://github.com/godaddy-wordpress/coblocks/pull/1619. The controls have only be implemented within a subset of CoBlocks Gallery blocks which is causing some quirks with transforms. 

Legacy gutter controls and the newer gutter controls component linked in this issue have differing types of props. The schemas of these attributes cannot be easily transfigured which ultimately results in invalid attributes being defined for gallery blocks when transforming.

This issue should be resolved by 
1. Correctly implementing the Gutter controls across all gallery blocks.
 **Blocks that have new gutter controls**
[ ] Masonry
[ ] Stacked
[x] Offset
[x] Collage
[ ] Carousel

2. Re-enable gallery transforms for gutter props.
  * Changes to re-enable can be found here: https://github.com/godaddy-wordpress/coblocks/pull/1718

### To reproduce:

1. Add gallery masonry
2. Add images and transform to Offset.
3. Save the page and reload to see the invalid block error.

### Expected behavior:

Expect transforms to properly transform attributes in a compatible way.

### Isolating the problem:

- [x] This bug happens with no other plugins activated
- [x] This bug happens with a default WordPress theme active
- [x] This bug happens **without** the Gutenberg plugin active
- [x] I can reproduce this bug consistently using the steps above

### WordPress version:

5.5.1

### Gutenberg version:

N/A

AnthonyLedesma:
Tests for Gutter controls should be replaced as part of this PR. This commit here includes the removed tests that should be replaced. 
https://github.com/godaddy-wordpress/coblocks/pull/1718/commits/476a05a813c1b25129691ce1e80465034bc4c80c

PR #2117, opened 2021-10-26
Masonry block use inner core/image blocks

AnthonyLedesma:
### Description

Masonry Block Refactors.

This is moving the Masonry Gallery away from desandro masonry into a pure CSS solution for layout. 

New block architecture reflects that of the Core Gallery block in that this implementation uses Core Image inner blocks. 

Backward compatible to WP 5.5 by keeping the V1 block around. We conditionally load the V1 block to keep functionality as well as preserve masonry styles in the case of pre 5.6 WordPress.

**AC**
This criteria has been modified - Notes below.
- [x] <del>[ISBAT sort images in Masonry Gallery - Something like ascending, descending by upload date, or random, or append to top or bottom.](https://github.com/godaddy-wordpress/coblocks/issues/1890)</del>
  - This is a complicated feature and should be handled in a future iteration.
- [x] [ISBAT have a variety of image size display options available to users](https://github.com/godaddy-wordpress/coblocks/issues/1756) - (400px, 1200px etc)
- [x] <del>[ISBAT access image replace controls both on image and in block toolbar](https://github.com/godaddy-wordpress/coblocks/issues/1749)</del>
  - Image replace and move controls will not exist presently on the Masonry block. This feature is not mirrored in the new Gallery architecture but may come from the core as the v1 Gallery did.
- [x] [ISBAT transform into other blocks that have gutter controls](https://github.com/godaddy-wordpress/coblocks/issues/1719)
- [x] [ISBAT update image captions from within the media library as well as block editor](https://github.com/godaddy-wordpress/coblocks/issues/1620)
- [x] [ISBAT has a mobile view that displays images in a single column](https://github.com/godaddy-wordpress/coblocks/issues/1538)
- [x] [ISN use jQuery in refactored Masonry Gallery](https://github.com/godaddy-wordpress/coblocks/issues/1365)
- [x] <del>[ISB using the newer size controls and deprecated the old controls](https://github.com/godaddy-wordpress/coblocks/issues/1041)</del>
  - Size controls have been deprecated and removed. I think we should consider the fourth masonry column when the block is set to full width. 

### Screenshots

MasonryCSS

### Types of changes

Major changes discarding all of our old CSS and JS in favor of a mirrored core implemntation using hooks. This new method matches the new paradigm and allows us to refactor other galleries to this paradigm more easily.

### How has this been tested?

WIP

### Checklist:
- [ ] My code is tested


jrtashjian:
Can we change the arrows to point up and down instead of left and right? I think it's a prop we can set.

olafleur-godaddy:
When uploading a PNG image, I have this error

Screen Shot 2021-11-16 at 15 19 42

olafleur-godaddy:
When creating a Masonry block in master, it deprecates without error in your branch, but I don't see my images in the Gallery anymore. Is it normal?

src/blocks/gallery-masonry/controls.js
```diff
@@ -1,28 +0,0 @@
-/**
```
olafleur-godaddy, line 1: Why is this removed?

src/blocks/gallery-masonry/inspector.js
```diff
@@ -1,124 +0,0 @@
-/**
```
olafleur-godaddy, line 1: Same question here. It's not needed anymore?

AnthonyLedesma:
> When creating a Masonry block in master, it deprecates without error in your branch, but I don't see my images in the Gallery anymore. Is it normal?

No, the block is not deprecating as expected.

src/blocks/gallery-masonry/controls.js
AnthonyLedesma: Media Filters were applied ad-hoc to the custom inner images that we used in the galleries. We apply the Media Filters by Extension to the core/image block so we already have those controls on the images thus we do not need them here.

olafleur-godaddy: Ok, thanks for the explanation!

src/blocks/gallery-masonry/inspector.js
AnthonyLedesma: These controls have been moved to the edit.js file now. I wonder which might be improved readability either to split the code or to have all of the controls within the edit function.

AnthonyLedesma:
> >    When creating a Masonry block in master, it deprecates without error in your branch, but I don't see my images in the Gallery anymore. Is it normal?

> No, the block is not deprecating as expected.

This should be resolved now with a successful deprecation and conversion to innerblocks.

olafleur-godaddy:
You can probably change the minimal version of WP from 5.5 to 5.6 in the readme, as the v2 block API was introduced in WP 5.6

olafleur-godaddy:
When creating a block in master, saving it, loading it in your branch it works great, but next time I do a save, it breaks the block.

AnthonyLedesma:
> When creating a block in master, saving it, loading it in your branch it works great, but next time I do a save and refresh, it breaks the block.

I think I got it this time. Now it is working with markup generated from master and is persisting past save.

olafleur-godaddy:
Nice! Yes, it's deprecating properly now :)

olafleur-godaddy:
When uploading one or multiple images in the Gallery, I have this message show in the console : 
Screen Shot 2021-11-19 at 11 44 50

Do you also have that?

AnthonyLedesma:
The Jest tests for Masonry deprecations only work properly when using the updated block-editor package.

AnthonyLedesma:
Now that I look at these jest errors it seems like the snapshots just need updating. The markup in the snapshots seems to be produced by upstream dependency updates.

olafleur-godaddy:
Not sure if it's expected behaviour, but when adding multiple images at the same time, the gallery reloads a few times (I think the number of images added) before stabilizing.

olafleur-godaddy (approved):
This is monumental work, good job! :)

olafleur-godaddy:
Whenever you feel ready, you can merge, as the failing tests for OpenTable are not related to your PR :)

jrtashjian (approved):

jrtashjian (approved):
Impressive work!

src/blocks/gallery-masonry/controls.js
```diff
@@ -8,11 +9,10 @@ import MediaFilterControl from '../../components/media-filter-control';
  */
 import { BlockControls } from '@wordpress/block-editor';
 
-const Controls = ( props ) => {
-	const { attributes } = props;
+const Controls = ( props, ) => {
```
jlemay-godaddy, line 12: rm trailing comma

babel.config.json
```diff
@@ -0,0 +1,8 @@
+{
+    "presets": [
+        "@wordpress/babel-preset-default"
+    ],
+    "plugins": [
+        "@babel/plugin-proposal-class-properties"
+    ]
+}
```
jlemay-godaddy, line 8: New Line at EOF

src/blocks/gallery-masonry/edit.js
```diff
@@ -2,268 +2,559 @@
...
+			ALLOWED_MEDIA_TYPES.some(
+				( mediaType ) => file.type?.indexOf( mediaType ) === 0
+			) || file.url?.indexOf( 'blob:' ) === 0
+		);
+	}
+
+	function updateImages( selectedImages ) {
+		const newFileUploads =
+			Object.prototype.toString.call( selectedImages ) ===
+			'[object FileList]';
```
jlemay-godaddy, line 234: First time I see something like that. Doesn't mean it's wrong

src/blocks/gallery-masonry/styles/editor.scss
```diff
@@ -18,118 +17,119 @@
...
+				padding: 0;
+				width: 100%;
 			}
 
-			&:not(.mce-content-body) {
-				opacity: 0.6;
+			figcaption {
+				background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.7) 0, rgba($color: $black, $alpha: 0.3) 50%, transparent);
+				bottom: 0;
+				font-size: 13px;
```
jlemay-godaddy, line 64: We should try to use "em" or "rem" for font-sizes. It will scale better when zooming

src/blocks/gallery-masonry/styles/editor.scss
```diff
@@ -18,118 +17,119 @@
...
-		// IE calculates this incorrectly, so leave it to modern browsers.
-		@supports (position: sticky) {
-			bottom: 4px;
-			left: 4px;
-			margin-top: -4px;
-			right: 4px;
-			width: calc(100% - 8px);
+		/* Masonry on tablets */
+		@media only screen and (min-width: 768px) and (max-width: 1023px) {
+			& {
```
jlemay-godaddy, line 91: I think (I might be wrong) you can remove lines 91 and 93. Same with 98 and 100

src/blocks/gallery-masonry/styles/style.scss
```diff
@@ -1,44 +1,105 @@
-.wp-block-coblocks-gallery-masonry {
-	position: relative;
+// Import styles for rendering the static content of deprecated gallery versions.
+@import "../v1/deprecated.scss";
 
-	ul {
-		list-style: none !important;
-		padding: 0 !important;
+figure.wp-block-coblocks-gallery-masonry.masonry-grid {
+	column-fill: initial;
```
jlemay-godaddy, line 5: I see a lot of code duplication in the editor.scss and style.scss files. Is the backend only loading the editor.scss and the frontend only loading the style.scss? I'm afraid we do a fix on the editor.scss and forget style.scss or something like that. How do we handle that usually?

src/blocks/gallery-masonry/styles/style.scss
```diff
@@ -1,44 +1,105 @@
...
+			margin: 0;
+			max-width: 100%;
+			padding: 0;
+			width: 100%;
+		}
+
+		figcaption {
+			background: linear-gradient(0deg, rgba($color: $black, $alpha: 0.7) 0, rgba($color: $black, $alpha: 0.3) 50%, transparent);
+			bottom: 0;
+			font-size: 13px;
```
jlemay-godaddy, line 36: see editor.scss

src/blocks/gallery-masonry/v1/block.json
```diff
@@ -0,0 +1,10 @@
+{
+	"name": "coblocks/gallery-masonry",
+	"category": "media",
+	"attributes": {
+		"gridSize": {
+			"type": "string",
+			"default": "xlrg"
+		}
+	}
+}
```
jlemay-godaddy, line 10: NL at EOF

jlemay-godaddy (commented):
Good job! Code review done. I'll test it IRL now :)

src/blocks/gallery-masonry/v1/edit.js
```diff
@@ -0,0 +1,272 @@
...
+			[ `coblocks-animate ${ animation }` ]: animation,
+		}
+	);
+
+	return (
+		<>
+			{ isSelected &&
+			<Controls
+				{ ...props }
+			/>
```
jonathanbardo, line 205:
```suggestion
			<Controls { ...props } />
```

src/blocks/gallery-masonry/v1/edit.js
```diff
@@ -0,0 +1,272 @@
...
+			}
+			{ isSelected &&
+			<Inspector
+				{ ...props }
+			/>
```
jonathanbardo, line 210: I don't think we need the extra spacing here.
```suggestion
			<Inspector { ...props } />
```

src/blocks/gallery-masonry/styles/editor.scss
AnthonyLedesma: That sounds good to me. My only hesitation is that a quick scan of the code reveals that we set captions on most galleries to `font-size: 13px` including stacked and offset galleries. These are similar styles in that way. I wonder if we should change them all at once?

src/blocks/gallery-masonry/styles/style.scss
AnthonyLedesma: This was just oversight on my part most likely. Thank you for finding this. The editor.scss should load only in the editor whereas the style.scss should load in both the editor as well as the front end.

src/blocks/gallery-masonry/styles/editor.scss
AnthonyLedesma: Good call. This logic should exist only in the style.scss file and not here. This is duplicated code.

AnthonyLedesma: Oh weird. I just tried removing these and the styles in the editor broke. I wonder if I didn't install the ones under styles.scss correctly. 🤔

AnthonyLedesma: Oh, I got it now. I ended up having some duplicated code because we are targeting differing hierarchies of DOM selectors in the edit and the save markup. Because of this variation, the solution I went for was to include the same styles in both the editor.scss and style.scss file but using the applicable selector. I've updated these files to include a comment about the code duplication and why it exists so that we might improve it in the future.

src/blocks/gallery-masonry/v1/edit.js
AnthonyLedesma: Fixed the spacing 👍

AnthonyLedesma:
Pushed up fixes to address the formatting feedback. Also found out that the stacked jest tests were formatting incorrectly causing the results not to be correct. I fixed that here too.

src/blocks/gallery-masonry/controls.js
AnthonyLedesma: For posterity, I ended up re-adding the media-filter controls the masonry block with some refactors to extension logic. We now have a hierarchy of controls.

Issue #1365, opened 2020-02-17
ISBAT use CoBlocks without jQuery


AnthonyLedesma:
### Is your feature request related to a problem? Please describe.
CoBlocks should consider removing jQuery as a required dependency for the plugin. Getting a look at the plugin I find that 20 files reference `jQuery` which could be refactored to use plain JavaScript. This change would help to make CoBlocks more performant. 

## Required changes
Events & Post Carousel block
- [ ] Remove Slick carousel in favor of Flickity. (Slick is a jQuery extension) - Will be removed following Events/Post-Carousel refactor
- [ ] `src/js/coblocks-events.js` initializes slick in the DOM. - Needs Block refactor
- [ ] `src/js/coblocks-slick-initializer-front.js` - Will be removed following Events/Post-Carousel refactor ( In progress )
- [ ] `src/js/coblocks-slick-initializer.js` - Will be removed following Events/Post-Carousel refactor ( In progress )

Form block - jQuery UI Datepicker | Should be another open-source date picker.
- [x] Remove jQuery from Form block `edit.js` file ( #1376 )
- [ ] `src/js/coblocks-datepicker.js` Should be removed. ( #2180  )
- [ ] `includes/class-coblocks-form.php` Should not reference jquery-ui-datepicker. ( #2180 )
- [x] `src/js/coblocks-google-recaptcha.js` Should not use jQuery. ( #2131 )

Gif block
- [x] Remove jQuery from Gif `edit.js` file ( #1376 )

Gallery Carousel `src/js/coblocks-accordion-carousel.js` - ( PR: https://github.com/godaddy-wordpress/coblocks/pull/2147 )
  - [x] Refactor to vanilla-js swiper package
  
Gallery Masonry `src/js/coblocks-masonry.js`
  - [ ] Use vanilla JS to initialize Masonry. (https://github.com/desandro/masonry#initialize)

Google Maps Script `src/js/coblocks-google-maps.js`
- [x] Refactor script ( #2130 )

CoBlocks Lightbox `src/js/coblocks-lightbox.js`
- [x] Refactor script ( https://github.com/godaddy-wordpress/coblocks/pull/1429 ) 

CoBlocks Crop Settings Extension & Component
- [x] Remove jQuery from Crop Settings extension and component ( #1376 ) 

**Cleanup tasks**
- [ ] Review `includes/class-coblocks-block-assets.php` and remove jQuery from dependency arrays.  
[...]

PR #2130, opened 2021-10-30
Remove jQuery from Google Maps script

olafleur-godaddy:
### Description
One of the tasks in #1365 

### Types of changes
Refactor

### How has this been tested?
By making sure that the script is still called when the Map block is shown on the page.

### Checklist:
- [X] My code is tested


src/js/coblocks-google-maps.js
```diff
@@ -1,1064 +1,1057 @@
...
+	return GMapStyles[ key ];
 };
 
-jQuery( document ).ready( function() {
-	coblocksMap.init();
+// Disable note: This rule is only meant to be used for React components
+// See https://github.com/WordPress/gutenberg/pull/26810
+// eslint-disable-next-line @wordpress/no-global-event-listener
+document.addEventListener( 'DOMContentLoaded', function() {
+	const elements = document.getElementsByClassName( 'wp-block-coblocks-map' );
```
jonathanbardo, line 1054: I guess the const here might be a little too verbose. I think this is readable enough `renderMap( document.getElementsByClassName( 'wp-block-coblocks-map' ) );`

Unless we want to make sure there are elements on the page before calling renderMap.

src/js/coblocks-google-maps.js
```diff
@@ -1,1064 +1,1057 @@
...
-					),
-				},
-			} );
-			map.setCenter( new google.maps.LatLng( gmapAttr.lat, gmapAttr.lng ) );
-		} );
-	},
-
-	mapStyles( key ) {
-		const GMapStyles = {};
+		if (
```
jonathanbardo, line 11: I wonder if it would not be cleaner to also refactor this like
```js
if ( [ gmapAttr.address, gmapAttr.lat, gmapAttr.lng ].includes('undefined') ) { 
    return;
}
```

jonathanbardo (approved):

PR #2134, opened 2021-11-03
Upgrade Husky version

olafleur-godaddy:
### Description
I made some experimentation with the newest version of Husky.
The setup is a little bit different, but the result at the end is the same (pre-commit hook that lint and execute JS tests).

At the start, I began doing it in #2133 but felt it was a change that was a little bit bigger than the other dependencies change and also, may need some discussion : 
1. do we still need this pre-commit hook? I know for myself that I often use `--no-verify` to be quicker (maybe wrong on my part)
2. do we still want to execute the JS tests on pre-commit? Maybe linting is enough?


olafleur-godaddy:
Oh, weird stuff, in the build step, the JS linting is now done on everything instead of just what was changed.

jonathanbardo:
@olafleur I think we are missing the husky install step here no? Also can we move this to the `.dev` directory? It seems possible by https://typicode.github.io/husky/#/?id=custom-directory

olafleur:
The husky install is done automativally with yarn install.

Yeah, good Idea, will put it on .dev

EvanHerman (approved):
These changes are looking good on my end 👍 Tested and working well. Just need to resolve the merge conflicts before we merge it.

PR #2139, opened 2021-11-05
Run all Cypress tests but with more parallelization to make it faster

olafleur-godaddy:
### Description
- Run all the Cypress tests always
- Use more parallelization so that it does not slow the build


EvanHerman (approved):
Looks good to me 👍

PR #2144, opened 2021-11-09
Enable PHPCS cache

EvanHerman:
### Description
This should decrease `phpcs` runs in CI builds.

### Checklist:
- [x] My code is tested


olafleur-godaddy (approved):
Good idea!

jonathanbardo (approved):

PR #2152, opened 2021-11-15
Fix build, deactivate Cypress tests for previous version temporarily, remove hack and prepare Cypress tests for WP 5.9

olafleur-godaddy:
### Description
- Fix build and solidify it by having a bigger screen and waiting a little bit more
- Deactivate Cypress tests for previous version (WP 5.7 and Cypress are really problematic together). I would reactivate it in a few days when 5.8 would be the previous one.
- Remove JS hack that I inserted a few weeks ago
- Start to prepare Cypress tests for WP 5.9


.circleci/config.yml
```diff
@@ -181,7 +181,7 @@ workflows:
             - unit_testing_php_74
             - unit_testing_php_80
             - js
-      - e2e_chrome_wp_previous_major:
```
olafleur-godaddy, line 184: Modifications in this file will be reverted before merging. It's just to be able to test the changes on the nightly release.

jonathanbardo (approved):

PR #2153, opened 2021-11-15
Revert "Issue 1036 - remove all instances of useless classes includin…

olafleur-godaddy:
…g no (#2109)"

This reverts commit ffdd73e8e9200c36f4f8f5b75689a4c296dfe176.

### Description
A deprecation of Row bug was introduced in `2.19.0`

We do not have an easy way to fix it, so we are reverting the PR for the moment.

When creating a Row block in `2.18.0` and updating to this branch, there is no problem anymore.

A user that has created a block before `2.19.0` that updated to `2.19.0` will have the bug, but if he updates to this PR, it solves the problem.

A row block that has been created in `2.19.0` will still cause a problem for the moment.


AnthonyLedesma (approved):
Working on my end for row blocks that were created in 2.18. LGTM.

PR #2154, opened 2021-11-16
Attach artifacts in build

EvanHerman:
### Description
- Attach the built .zip to the artifacts of the `build` job.
- Rename the .zip file to `coblocks-{PR_NUMBER}.zip` when not on master (master is still `coblocks-v{VERSION}.zip`).
- Update the version number in non-master branches to proper semver format (`<version> "-" <pull-request-number> "+" <short-commit-hash>`)
- Setup a bot to comment on the PR with a link to the .zip of the build for the related PR.

### Screenshots
image
image

### Types of changes
Tooling

### How has this been tested?
https://app.circleci.com/pipelines/github/godaddy-wordpress/coblocks?branch=attach-artifacts

### Checklist:
- [x] My code is tested


.circleci/config.yml
```diff
@@ -565,13 +577,85 @@ jobs:
...
       - persist_to_workspace:
           root: ~/project
           paths:
             - build
             - dist
+            - build-num.txt
+
+  pr_artifact_comment:
+    docker:
+      - image: circleci/php:7.3-node
```
jonathanbardo, line 607: @EvanHerman can we bump that to at least 7.4?

EvanHerman: 👍 Good call

jonathanbardo:
@EvanHerman We might be better off using https://cli.github.com/manual/gh_pr_comment instead of the repo you enqueued which is pretty old and relying on ruby. I think it would speed things up not to install Ruby as a step.

EvanHerman:
@jbardo-godaddy I'll give it a try, I was having some issues with that but I'll spend some more time on it. I think you are right.

EvanHerman:
@jbardo-godaddy It looks like there isn't currently a way to edit/delete comments and it's not on their roadmap.
https://github.com/cli/cli/issues/1865#issuecomment-952824640

jonathanbardo:
@eherman-godaddy Yeah I would try with simply curl. Seems easy enough.

EvanHerman:
@jonathanbardo I've removed ghi, ruby etc. This script was the best I could do. I re-used the same script for Go and it seems to be working pretty well, it's about 15 seconds faster this way.

jonathanbardo:
Nice job!

olafleur-godaddy (approved):
Awesome work!

PR #2157, opened 2021-11-17
WPEX 1633 - ISBAT see a preview of all blocks before inserting them onto my page

snovosel:
### Description

This PR adds previews for all CoBlocks blocks that did not previously have a preview set within the side bar inserter that displays previews for each blocks.

### Types of changes

Config change (examples)

### How has this been tested?

Visually

### Checklist:
- [x] My code is tested


olafleur-godaddy:
A few of them, I don't see a preview or only see a preview with the placeholders.
Wouldn't that be more interesting to also add an example with real-ish data to the block that would show, like some other blocks?

olafleur-godaddy (commented):
A few of them, I don't see a preview or only see a preview with the placeholders.
Wouldn't that be more interesting to also add an example with real-ish data to the block that would show, like some other blocks?

snovosel:
@olafleur-godaddy which blocks did not include a preview? Were they core blocks?

olafleur-godaddy:
No, blocks in CoBlocks.

For example, I don't see anything for the Food & Drink block and the Gif block.
There is a popup when hovering, but no visualization in it.

snovosel:
@olafleur-godaddy  I am able to see previews for both blocks currently - is there a version mismatch potentially? 

https://user-images.githubusercontent.com/18115694/143901995-0b275d15-17bd-4d67-b31e-ba21bf5fa5f5.mov

olafleur-godaddy:
What I mean is : instead of having "Menu title" and "Add a title" in the visualization of Food & Drink (and other blocks), wouldn't it be better to have data that looks real like, for example, when you look at the Pricing Table block. Something to inspire the user to see what the block would look like with data in it?

AnthonyLedesma:
On my end, I do not have a preview for Gist block, Form block, or Row blocks. The form and row blocks show variation pickers. The Gist block states to have no preview available.

olafleur-godaddy:
Do you have previews for the Gif, Hero and Food & drink blocks? We both see the Gif one Stephen and I, but I do not see the Hero and Food & Drink on my side.

AnthonyLedesma:
I have previews for Hero, Gist, and Food & Drink but the thing is that these reviews do not seem to be the most reliable. For example, I needed to mouse over and out several times before the Food & Drinks preview was visually displayed.

I think that reliability is actually a core issue, however, but I do not have any existing issue to point to.

olafleur-godaddy (approved):
Works great, especially with WP 5.9 beta 1.

I am comfortable if you want to merge it even if it's not perfect in 5.8, as if the problem is related to core, there's not much we can do to make it better there. Also, the 3 failing tests in OpenTable are not related to your PR.

PR #2161, opened 2021-11-18
Parallelize performance tests

EvanHerman:
### Description
Refactor the performance test script to run performance tests in parallel across to CircleCI jobs.

### Types of changes
Tooling (CircleiCI)

**Current Performance test time:** 28mins
**This PR Performance test time:** 15mins

Example: https://app.circleci.com/pipelines/github/godaddy-wordpress/coblocks/7695/workflows/1089a336-5a63-4313-9730-aff3135f69ce


.circleci/config.yml
```diff
@@ -189,29 +189,37 @@ workflows:
             - unit_testing_php_74
             - unit_testing_php_80
             - js
-      #- e2e_chrome_wp_previous_major:
```
olafleur, line 192: Did you fix the issues with WP 5.7 ?
I disabled this step while there were tests that were problematic.

EvanHerman: @olafleur No, I didn't. I must have accidentally re-enabled that. I'll disable it again.

olafleur: See for example https://app.circleci.com/pipelines/github/godaddy-wordpress/coblocks/7700/workflows/3aa3739e-eaf5-4c44-a4b5-c52ad4d45c7a/jobs/35347
Which is the current status of the 5.7 tests

.circleci/config.yml
```diff
@@ -202,16 +203,23 @@ workflows:
       #      - js
 
       # PERF TESTING
-      - perf_tests:
+      - master_perf_tests:
+          filters:
+            tags:
+              only: /^(?!canary).*$/
+
```
olafleur-godaddy, line 210: shouldn't the perf require previous steps? At least the build?

EvanHerman: I don't think that the performance tests require the build files. The performance tests still run, and if they don't require those files then we can start the performance tests right away to help save some time.
https://app.circleci.com/pipelines/github/godaddy-wordpress/coblocks/7706/workflows/121983b9-74eb-43e1-8174-40a9bf3c8655

olafleur-godaddy: Oh okay, right, I thought the perf files required to have a build done.

olafleur-godaddy:
Looks good to me, but the JS step is failing because of linting issues, I think.

PR #2162, opened 2021-11-18
Performance - Split JS into chunks

jonathanbardo:
### Description
This PR split the coblocks js entry into 8 chunks. This will improve loading as HTTP/2 can make request in parallel and thus further improve load time. 

Lighthouse seems to like this new changes, a clear reduction in JS parsing time. Smaller chunks take less time to parse and therefore can enable the JS engine to move quicker to other part of the application.

Before:
[Before.pdf](https://github.com/godaddy-wordpress/coblocks/files/7571441/Before.pdf)

After:
[After.pdf](https://github.com/godaddy-wordpress/coblocks/files/7571442/After.pdf)


EvanHerman:
Test comment

olafleur-godaddy (approved):
Good job!

PR #2164, opened 2021-11-19
@wordpress/no-global-event-listener disable is not needed anymore

olafleur-godaddy:
### Description
It seems to have been removed as a warning, at least for pure JS files.


EvanHerman (approved):
Looks fine to me 👍

PR #2167, opened 2021-11-19
Fix layout test for WP 5.9

olafleur-godaddy:
### Description
Properly enter into the iframe to make sure it is loaded. Needed to make the test pass in WP 5.9


EvanHerman (approved):
Looks good 👍

PR #2170, opened 2021-11-23
Cleanup helpers.js + remove a warning

olafleur-godaddy:
### Description
This is essentially a PR to remove dead code : 
- In the first if, we remove the condition that was used before WP 5.4 (we now support > 5.5 and test only > 5.7)
- In the getBlockSlug it is only a simplification of the code
- After that, it's just removing unused functions
- The last line modified in index.js it's using `select( 'core/block-editor' )` instead of `select( 'core/editor' )`, which is deprecated (there was a warning message about it).


AnthonyLedesma (approved):
Thank you for these much needed updates.

Issue #2173, opened 2021-11-23
ISBAT select a size that is not Custom in the Click To Tweet block


olafleur-godaddy:
### Describe the bug:
The Text Settings size stays stuck in the Custom choice, not allowing us to return to Normal or Huge or any other of the pre-selected categories. Also, when trying to do so, it resets the custom size to 1px (which is really tiny!)

### Expected behavior:
Being able to navigate between Custom and non-Custom categories without a problem. See other blocks that are working correctly like 

### Isolating the problem:

- [X] This bug happens with no other plugins activated
- [X] This bug happens with a default WordPress theme active
- [X] This bug happens **without** the Gutenberg plugin active
- [X] I can reproduce this bug consistently using the steps above

### WordPress version:
5.9-alpha

PR #2183, opened 2021-11-29
Fix CoBlocksFontSizePicker component

EvanHerman:
Resolves https://github.com/godaddy-wordpress/coblocks/issues/2173

### Description
This PR fixes the `CoBlocksFontSizePicker` component not updating the dropdown value to the correct selected value. This PR also allows users to switch back to `default`.

### Screenshots
image

image

### Types of changes

### How has this been tested?
Manually tested each block using the `CoBlocksFontSizePicker` component.

### Checklist:
- [x] My code is tested


olafleur-godaddy:
It seems to break the font size settings of the Highlight block, though :(

EvanHerman:
@olafleur-godaddy You are right, I didn't see that. I'll take another look.

olafleur-godaddy (changes requested):
The added Click To Tweet Cypress test does not seem to work on WP 5.9 beta 2

AnthonyLedesma (changes requested):
There is a bit of strange behavior here. 

When setting the font size to `large` the FontSize controls instead set to `custom` with 26px value. This should show `large` instead.

When setting the font size to `custom` the label instead changes to `default` with no custom value set. Ideally, we should set a default custom value that allows the label to specify `custom` and still have a sample value for a user to modify.
FontSizeBugged

snovosel:
@AnthonyLedesma I wasn't able to reproduce that functionality above regarding not being able to select named options:

https://user-images.githubusercontent.com/18115694/145290879-fbfb869a-3d5f-4206-bbc8-59a1a7429ad9.mov

However, I am able to reproduce the second comment as shown in the video. I've determined that the cause of this is due to the `<FontSizePicker />` core component which provides an `onChange` callback prop with a single argument. In both scenarios , the "Default" option and "Custom" option will pass `undefined` . On the other hand, the `<FontSizePicker />` component takes in a value which if `undefined` will display the "Default" option. I believe this is a bug in core and so can submit an issue .

Otherwise, the UX for this component has been updated in 5.9 and this is no longer an issue with the updated version.

AnthonyLedesma:
Sounds good thanks for checking. I had a thought that It could be because we have different themes activated but I have not tested that yet.

jonathanbardo:
@EvanHerman won't this PR basically remove the class name added when choosing from the Dropdown? eg `has-small-font`

olafleur-godaddy (approved):
Oh, the tests do not seem to be properly filtered on WP 5.8 and WP 5.9

olafleur-godaddy (approved):

AnthonyLedesma (approved):
The issue where the label does not show still exists. As expected this only affects core themes where TwentyTewnety is affected and possibly others. The case of a missing label is not a big deal for me given that the pixel value is shown regardless. No issues with deprecations were detected. LGTM.

PR #2185, opened 2021-12-02
Disable perf tests temporarily

olafleur-godaddy:
### Description
For the moment, these tests are broken and don't provide value, while breaking our build.


AnthonyLedesma (approved):
LGTM

PR #2186, opened 2021-12-03
Rewrite the Map Block to hooks and modern libraries

### LINKED ISSUE #2033: ISBAT use the Map block with React hooks
### Describe the bug:

Recent [refactor of Map block to React hooks](https://github.com/godaddy-wordpress/coblocks/pull/2010) did not cover the case of API key use. Standard block behavior is to use a simple Iframe and source the Google API. The API key method is to use a 3rd party package `react-google-maps` for its Google Maps Implementation. The way in which the component functions is not compatible with React Hooks and only functions in this context within class-based components. Part of the issue has to do with the fact that `react-google-maps` was written using a now legacy version of React and has features that are now deprecated, specifically context.

With the [Performance Metrics PR, the Map block has been reverted back to a Class-based component](https://github.com/godaddy-wordpress/coblocks/pull/2031/commits/aebbc78e21b081edb9260b870b5bb03a0c9615a9). We must investigate what it will take to refactor the Google Maps API key functionality into a state where we have functionality with hooks. I posit that perhaps we should consider looking for an alternate (more up-to-date) or write our own solution to be housed within CoBlocks codebase. 

- [The block should be able to deprecate from the following block markup](https://github.com/godaddy-wordpress/coblocks/issues/1596)
```
<!-- wp:coblocks/map {"address":"Scottsdale Arizona","pinned":true,"height":534,"align":"full"} -->
<div style="min-height:534px" data-map-attr="/qaddress/q:/scottsdale arizona/q||/qlat/q:/qundefined/q||/qlng/q:/qundefined/q||/qskin/q:/qstandard/q||/qzoom/q:/q12/q||/qiconSize/q:/q36/q||/qmapTypeControl/q:/qtrue/q||/qzoomControl/q:/qtrue/q||/qstreetViewControl/q:/qtrue/q||/qfullscreenControl/q:/qtrue/q" class="wp-block-coblocks-map alignfull"><iframe title="Google Map" frameborder="0" style="width:100%;min-height:534px" src="https://www.google.com/maps?q=scottsdale%20arizona&amp;language=ja&amp;output=embed&amp;hl=%s&amp;z=12"></iframe></div>
<!-- /wp:coblocks/map -->
```

- [IS not use jQuery in the Map block refactor](https://github.com/godaddy-wordpress/coblocks/issues/1365)
- [The Map block is affected by specific conditional user permissions that can affect UX. This is a bit of an outstanding issue that also affects the Gist block but it should be resolved for the Map block here.](https://github.com/godaddy-wordpress/coblocks/issues/175) - We might be able to resolve this issuse by using an `oembed` as suggested in the linked thread. We might also need to go more of a SSR route but I am not positve either way. That research is part of the refactors.

olafleur-godaddy:
Fixes #2033 

Todo : 
- [x] No deprecation needed
- [x] Make sure that the map load properly with an API key
- [x] Make sure that all the properties are reflected in the Map
- [x] Add test for the components usable with the API key


AnthonyLedesma (approved):
Pushed up a change to explicit return null in one place instead of fragments. Updated example to Tempe Arizona as the GoDaddy headquarters is no longer in Scottsdale.

src/blocks/map/edit-gmaps-iframe-render.js
```diff
@@ -0,0 +1,27 @@
+import { __ } from '@wordpress/i18n';
+
+const locale = document.documentElement.lang;
+
+const GoogleMapIframeRender = ( props ) => {
+	const { attributes } = props;
+
+	const { address, height, zoom } = attributes;
+
+	return <>
```
jonathanbardo, line 10: @olafleur-godaddy I would wrap this in parenthesis.

src/blocks/map/edit-gmaps-with-api-key.js
```diff
@@ -0,0 +1,115 @@
...
+		skin,
+		streetViewControl,
+		zoom,
+		zoomControl,
+	} = attributes;
+
+	const [ coords, setCoords ] = useState( null );
+
+	useEffect( () => {
+		if ( isLoaded ) {
```
jonathanbardo, line 41: Can we reverse the condition here? `if ( ! isLoaded ) return` as to prevent identation level.

src/blocks/map/edit-gmaps-with-api-key.js
```diff
@@ -0,0 +1,115 @@
...
+			const geocoder = new window.google.maps.Geocoder();
+			geocoder.geocode(
+				{ address },
+				function( results, status ) {
```
jonathanbardo, line 45: I think you can use an arrow function here instead. Would remove the `.bind(this)`

src/blocks/map/edit-gmaps-with-api-key.js
olafleur-godaddy: Good idea!

src/blocks/map/edit-gmaps-with-api-key.js
olafleur-godaddy: Oh yeah, would be a lot cleaner, agreed ;)

jonathanbardo (changes requested):
LGTM just a couple of styling changes

PR #2187, opened 2021-12-03
Fix OpenTable Cypress tests by mocking network call

olafleur-godaddy:
### Description
This PR fixes and solidifies the OpenTable Cypress tests by mocking the API call. 
As @AnthonyLedesma said : 
"We are not testing the availability of the API but rather that the block reacts as expected with typical API response"

### Types of changes
Tests improvement

### Checklist:
- [X] My code is tested


olafleur-godaddy:
Putting this PR on hold for the moment, as there seems to be a real problem with the block.

Issue #2073, opened 2021-10-12
ISBAT use performant CoBlocks extensions


AnthonyLedesma:
### Describe the bug:

This screenshot is a run on CoBlocks performance metrics where all CoBlocks extensions have been disabled. 
![image](https://user-images.githubusercontent.com/30462574/136973344-5bd02aec-325d-4177-ab76-ece5391fb8b5.png)
Disabling all CoBlocks extensions we see a 70% reduction in load time as well as performance improvements on virtually every other metric as well. 

We call BlockEdit within 11 files and BlockListBlock within 7 files each call causing subsequent re-renders. Can we batch these calls? What can we memo? What can we deprecate and remove? 
We need to find a way to optimize these extensions.

AnthonyLedesma:
POC to improve CoBlocks extensions performance.
https://github.com/godaddy-wordpress/coblocks/pull/2189

PR #2189, opened 2021-12-06
Batch use of filters across custom extensions and components

AnthonyLedesma:
### Description

We utilize filters for each CoBlocks extension as well as High Order Components that result in excessive renders. I propose that by batching our extensions we can reduce excessive calls and improve performance across the editor.

### Types of changes

Changes how CoBlocks extensions are loaded. Refactors to extensions to work properly in this way.

### How has this been tested?

E2E, Manually.

### Checklist:
- [x] My code is tested


src/extensions.js
```diff
@@ -5,6 +5,7 @@ import './utils/block-category';
 import './js/coblocks-color-fix';
 
 // Extensions
+import './extensions/apply-extensions';
```
AnthonyLedesma, line 8: The other extensions import styles directly here in some cases. To prevent the import now would beak existing styles.

AnthonyLedesma: I've identified in another commit where this is an issue while fixing its side-effects.

https://github.com/godaddy-wordpress/coblocks/pull/2196/files#diff-715fee81e18461db0b1a0f37d048f8277f43cb3b568654b48c32d31444ac8687R1-R7

src/components/info-popover/index.js
```diff
@@ -1,14 +1,12 @@
...
  */
+import { info } from '@wordpress/icons';
 import { useState } from '@wordpress/element';
 import {
 	Button,
 	Popover,
 } from '@wordpress/components';
-import { info } from '@wordpress/icons';
-
-import './styles/editor.scss';
```
AnthonyLedesma, line 11: Styles should not be imported directly.

src/components/swiper/index.js
```diff
@@ -5,11 +5,6 @@ import { v4 as generateUuid } from 'uuid';
 import TinySwiper from 'tiny-swiper';
 import TinySwiperPluginNavigation from 'tiny-swiper/lib/modules/navigation.min.js';
 
-/**
- * Internal dependencies
- */
-import './style.scss';
```
AnthonyLedesma, line 11: Styles should not be imported directly.

src/extensions/coblocks-settings/index.js
```diff
@@ -1,19 +1,13 @@
 /*global coblocksSettings*/
-
-/**
- * Styles
- */
-import './styles/style.scss';
```
AnthonyLedesma, line 6: Styles should not be imported directly.

src/blocks/accordion/edit.js
```diff
@@ -71,44 +70,42 @@ const AccordionEdit = ( props ) => {
...
+
+			if ( lastBlockClient.customTextColor ) {
+				copyAttributes = Object.assign( copyAttributes, {
+					customTextColor: lastBlockClient.customTextColor,
+				} );
+			}
 		}
+
+		const created = createBlock( 'coblocks/accordion-item', copyAttributes );
+		insertBlock( created, undefined, clientId );
```
AnthonyLedesma, line 104: Accordion block tests were broken in previous commits. It looks like it has to do with the fact that `items[ 0 ].innerBlocks` were being interpreted as truthy under any circumstance ( [] == true | [...] == true). This was an issue when on line 73 we checked for the block length without using the `?.` optional selector.

Fixes here allow blocks to be properly inserted when this function fires.

src/blocks/shape-divider/deprecated.js
```diff
@@ -199,7 +200,7 @@ const deprecated = [
 			...metadata.attributes,
 			shapeHeight: { type: 'number', default: 100 },
 			backgroundHeight: { type: 'number', default: 50 },
-			[ attr ]: false,
+			[ attr ]: { type: 'boolean', default: false },
```
AnthonyLedesma, line 203: Deprecation Jest tests were failing on this attribute because the block parser was choking while checking for the `default` property. Now that a definition is an actual object, the parser works as expected.

AnthonyLedesma:
This story was originally intended to handle extensions specifically however after reviewing how we use some of our custom components it is clear that we are miss-categorizing software because we have components extending using filters where they should likely be classified as an extension. I've made the effort to merge all components and extensions using unified filters.

On future iterations It would be wise to consider re-organizing our components and extensions.

src/extensions.js
AnthonyLedesma: Since I have taken time to remove the excess style imports I feel comfortable removing the superfluous imports now.

src/blocks/accordion/edit.js
```diff
@@ -71,44 +70,42 @@ const AccordionEdit = ( props ) => {
...
-					copyAttributes = Object.assign( copyAttributes, {
-						customTextColor: lastBlockClient.customTextColor,
-					} );
-				}
+		const lastId = items[ 0 ].innerBlocks[ items[ 0 ].innerBlocks.length - 1 ]?.clientId;
+		let copyAttributes = {};
+
+		if ( !! lastId ) {
+			const lastBlockClient = getBlockAttributes( lastId );
+			if ( lastBlockClient.backgroundColor ) {
```
jonathanbardo, line 78: @aledesma-godaddy I know this is only a copy paste but I feel like we can utilize a loop here instead of multiple if statement. `['backgroundColor','textColor'].forEach( prop => { if...etc } )`

src/extensions/animation/index.js
```diff
@@ -13,60 +13,51 @@ import classnames from 'classnames';
...
-addFilter(
-	'editor.BlockEdit',
-	'coblocks/animations',
-	withControls
-);
+const useAnimationControls = ( ( props ) => {
+	const block = select( 'core/block-editor' ).getBlock( props.clientId );
+
+	return (
+		<>
```
jonathanbardo, line 47: @aledesma-godaddy can you check if the fragment is even required here?

src/extensions/animation/index.js
```diff
@@ -108,89 +93,73 @@ function applyAnimationSettings( extraProps, blockType, attributes ) {
...
+	const [ isAnimating, setIsAnimating ] = useState( false );
 
-			return <BlockListBlock { ...props } />;
+	const blockIsAllowed = allowedBlocks.some( ( allowedBlock ) => allowedBlock.blockType === blockName && ! allowedBlock.animateChildren );
+
+	useEffect( () => {
+		if ( !! blockIsAllowed && !! animation ) {
+			prevAnimation.current = animation;
 		}
+	}, [ animation ] );
```
jonathanbardo, line 120: You don't want useEffect to rerun if blockIsAllowed is changed? If you do you need it in the passed props.

src/extensions/animation/index.js
```diff
@@ -108,89 +93,73 @@ function applyAnimationSettings( extraProps, blockType, attributes ) {
...
+	}
 
-		// Apply animation classes to block wrapper only if the animation attribute has changed.
-		const prevAnimation = useRef();
-		useEffect( () => {
-			prevAnimation.current = animation;
-		}, [ animation ] );
+	if ( ! animation ) {
+		// If animation is unset ensure that `coblocks-animate` class is unset.
+		if ( props.attributes?.className?.includes( 'coblocks-animate' ) ) {
```
jonathanbardo, line 129: I wonder if we can use classnames utility to put a class to false. Avoiding the verry verbose code below.

src/extensions/coblocks-settings/index.js
jonathanbardo: @jrtashjian can you double check? I think there was a specific reason for that.

src/extensions/colors/inspector.js
```diff
@@ -62,29 +63,16 @@ function addAttributes( settings ) {
...
-				<BlockEdit { ...props } />
-				{ props.isSelected && allowedBlocks.includes( props.name ) && <Inspector { ... { ...props } } /> }
-			</Fragment>
-		);
-	};
-}, 'withInspectorControl' );
+const useColorControls = ( props ) => {
+	const allowedBlocks = [ 'core/list', 'core/quote' ];
+	return (
+		<>
```
jonathanbardo, line 71: I don't think the fragment is needed here too.

src/extensions/image-crop/index.js
```diff
@@ -14,183 +14,169 @@ import CropControl from '../../components/crop-settings/crop-control';
...
+		positioningControlData.updateDebounce( updateImage );
 	};
-},
-'positioningControl'
-);
 
-addFilter( 'blocks.registerBlockType', 'coblocks/imagePositioning/attributes', addPositioningControl );
-addFilter( 'editor.BlockEdit', 'coblocks/positioning', positioningControl );
+	return (
+		<>
```
jonathanbardo, line 157: Fragment is not needed here.

src/extensions/image-filter/index.js
```diff
@@ -22,67 +20,59 @@ const allowedBlocks = [
...
-}, 'withGalleryExtension' );
+const useImageFilter = ( props ) => {
+	const { name } = props;
+	if ( ! allowedBlocks.includes( name ) ) {
+		return;
+	}
 
-addFilter( 'editor.BlockEdit', 'coblocks/coreImageFilter', coreImageFilter );
+	return (
+		<>
```
jonathanbardo, line 33: Fragment not needed here either.

src/extensions/lightbox-controls/index.js
```diff
@@ -20,85 +19,93 @@ const blocksWithLightboxSupport = [ 'core/gallery', 'core/image' ];
...
-		);
+	const getLightboxHelp = ( checked ) => {
+		return checked
+			? __( 'Image lightbox is enabled.', 'coblocks' )
+			: __( 'Toggle to enable the image lightbox.', 'coblocks' );
 	};
-}, 'withLightbox' );
+
+	return (
+		<>
```
jonathanbardo, line 46: I think we can avoid the fragment here as well.

src/extensions/coblocks-settings/index.js
AnthonyLedesma: If that is the case, this was duplicated already here:
https://github.com/godaddy-wordpress/coblocks/pull/2189/files#diff-0baec429b7317cde7bf9145f886b8402d07d4143756b8288e98c7958ad295f78R26

src/extensions/animation/index.js
AnthonyLedesma: The `blockIsAllowed` variable is based on the constant set on line 19. This constant does not receive mutations and therefore we do not expect the value to change after the initial computation of its value.

src/extensions/animation/index.js
AnthonyLedesma: Things got a little confusing here but I think the primary issue is that the `coblocks-animate` class, after extended, would reside under `attributes.className` as opposed to being regularly dynamically created on the fly. So it was kinda added a single time and therefore does not remove itself from `attributes.className`. This complicated logic was used to force removal of the `blocks-animate` class even when it was part of the `attributes.className` string.

That being said there were a lot of moving parts with this effort. It might be worth revising the logic in a separate PR. 

I think part of the issue with revisiting the code here to re-write is that we are likely not addressing any serious performance issue but rather code quality. While code quality of existing software is important it is not relevant to these changes.

src/extensions/animation/index.js
AnthonyLedesma: Tested this again. Without the JSX fragments, I am receiving parsing errors with unexpected tokens. I think it might be because it's a conditional wrap meant to be within JSX as opposed to a return null early or something like that.

src/extensions/colors/inspector.js
AnthonyLedesma: Unfortunately same issue with wrapped JSX conditional I am receiving an unexpected token error. I suspect the better way to handle this would be to return null early instead of the conditional wrap method.

AnthonyLedesma: Oh, I like the tertiary idea. I'll get this pushed up with a tertiary instead it looks nice.

src/blocks/accordion/edit.js
AnthonyLedesma: Love the idea. I went ahead and made that change to the edit.js logic there.

jrtashjian (changes requested):
So I’m having trouble understanding why this is the correct solution to the [linked issue](https://github.com/godaddy-wordpress/coblocks/issues/2073). The chart indicating performance gains if we were to completely remove the extension’s code is very interesting but what’s the root cause for the re-renders and are we sure the components are re-rendering?

Higher Order Components are a very standard way of adding functionality without directly putting the wrapper in each component that uses it, and [WordPress filters](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/higher-order/with-filters/index.js) just make that a lot easier in the “WordPress way” that is familiar. Core uses both all over the place:

- https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/colors/with-colors.js
- https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/higher-order/with-constrained-tabbing/index.js
- https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/hooks/anchor.js
- https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/hooks/align.js

I would suggest running a performance test locally on this solution and on the current master branch to see the difference (make sure to run both locally to have the same baseline). It would also be important to use the React Tools Profiler to see if we can identify the specific reason for supposed re-renders. Are state updates changing props in the tree forcing re-renders? Are we doing more than we should be in each filter? Are filters that shouldn't be processed still being processed because our conditions aren't effective?

There is nothing I’m seeing wrong with the code you wrote I just strongly believe we should be able to see why this is the correct solution or if we were just doing something incorrect in the original code and need to adjust.

AnthonyLedesma:
In terms of what this story aims to achieve are outlined about the questions written in the issue “Can we batch these calls? What can we memo? What can we deprecate and remove?”
We now know that batching of our extension calls is possible which allows for several benefits including readability and conformity of code.

Your callout of whether reduction of nesting HighOrderComponets from a depth of 16 to a depth 1 is an interesting thought that I hadn’t considered frankly. Similarly, I presumed that calling filters a similar number of times instead of once would also come with implicit performance drops and I presumed that removal of the repeated code would help as well. 

Locally we do not yet had a method to create median and compare point meaning there is a large deal of tedious menial calculations that are required. Likely I will instead wait for https://github.com/godaddy-wordpress/coblocks/pull/2192 which should resolve the perf test issues in CI.

jrtashjian:
> In terms of what this story aims to achieve are outlined about the questions written in the issue “Can we batch these calls? What can we memo? What can we deprecate and remove?” We now know that batching of our extension calls is possible which allows for several benefits including readability and conformity of code.

This is where my question lies. What information do we have that got us from "commenting this code out saw performance gains" to "make less calls by consolidating"? I'm not opposed to merging this if it is the right solution but there isn't any information or data telling me this is. It just seems like we jumped to a conclusion here. Additionally if this way we're using filters is causing problems, having information we can present to Core and get feedback or expose a problem that can be fixed in Core is invaluable.

> Your callout of whether reduction of nesting HighOrderComponets from a depth of 16 to a depth 1 would improve performance is an interesting thought that I hadn’t considered frankly. Similarly, I presumed that calling filters a similar number of times instead of once would also come with implicit performance drops and I presumed that removal of the repeated code would help as well.

I think this is ok to presume and I would presume the same. We should still prove our our presumptions so we are solving the correct problem. React is pretty smart with how it updates the DOM tree when components update and knowing that my mind went immediately to how props and state is changing (which is how React decides what to re-render).

> Locally we do not yet had a method to create median and compare point meaning there is a large deal of tedious menial calculations that are required. Likely I will instead wait for #2192 which should resolve the perf test issues in CI.

That's fine. I only mentioned the performance test because that's the only piece of data that was posted in the issue and so it would be the minimal piece of information to compare against. I would put more weight on the React Dev Tools profiler and see if you can identify how each filter is performing and compare a profile in master vs this branch.

AnthonyLedesma:
> This is where my question lies. What information do we have that got us from "commenting this code out saw performance gains" to "make less calls by consolidating"? I'm not opposed to merging this if it is the right solution but there isn't any information or data telling me this is. It just seems like we jumped to a conclusion here.

I have to disagree that there were any jumps to conclusions here. The performance test chart located within the linked issue shows us a few things. From the data, we can glean that a significant percentage of what makes up the processing load-time for CoBlocks is directly related to Extensions (~70%). What this data does not show us is precisely why that is the case. What we also know is that this data is hypothetical and unrelated to a realistic goal of performance gain. The charts do not tell us what type of performance gains to achieve or anything of that nature but rather that a small subset of all of our code, read extensions and components make up a majority of the load. I think there are several questions that you are asking which I think are critical and we should explore those questions.


> Additionally if this way we're using filters is causing problems, having information we can present to Core and get feedback or expose a problem that can be fixed in Core is invaluable.

I agree with that sentiment I love the idea of component smashing and minimizing the footprint of superfluous code.

jrtashjian:
> What this data does not show us is precisely why that is the case.

IMO this is the most important question. I can do some exploration around this as well to assist in answering it. If we can document our learnings around this it'll be easier to apply the newfound knowledge across our codebase, creating a more stable and performant product.

jrtashjian:
I was able to run the performance tests using the Gutenberg repo and wp-env so we can get the baseline from WP 5.8, the latest release of CoBlocks (2.19.1), and this branch. Here are the results:

|                                                                   | WP 5.8 | CoBlocks 2.19.1 |        | CoBlocks try/merge-extensions |        |
| :---------------------------------------------------------------- | -----: | --------------: | :----- | ----------------------------: | :----- |
| Average time to load                                              | 5114ms |         17227ms | 236.9% |                       15641ms | -9.2%  |
|                                                                   |        |                 |        |                               |        |
| Average time to type character                                    |   31ms |            45ms | 46.0%  |                          34ms | -24.7% |
| Slowest time to type character                                    |  273ms |           287ms | 4.9%   |                         251ms | -12.5% |
| Fastest time to type character                                    |   26ms |            40ms | 52.6%  |                          30ms | -24.0% |
|                                                                   |        |                 |        |                               |        |
| Average time to select a block                                    |   68ms |           109ms | 60.8%  |                          84ms | -23.4% |
| Slowest time to select a block                                    |  240ms |           296ms | 23.4%  |                         244ms | -17.8% |
| Fastest time to select a block                                    |   39ms |            64ms | 63.8%  |                          46ms | -28.8% |
|                                                                   |        |                 |        |                               |        |
| Average time to open global inserter                              |  535ms |           590ms | 10.1%  |                         505ms | -14.3% |
| Slowest time to open global inserter                              |  900ms |           960ms | 6.6%   |                         919ms | -4.3%  |
| Fastest time to open global inserter                              |  465ms |           533ms | 14.4%  |                         429ms | -19.5% |
|                                                                   |        |                 |        |                               |        |
| Average time to move mouse between two block item in the inserter |   19ms |            17ms | -9.8%  |                          17ms | 0.4%   |
| Slowest time to move mouse between two block item in the inserter |   40ms |            26ms | -34.6% |                          24ms | -9.6%  |
| Fastest time to move mouse between two block item in the inserter |   14ms |            15ms | 5.4%   |                          13ms | -12.8% |

We are definitely seeing a performance improvement here.

AnthonyLedesma:
Thank you for doing that labor 😫 Now i am looking at these CI perf tests. Hoping to resolve that issue.

src/extensions/coblocks-settings/index.js
jrtashjian: yeah, it doesn't matter anymore with the new wp-scripts setup. Any file named `index.scss` will be merged into the editor stylesheet, and `style.scss` will be merged into the general stylesheet.

jrtashjian (approved):

jrtashjian:
I'm approving this so we can merge it. It is obviously a net benefit base performance wise. I'd still like us to figure out the why but I don't want that to keep us from pushing it forward.

Issue #2148, opened 2021-11-10
ISBAT ...


Budsy:
### Describe the bug:

Forms emails sent by Wordpress are often detected as spam, filtered into spam or garbage folders in email clients.

### To reproduce:

Fill out a Form (such as an Events form) with Email address with a domain different than the site domain. and submit. Check email receipt in email client. Very often it will be flagged as spam. Tested with Sinatra theme.

### Expected behavior:

Expect email messages to use Wordpress general settings admin email as the email message From: field, not the contents of the Email text box supplied by the site visitor. Rather would have that email address be used in the ReplyTo: field instead.
### Isolating the problem:

I think the spam detection behaviors observed in clients is because the email message uses the input from the form Email text box (the user's email address, not the default admin address) to populate the **From:** address, while it would be best for the forms to send emails with the user email in the **ReplyTo** field, and use the WordPress general admin email address for the **From:** address field. This would be better, since some hosts (SmarterASP.NET) won't serve emails sent with From: set to an address not present in their own email servers. Also, this rearrangement might/should help to avoid spam detection.

- [ ] This bug happens with no other plugins activated
- [ ] This bug happens with a default WordPress theme active
- [ ] This bug happens **without** the Gutenberg plugin active
- [ ] I can reproduce this bug consistently using the steps above

### WordPress version:

Wordpress version 5.8.1

### Gutenberg version:

Not known. (Sinatra theme v 1.2.1 used)

PR #2191, opened 2021-12-07
Remove the From header in the form block

### LINKED ISSUE #1969: ISBAT not have the FROM field of the form block automatically changed to the submitter email to prevent deliverability issues
### Is your feature request related to a problem? Please describe.

I should be able to change the behaviour of rewriting the "FROM" header or the form should not set the from header to the submitted user's email address.

The issue with forcing the from header to be the submitter email is that it can caus deliverability issues. If a user from xyz@gmail.com tries to submit a form and the form sends this to abc@gmail.com this would likely get flagged by GMail among many other mail providers as spam since it recognises that the form was not sent by a GMail server even though the from address indicates it is.

The only solution to this is to make changes to the mail sending on the server manually, with custom code or with another plugin.

### Describe the solution you'd like

The above setup may be desirable in some situations but ideally shouldn't be the default behaviour. To improve deliverability out-of-the-box it would make sense to keep the from email as the default provided in the `wp_mail()` function since this is likely to be managed by the site, server or a plugin already.

### Describe alternatives you've considered

An alternative might be to set the reply-to field only so that the email still comes from the default site email address but then replies will be sent to the submitter email address.

Another option is to make the from field configurable somehow.

The only other option is to override this behaviour with a plugin that sets the from field _after_ CoBlocks changes this field.

### Additional context/screenshots

Markup on 2021-07-28 at 08:27:52

Markup on 2021-07-28 at 08:27:08

EvanHerman:
Resolves #1969 and #2148

### Description
Remove the "From" header when the form block sends an email address. This is to help mitigate the issue of email providers marking the emails as spam. In this solution we are leaving all default `wp_mail()` headers, except we are setting a custom `Reply-To` header.

#### Email + Name
image

#### Email + No Name
image

#### No Email + Name
image

### Types of changes

### How has this been tested?
Manually tested.

### Checklist:
- [x] My code is tested

### DIFF
--- a/includes/class-coblocks-form.php
+++ b/includes/class-coblocks-form.php
@@ -961,7 +961,8 @@ class CoBlocks_Form {
 		 */
 		$email_content = (string) apply_filters( 'coblocks_form_email_content', $this->email_content, $_POST, $post_id );
 
-		$reply_to = isset( $_POST[ $email_field_id ]['value'] ) ? esc_html( $_POST[ $email_field_id ]['value'] ) : esc_html( get_bloginfo( 'admin_email' ) );
+		$sender_name  = isset( $_POST[ $name_field_id ]['value'] ) ? esc_html( $_POST[ $name_field_id ]['value'] ) : '';
+		$sender_email = isset( $_POST[ $email_field_id ]['value'] ) ? esc_html( $_POST[ $email_field_id ]['value'] ) : '';
 
 		/**
 		 * Filter the form email headers.
@@ -973,7 +974,7 @@ class CoBlocks_Form {
 		$email_headers = (array) apply_filters(
 			'coblocks_form_email_headers',
 			array(
-				"Reply-To: {$reply_to}",
+				"Reply-To: {$sender_name} <{$sender_email}>",
 			),
 			$_POST,
 			$post_id
--- a/includes/class-coblocks-form.php
+++ b/includes/class-coblocks-form.php
@@ -964,7 +964,7 @@ class CoBlocks_Form {
 		$sender_email = isset( $_POST[ $email_field_id ]['value'] ) ? esc_html( $_POST[ $email_field_id ]['value'] ) : '';
 		$sender_name  = isset( $_POST[ $name_field_id ]['value'] ) ? esc_html( $_POST[ $name_field_id ]['value'] ) : '';
 
-		$headers = [];
+		$headers = array();
 
 		if ( ! empty( $sender_email ) ) {
 
--- a/includes/class-coblocks-form.php
+++ b/includes/class-coblocks-form.php
@@ -961,8 +961,7 @@ class CoBlocks_Form {
 		 */
 		$email_content = (string) apply_filters( 'coblocks_form_email_content', $this->email_content, $_POST, $post_id );
 
-		$reply_to  = isset( $_POST[ $email_field_id ]['value'] ) ? esc_html( $_POST[ $email_field_id ]['value'] ) : esc_html( get_bloginfo( 'admin_email' ) );
-		$from_name = ( isset( $_POST[ $name_field_id ]['value'] ) && ! empty( $_POST[ $name_field_id ]['value'] ) ) ? esc_html( $_POST[ $name_field_id ]['value'] ) . " <{$reply_to}>" : /* translators: 1. Reply to email address; eg: email@example.com */ sprintf( __( 'Form Submission <%s>', 'coblocks' ), $reply_to );
+		$reply_to = isset( $_POST[ $email_field_id ]['value'] ) ? esc_html( $_POST[ $email_field_id ]['value'] ) : esc_html( get_bloginfo( 'admin_email' ) );
 
 		/**
 		 * Filter the form email headers.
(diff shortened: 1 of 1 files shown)


olafleur-godaddy:
Does this solve #1969 and #2148 ?

EvanHerman:
@olafleur-godaddy Yes, this one closes those two issues. I didn't know we had specific issues addressing that. I've added them to the top of the PR description.

olafleur-godaddy:
Yeah, they were in the description in JIRA, that's why I was asking ;)

includes/class-coblocks-form.php
```diff
@@ -974,7 +973,6 @@ public function process_form_submission( $atts ) {
 		$email_headers = (array) apply_filters(
 			'coblocks_form_email_headers',
 			array(
-				"From: {$from_name}",
 				"Reply-To: {$reply_to}",
```
fjarrett-godaddy, line 976: @EvanHerman I think we shouldn't scrap the senders name along with the `From` header. It seems like we should do `"Reply-To: {$sender_name} <{$sender_email}>"`

includes/class-coblocks-form.php
```diff
@@ -961,8 +961,7 @@ public function process_form_submission( $atts ) {
 		 */
 		$email_content = (string) apply_filters( 'coblocks_form_email_content', $this->email_content, $_POST, $post_id );
 
-		$reply_to  = isset( $_POST[ $email_field_id ]['value'] ) ? esc_html( $_POST[ $email_field_id ]['value'] ) : esc_html( get_bloginfo( 'admin_email' ) );
-		$from_name = ( isset( $_POST[ $name_field_id ]['value'] ) && ! empty( $_POST[ $name_field_id ]['value'] ) ) ? esc_html( $_POST[ $name_field_id ]['value'] ) . " <{$reply_to}>" : /* translators: 1. Reply to email address; eg: email@example.com */ sprintf( __( 'Form Submission <%s>', 'coblocks' ), $reply_to );
+		$reply_to = isset( $_POST[ $email_field_id ]['value'] ) ? esc_html( $_POST[ $email_field_id ]['value'] ) : esc_html( get_bloginfo( 'admin_email' ) );
```
fjarrett-godaddy, line 964: @EvanHerman I'm curious why we are using `admin_email` as the fallback for the sender's email. I can't think of a reason why a site owner would want to reply back to their own email.

fjarrett-godaddy (changes requested):

includes/class-coblocks-form.php
```diff
@@ -973,7 +974,7 @@ public function process_form_submission( $atts ) {
 		$email_headers = (array) apply_filters(
 			'coblocks_form_email_headers',
 			array(
-				"Reply-To: {$reply_to}",
+				"Reply-To: {$sender_name} <{$sender_email}>",
```
fjarrett-godaddy, line 977: @EvanHerman What happens if one or more of these vars is empty? Just curious what the outcome is.

fjarrett-godaddy: @EvanHerman For instance, a name without an email is useless for a reply to field.

EvanHerman: @fjarrett-godaddy Good question. I thought there was a catch in there that if the email field isn't present the form wouldn't submit. There isn't one, so it looks like in that instance the `sender_email` would be empty, and if the name wasn't present that would also be empty and the `Reply-To` field would just be empty.

Maybe we can sidebar this in standup to come up with a reasonable solution. Maybe if those aren't set then we just don't pass a `Reply-To` header and let `wp_mail()` handle it.

fjarrett-godaddy:
@EvanHerman I think as long as there is a `$sender_email` then we should include the `Reply-To` header, otherwise we send the message with default headers.

includes/class-coblocks-form.php
fjarrett-godaddy: @EvanHerman Yeah I think as long as there is a `$sender_email` then we should include the Reply-To header, otherwise we send the message with default headers. The `$sender_name` is a bonus, if we have it we should include it.

includes/class-coblocks-form.php
```diff
@@ -961,8 +961,16 @@ public function process_form_submission( $atts ) {
 		 */
 		$email_content = (string) apply_filters( 'coblocks_form_email_content', $this->email_content, $_POST, $post_id );
 
-		$sender_name  = isset( $_POST[ $name_field_id ]['value'] ) ? esc_html( $_POST[ $name_field_id ]['value'] ) : '';
 		$sender_email = isset( $_POST[ $email_field_id ]['value'] ) ? esc_html( $_POST[ $email_field_id ]['value'] ) : '';
+		$sender_name  = isset( $_POST[ $name_field_id ]['value'] ) ? esc_html( $_POST[ $name_field_id ]['value'] ) : '';
+
+		$headers = [];
```
fjarrett-godaddy, line 967: @EvanHerman Is WP cool with shorthand array syntax now? (I am)

fjarrett-godaddy:
@EvanHerman This looks good. Assuming you've tested and verified the emails look good and replies work properly in both scenarios (name + email and email only) then I say this is g2g!

fjarrett-godaddy (approved):

includes/class-coblocks-form.php
EvanHerman: @fjarrett-godaddy You are right - I forgot it's against WP coding standards. Good call.

PR #2192, opened 2021-12-07
Change accordion file, test performance

EvanHerman:
### Checklist:
- [ ] My code is tested

### DIFF
--- a/src/blocks/accordion/edit.js
+++ b/src/blocks/accordion/edit.js
@@ -9,7 +9,7 @@ import times from 'lodash/times';
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { createBlock } from '@wordpress/blocks';
+import { createBlock } from '@wordpress/blocks'; 
 import { Fragment } from '@wordpress/element';
 import { InnerBlocks } from '@wordpress/block-editor';
 import { Button, Tooltip } from '@wordpress/components';


olafleur-godaddy:
We don't see the perf results posted. Is it because there is no change in the code of the blocks?

EvanHerman:
@olafleur-godaddy The test results should be here in this comment. I had tweaked one of the blocks just to get them to run, and then reverted the change when things were working
https://github.com/godaddy-wordpress/coblocks/pull/2192#issuecomment-989940007

.dev/performance/index.js
```diff
@@ -233,31 +233,31 @@ async function runPerformanceTests( branches, options ) {
 		results[ testSuite ] = {};
 		/** @type {Array<Record<string, WPPerformanceResults>>} */
 		const rawResults = [];
-		// Run the test three times
-		for ( let i = 0; i < 3; i++ ) {
+		// Run the test two times
```
AnthonyLedesma, line 236: I am not opposed to running this two times as opposed to three but I am curious of the reason for this decision.

.dev/performance/results.js
```diff
@@ -74,13 +74,43 @@ async function getPerformanceTestResults( branch ) {
...
+			}
+
+			iteration++;
+		}
+
+		// eslint-disable-next-line
+		for ( dataKey in dataKeys ) {
+			resultsTable += dataKeys[ dataKey ];
+			// eslint-disable-next-line
+			resultsTable += ' | ' + results[ testSuite ][ 'master' ][ dataKeys[ dataKey ] ] + ' | ' + results[ testSuite ][ branch ][ dataKeys[ dataKey ] ] + ' | ' + results[ testSuite ][ 'change %' ][ dataKeys[ dataKey ] ] + '%\\r\\n';
```
AnthonyLedesma, line 106: Do you think this would be more readable using template literals?
```
` | ${ results[ testSuite ][ 'master' ][ dataKeys[ dataKey ] ] } | ${ results[ testSuite ][ branch ][ dataKeys[ dataKey ] ] } | ${ results[ testSuite ][ 'change %' ][ dataKeys[ dataKey ] ] }%\\r\\n`
```

.dev/bin/comment-on-pr.sh
```diff
@@ -50,7 +50,7 @@ BOT_COMMENTS=$(curl -sS \
   -H "Accept: application/vnd.github.v3+json" \
   "https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME/issues/$PR_ID/comments")
 
-COMMENT=$(echo $BOT_COMMENTS | jq '[.[] | select( .user.login == "godaddy-wordpress-bot" and ( .body | contains( "Code Coverage" ) | not ) )]' | jq '.[-1]')
+COMMENT=$(echo $BOT_COMMENTS | jq '[.[] | select( .user.login == "godaddy-wordpress-bot" and ( .body | contains( "Download" ) ) )]' | jq '.[-1]')
```
AnthonyLedesma, line 53: This is unfamiliar to me. I suppose we are looking perviously for 'code coverage' but that is now changed to 'dowload'. Do you mind explaning why this change was made?

AnthonyLedesma (approved):

.dev/bin/comment-on-pr.sh
EvanHerman: @AnthonyLedesma That should not be "Download" I'm not sure why that was changed. Good catch.

.dev/performance/index.js
EvanHerman: @AnthonyLedesma Yeah, honestly I lowered it to 2 from 3 and it started to stabilize a bit more. I found that running it 3 times caused more failures, and increased the time to run. If you run these tests frequently enough, the performance test builds will fail with the same `Error ""` that we were running into before. The primary reason was just because it appeared the performance test runs were more stable.

.dev/bin/comment-perf-results-on-pr.sh
```diff
@@ -0,0 +1,74 @@
...
+success "Pull Request ID: $PR_ID"
+
+PERF_TEST_RESULTS=$(cat ~/project/post-editor-performance-results.txt)
+
+BOT_COMMENTS=$(curl -sS \
+  -u $GH_LOGIN:$GH_AUTH_TOKEN \
+  -H "Accept: application/vnd.github.v3+json" \
+  "https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME/issues/$PR_ID/comments")
+
+COMMENT=$(echo $BOT_COMMENTS | jq '[.[] | select( .user.login == "godaddy-wordpress-bot" and ( .body | contains( "Performance Test Results" ) ) )]' | jq '.[-1]')
```
EvanHerman, line 43: @AnthonyLedesma I'm not sure why the PR has a different line than this, but I confirmed locally this is what I have.

.dev/bin/comment-on-pr.sh
EvanHerman: @AnthonyLedesma My mistake. I thought this was the other `comment-perf-results-on-pr.sh`. This file is the script that posts the CircleCI artifact link (the CoBlocks build .zip) as a comment on GitHub. There are three comments being posted on each PR. The Download link to the .zip from CircleCI, the Coverage Results and now the Performance Test Results.

Previously, I was just checking for any comment by the `godaddy-wordpress-bot` that didn't contain "Code Coverage", but now that is another comment being posted I specifically am searching for the text "Download" in the comment.

.dev/performance/results.js
EvanHerman: @AnthonyLedesma Yes, it would probably would be.

.dev/bin/comment-on-pr.sh
AnthonyLedesma: Just for my clarity, this needs to be reverted is that correct?

PR #2193, opened 2021-12-08
Change FAQ block title and remove block when empty

jlemay-godaddy:
### Description
Change FAQ block title and remove block when empty

### How has this been tested?
Manually tested

### Checklist:
- [ ] My code is tested

### DIFF
--- a/src/blocks/faq/edit.js
+++ b/src/blocks/faq/edit.js
@@ -32,6 +32,8 @@ const Edit = ( props ) => {
 		clientId,
 	} = props;
 
+	const prevCountRef = useRef();
+
 	const { innerBlocks } = useSelect( ( select ) => ( {
 		innerBlocks: select( 'core/block-editor' ).getBlocks( clientId ),
 	} ) );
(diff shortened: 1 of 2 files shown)


src/blocks/faq/edit.js
```diff
@@ -56,6 +56,7 @@ const Edit = ( props ) => {
 			<div className={ className }>
 				<InnerBlocks
 					allowedBlocks={ ALLOWED_BLOCKS }
+					placeholder={ ( <em>{ __( 'Click here to add an item to the FAQ', 'coblocks' ) }</em> ) }
```
AnthonyLedesma, line 59: This is a cleaner solution than removing blocks. Strange that the other behavior is to show nothing at all. 🤷

AnthonyLedesma (approved):
This works well to ensure that there is not an empty block state in the editor. When you take the time to delete the inner blocks from the FAQ you are presented with the placeholder. However, if you save the page and then reload, the basic empty template is loaded instead of the placeholder.

The only issue I could imagine with this is there might be an unexpected UX for users who are unfamiliar with using the block list view to select and then delete unwanted blocks. That being said I wonder if the ideal behavior would be instead to delete the block given that a user has taken the time to clear out all inner blocks. 

I am on the fence on if repopulating the empty block or deleting the empty block are the best UX on this. I can't really think of a scenario where a person has deleted inner blocks but yet expects to keep the FAQ parent around. If they did they could always replace the block.

I'll approve this so that you can make the call you deem the best fit.

jonathanbardo (approved):

PR #2196, opened 2021-12-09
Fix Linkto controls for Masonry gallery and fix some lightbox styles

AnthonyLedesma:
### Description

Masonry gallery LinkTo settings were not properly applying because I did not have the per_page attribute set on the `getMediaItems` call. Now I handle the call correctly with a 100 image limit or a chunking and iterating method if there are greater than 100 images in the gallery.

I also fix the lightbox cursor magnifying glass on the front-end by improving the selectors on conflicting styles. This is actually part of a bug because we have `editor.scss` styles being loaded on the front-end which needs to be investigated.

### Checklist:
- [x] My code is tested


src/extensions/lightbox-controls/styles/editor.scss
```diff
@@ -1,4 +1,8 @@
-.has-lightbox figure,
-figure.has-lightbox {
-	cursor: default !important;
+.edit-post-visual-editor {
+	// These styles load on the front-end causing lightbox hover state to be invalid.
+	// Modified to be within `.edit-post-visual-editor` selector.
+	.has-lightbox figure
+	figure.has-lightbox {
+		cursor: default !important;
+	}
```
AnthonyLedesma, line 7: Created a related issue here:
https://github.com/godaddy-wordpress/coblocks/issues/2199

src/blocks/gallery-masonry/edit.js
```diff
@@ -285,7 +284,7 @@ function GalleryEdit( props ) {
 			)
 			: innerBlockImages;
 
-		const newCaptions = selectedImages?.reduce( ( previous, image ) => {
+		const newCaptions = Array.from( selectedImages )?.reduce( ( previous, image ) => {
```
olafleur-godaddy, line 287: Is the `?` still necessary? If `selectedImages` is `null` or `undefined`, `Array.from` will fail, right? If it is the empty string, `Array.from` will return `[]` which does not need the `?`

src/components/block-gallery/use-get-media.js
```diff
@@ -18,27 +18,46 @@ export default function useGetMedia( innerBlockImages ) {
...
+			return mediaData || [];
 		},
 		[ innerBlockImages ]
 	);
 
 	if (
-		!! imageMedia && ( imageMedia?.length !== currentImageMedia?.length ||
-        imageMedia?.some( ( newImage ) => ! currentImageMedia.find(	( currentImage ) => currentImage.id === newImage.id	) ) )
+		imageMedia?.length !== currentImageMedia?.length ||
+        imageMedia?.some( ( newImage ) => ! currentImageMedia.find(	( currentImage ) => currentImage.id === newImage.id	) )
```
olafleur-godaddy, line 60: That's a lot of space in the `find(...` ;)

src/blocks/gallery-masonry/edit.js
AnthonyLedesma: Good call. There would be no optional just a fail. This is superfluous and I will remove it.

PR #2209, opened 2021-12-16
Remove disabled styles inside editor.

jrtashjian:
Form inputs that are disabled were applying the default "disabled" styles when viewed inside the editor making them hard to see.

Although we want them to not be actionable, we want them to look like they would on the front-end. Resetting the styles on the editor side resovles this.


AnthonyLedesma (approved):
LGTM

PR #2210, opened 2021-12-16
Check global CoBlocks data to determine if typography controls are enabled

### LINKED ISSUE #2103: Custom font controls only for administrators
### Describe the bug:
`Authors` should be able to change the custom font settings introduced by CoBlocks. But only `Administrators` see them.

### To reproduce:
1. Create a new post
2. Type some text so a paragraph block is created
3. Select the text
4. If you are a user with the role `Administrator` you see a custom font control icon in the popup. If you are a user with a different role, there is only the standard settings available.

### Expected behavior:
Custom font controls should be available to anybody who can edit a post or a page, not just administrators.

### Isolating the problem:

- [x] This bug happens with a default WordPress themes active (Twenty Twenty, Twenty Twenty-One)

### Versions
WordPress version 5.8.1
CoBlocks Plugin version 2.17.0

snovosel:
### Description
This PR fixes an issue where in some cases the global variable `typographyControlsEnabled` was being found as undefined which would cause the Typography controls to render in some cases but not in others. By virtue of being able to access the editor, users should be able to edit typography controls (as long as this variable is set to true).

https://user-images.githubusercontent.com/18115694/146565874-6a977b00-ae6e-4604-9235-eb52e410ba8d.mov

Fixes #2103 

### Types of changes

Bug fix

### How has this been tested?

visually, testing with different user roles

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/components/typography-controls/index.js
+++ b/src/components/typography-controls/index.js
@@ -20,9 +20,9 @@ import TypographyTransforms from './transforms';
  */
 import { __ } from '@wordpress/i18n';
 import { DOWN } from '@wordpress/keycodes';
+import { useEntityProp } from '@wordpress/core-data';
 import { Component, Fragment } from '@wordpress/element';
 import { compose, ifCondition } from '@wordpress/compose';
-import { useEntityProp } from '@wordpress/core-data';
 import {
 	Dropdown,
 	RangeControl,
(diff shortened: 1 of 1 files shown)


EvanHerman (approved):
This looks good to me 👍

PR #2211, opened 2021-12-17
Performance tests fix : revert jsdom update

olafleur-godaddy:
### Description
Looks like the failure in perf tests was due to this update done in #2168 

Will remove the added character before merging.


src/blocks/accordion/index.js
```diff
@@ -1,5 +1,5 @@
 /**
- * External dependencies
+ * External dependencies a
```
AnthonyLedesma, line 2: Last thing to revert. We are good to merge.

AnthonyLedesma (approved):
One final thing to revert and let's get this thing merged! Great work!

EvanHerman:
Wow, nice catch. Hopefully this resolves the issues and we can get back to running perf tests without issue!

PR #2214, opened 2021-12-20
Reduce cache usage by being less specific on key

olafleur-godaddy:
### Description
Cache was generated (and taking more space) each time a new branch was created, even if the dependencies were the same. By removing the branch from the key identifier, new cache will be generated only when the dependencies change.


olafleur:
Merging, as it has been validated by the CircleCI staff.

EvanHerman:
Good call, makes sense!

PR #2219, opened 2022-01-10
Fix Gist file embed

AnthonyLedesma:
### Description

Gist embed seemed to have an issue embedding when any files were specified within the Gist URI. Similar to previous Gist solutions we need a custom solution to load files that include uppercase.

There was another issue where `isActive` was being evaluated as true regardless of `providerNameSlug`. This means that if a user were to select the `core/embed` block from the inserter, that the Gist variation would show regardless. Changes here allow for Gist variation to show when inserting `coblocks/gist`.

### Screenshots

image

### Types of changes

New PHP logic and minor change to JS.

### How has this been tested?

Tested manually in the browser using test Gists.

https://gist.github.com/AnthonyLedesma/33ad1a8cd86da3b6bddbdefa432cb51d
https://gist.github.com/AnthonyLedesma/7d0352e8bc50a8a009c2b930f23d110d

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/gist/index.js
+++ b/src/blocks/gist/index.js
@@ -14,8 +14,8 @@ import transforms from './transforms';
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Icon } from '@wordpress/components';
 import { dispatch } from '@wordpress/data';
+import { Icon } from '@wordpress/components';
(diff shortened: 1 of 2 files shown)


EvanHerman (approved):
Looks good and it's working well on my end. I tested with a few more URLs and had no issue. Good work 👍

PR #2222, opened 2022-01-13
WPEX 1691 - Refactor Events block to remove JQuery

snovosel:
### Description

Refactor the Events block to use a no-dependency carousel package.

### Types of changes

Refactor

### Checklist:
- [x] My code is tested


src/blocks/events/index.js
```diff
@@ -6,26 +6,28 @@ import { EventsIcon as icon } from '@godaddy-wordpress/coblocks-icons';
 /**
  * Internal dependencies.
  */
+import deprecated from './deprecated';
 import edit from './edit';
-import example from './example';
+// import example from './example';
```
olafleur-godaddy, line 11: you can remove commented out code

olafleur-godaddy (approved):
Looks good to me! The corrections you did seem to have fixed the problems that I saw earlier. Good job!

AnthonyLedesma (changes requested):
Dismissed this review. I got confused somehow when looking at this.

src/blocks/events/block.json
```diff
@@ -22,6 +22,10 @@
 		},
 		"customTextColor": {
 			"type": "string"
+		},
+		"showCarousel": {
+			"type": "boolean",
+			"default": true
```
AnthonyLedesma, line 28: I don't think that we need to have this as an attribute because we can base it on the `externalCalendarUrl` attribute whether or not to show the carousel. By not using attributes wherever possible we avoid having to deal with deprecations in the future related to the attribute.

AnthonyLedesma (changes requested):
Based on whether or not the 'External feed' radio is checked we should prevent the post-carousel from rendering unless there is a `externalCalendarUrl` value because we have a situation where the content is rendered using internal posts without the carousel.

We also have this error.

```
Warning: The `value` prop supplied to <select> must be a scalar value if `multiple` is false.
```
postCarouselBlockPlaceholderState

AnthonyLedesma:
Just out of curiosity is there a reason that we do not have the carousel active for external posts?

AnthonyLedesma:
Oh wait. I must have gotten seriously confused because I somehow ended up looking at the post carousel block. Let me clean this up real quick.

AnthonyLedesma (commented):
Okay, I am reviewing the correct thing now. 😂 

Here I have a published events block with the carousel enabled. I cannot see the carousel on the front-end, however.
One thing I saw about these slides in the DOM is they have `data-page="undefined"` attributes. Maybe something broke there. No console errors.
image

src/blocks/events/block.json
AnthonyLedesma: I see now what you are doing with this attribute. Sounds reasonable that we should have this as an attribute given its use on the front-end. Concerning this feature however, I think that we should aim to achieve the same aesthetic output for the Events block in the editor as we have in the front-end. That being said we should not render what looks like a carousel in the editor, regardless of edit state, when the attribute `showCarousel` is false.

AnthonyLedesma (changes requested):
Putting the changes requested back now that I have properly reviewed the Events block. There is a concern with the final output where we have a missing carousel even with the option enabled.

snovosel:
https://user-images.githubusercontent.com/18115694/152194323-4bcd884b-a451-46de-9a4d-fb74731d34b2.mov

@AnthonyLedesma i am not able to reproduce the issue that you found

src/js/coblocks-events.js
```diff
@@ -1,18 +1,45 @@
...
-}( jQuery ) );
+	document.addEventListener( 'DOMContentLoaded', function() {
+		setTimeout( () => {
+			const frontEndContainers = document.querySelectorAll( '.page .wp-block-coblocks-front-events-swiper-container' );
+
+			// there may be multiple event blocks
+			for ( let j = 0; j < frontEndContainers.length; j++ ) {
+				const frontEndContainer = frontEndContainers[ j ];
+
+				if ( frontEndContainer ) {
```
jonathanbardo, line 15: @snovosel I would reverse this condition and return early. It removes 1 level of indentation and makes it easier to read.

src/blocks/events/styles/editor.scss
```diff
@@ -54,3 +53,90 @@
...
+	justify-content: flex-start;
+}
+
+.coblocks-events-block-inner-blocks-container .block-editor-inner-blocks .block-editor-block-list__layout .block-list-appender {
+	position: relative;
+
+	& > .coblocks-block-appender {
+		margin-top: 0;
+	}
+}
```
jonathanbardo, line 142: I think we can simplify this a little bit.

```suggestion
.coblocks-events-block-inner-blocks-container .block-editor-inner-blocks .block-editor-block-list__layout {
	align-items: stretch;
	display: flex;
	flex-direction: column;
	justify-content: flex-start;
	
	.block-list-appender {
		position: relative;

		& > .coblocks-block-appender {
			margin-top: 0;
		}
	}
}
```

src/blocks/events/styles/editor.scss
```diff
@@ -54,3 +53,90 @@
...
+
+.coblocks-events-swiper-container-external-calendar-edit .coblocks-events-nav-button {
+	&__prev,
+	&__next {
+		top: 10%;
+	}
+}
+
+.coblocks-events-block-inner-blocks-container-edit .coblocks-events-nav-button {
```
jonathanbardo, line 101: @snovosel no need to repeat top level class with scss. `.coblocks-events-block-inner-blocks-container-edit` is repeated below. We can simplify.

jonathanbardo (approved):

PR #2224, opened 2022-01-13
Introduce Counter Block

AnthonyLedesma:
### Description

Introduce Counter block using the V2 blocks API. Allows controls for typography, insert an icon block, format numbers according to locale, and timer speed. 

### Screenshots

counterBlockUse

### Types of changes

A new block with JS and PHP logic. The bulk of code is JS with PHP managing the custom patterns.

### How has this been tested?

Cypress tests and Jest tests.

### Checklist:
- [x] My code is tested


AnthonyLedesma:
Unfortunately, the perf tests report is not of value in this case. We are having an inconsistent number of runs between branch and master and thus may introduce fringe results. We handle fringe results by obtaining the median in data. We now keep fringe results because we do not consistently obtain the median. This change in the data aggregation alters the compare point and lessens the value of the data.

src/js/coblocks-counter.js
```diff
@@ -0,0 +1,53 @@
...
+				}, 200 );
+			}
+		} );
+	}
+
+	init = () => {
+		const els = document.querySelectorAll( '.wp-block-coblocks-counter' );
+
+		// If reduce motion is on, don't animate.
+		const reduceMotion = window.matchMedia( '(prefers-reduced-motion: reduce)' ).matches;
```
jrtashjian, line 30: It's easy to forget about this kind of thing. Glad to see it here! We should do this for our carousel's as well.

jrtashjian (approved):
This looks really great!

PR #2228, opened 2022-01-19
Don't make the build pipeline fail if the perf tests fail

olafleur-godaddy:
### Description
The perf tests sometimes fail for a mysterious reason. Instead of making the build pipeline fail, we take what data has been calculated and show it on the PR (when it fails, it is always the second time it calculates the value, so there are already some calculated values).

So, basically, we allow the execution of the performance tests to fail on the pipeline, because it does not represent a failure of our product.

### DIFF
--- a/.dev/performance/index.js
+++ b/.dev/performance/index.js
@@ -11,6 +11,7 @@ const { mapValues } = require( 'lodash' );
 const { formats, log } = require( './logger' );
 const {
 	runShellScript,
+	runShellScriptIgnoreFailing,
 	readJSONFile,
 	askForConfirmation,
 	getRandomTemporaryPath,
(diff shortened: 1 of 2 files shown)


.dev/performance/index.js
```diff
@@ -146,7 +147,7 @@ async function setUpGitBranch( branch, environmentDirectory ) {
  * @return {Promise<WPPerformanceResults>} Performance results for the branch.
  */
 async function runTestSuite( testSuite, performanceTestDirectory ) {
-	await runShellScript(
+	await runShellScriptIgnoreFailing(
```
olafleur-godaddy, line 150: The change is basically this line.

.dev/performance/utils.js
```diff
@@ -47,6 +47,33 @@ function runShellScript( script, cwd, env = {} ) {
...
+					PATH: process.env.PATH,
+					...env,
+				},
+			},
+			function( error, _, stderr ) {
+				if ( error ) {
+					// eslint-disable-next-line no-console
+					console.log( stderr );
+					// We make the promise resolve even if the script fails
+					resolve( true );
```
olafleur-godaddy, line 68: The only different line from the other function is this one.

AnthonyLedesma (approved):
LGTM. Good work!

PR #2229, opened 2022-01-21
Add Testimonials block

jlemay-godaddy:
### Description
Add Testimonials block

### Types of changes
New block

### How has this been tested?
Manually tested + Jest

### Checklist:
- [ ] My code is tested


src/blocks/testimonials/testimonial/edit.js
```diff
@@ -0,0 +1,346 @@
...
+		showImage,
+		showRole,
+		styleName,
+		text,
+		textColor,
+	} = attributes;
+
+	const [ url, setUrl ] = useState( attributes.url || '' );
+
+	const prevShowRole = usePrevious( showRole );
```
jonathanbardo, line 60: @jlemay-godaddy why do we need so many conditions here with usePrevious. The control themselves are not able to modify the attributes directly?

jonathanbardo:
@jlemay-godaddy I was confused at first what "Size" meant under the Testimonials Settings.
image

I think the typography settings should live under their own menu like the paragraph block. See below:

image

jonathanbardo:
@jlemay-godaddy Can we rename item settings to something a little more precise. Eg. Testimonial Item settings? 

Ref:
image

jonathanbardo:
@jlemay-godaddy I think we need a custom appender in the testimonial block here. Right now it seems like the only solution to add another testimonial in the same group is to use the duplicate function inside of the Gutenberg list view. 

image

jonathanbardo:
@jlemay-godaddy I played with the column selector and it seems like when you go from 2 to 3 columns, and then back to 2 you are stuck with an extra testimonial. If the testimonial is not filled out it would be nice if it would go away. Could be a future improvement though.

https://user-images.githubusercontent.com/1933681/150856638-b3bab212-5f6d-4d06-90d9-da568a9cc28d.mov

src/blocks-11.js
```diff
@@ -3,11 +3,15 @@ import { registerBlock } from './utils/helper';
...
+import * as testimonials from './blocks/testimonials';
 
 /**
  * Function to register blocks provided by CoBlocks.
  */
 [
 	posts,
 	postCarousel,
+	testimonial,
+	testimonials,
```
AnthonyLedesma, line 16: Did you happen to check the chunk size for this yet? I wonder if we might just want to split into another blocks chunk. I ended up doing that for the Counter block due to chunk size.

src/blocks/testimonials/block.json
```diff
@@ -0,0 +1,60 @@
+{
+	"name": "coblocks/testimonials",
+	"category": "layout",
+	"attributes": {
```
AnthonyLedesma, line 4: We need to make new blocks using the v2 blocks API. Here is an example block.json file. https://github.com/godaddy-wordpress/coblocks/blob/master/src/blocks/counter/block.json

Specifically, we need to use the version declaration but we also can then use block supports. Another comment in this file regarding fontSize.

src/blocks/testimonials/block.json
```diff
@@ -0,0 +1,60 @@
...
+		},
+		"customBubbleBackgroundColor": {
+			"type": "string"
+		},
+		"fontSize": {
+			"type": "string"
+		},
+		"customFontSize": {
+			"type": "number"
+		}
```
AnthonyLedesma, line 58: These attributes should be removed completely in favor of block supports which define these attributes for us and allow us to align with core.
https://github.com/godaddy-wordpress/coblocks/blob/d233cbc771f8a481ffc771a6434601c3bf742ea3/src/blocks/counter/block.json#L35-L45

src/blocks/testimonials/inspector.js
```diff
@@ -0,0 +1,213 @@
...
+							} }
+							role="button"
+							tabIndex="0"
+						>
+							{ <img alt="" src={ `/wp-content/plugins/coblocks/dist/images/testimonial-style-${ style.name }${ ! showImages ? '--noimage' : '' }.png` } /> }
+							<div className="block-editor-block-styles__item-label block-editor-block-styles__item-label">
+								{ style.label || style.name }
+							</div>
+						</div>
+					) ) }
```
AnthonyLedesma, line 117: I find this solution interesting as we are essentially re-creating the wheel here hacking around the exisitng styles mechanism to accomplish this.

src/blocks/testimonials/inspector.js
```diff
@@ -0,0 +1,213 @@
...
+				</div>
+			</PanelBody>
+
+			<PanelBody
```
AnthonyLedesma, line 121: We have quite a few `PanelBody` elements as well as the `PanelColorSettings` which give us a pan-hell that is not good UX. Can we at minimum combine the color settings into one panel? We should be able to conditionally show them within the panel. 

image

AnthonyLedesma (changes requested):
Some changes are requested.

src/blocks/testimonials/inspector.js
AnthonyLedesma: I mean honestly, I am not huge on this custom component because WordPress core is moving away from showing style previews the same way as they had been coming in WordPress 6 expected this year. I just feel like this control might be diverging us from the ways that Gutenberg is doing things. What do you think @jlemay-godaddy?

jlemay-godaddy: Not sure what the solution would be here? Do we have a block v2 using different styles?

src/blocks-11.js
jlemay-godaddy: I'll double check that at the end once my code is good enough to merge :)

src/blocks/testimonials/inspector.js
AnthonyLedesma: Block styles are defined in the block settings as opposed to the block.json. This is basically the v1 feature that is still compatible with v2. You would use styles the same way that we already are in the codebase. 
https://github.com/godaddy-wordpress/coblocks/blob/f190e7e7922a95e35e583a8e55ffd3ae1b42c272/src/blocks/opentable/index.js#L41-L58

New block styles ( in Gutenberg and in WP 6 ) do not have image previews. Here is a screenshot of what the block styles sidebar will transform into. 
image

If we can use the core mechanism we should. I understand we build something here that is essentially a replacement for that core feature. As Core moves ahead with the style mechanism changes we will start to diverge.

src/blocks/testimonials/styles/style.scss
```diff
@@ -0,0 +1 @@
+// .wp-block-coblocks-testimonials {}
```
AnthonyLedesma, line 1: Are we safe to remove this stylesheet and related imports?

webpack.config.js
```diff
@@ -61,6 +62,14 @@ module.exports = {
...
 			filename: '[name]-rtl.css',
 		} ),
+		new CopyPlugin( {
+			patterns: [
+				{
+					from: path.resolve( process.cwd(), 'assets', 'testimonials' ),
+					to: '[path]/images/[name][ext]',
+				},
+			],
+		} ),
```
AnthonyLedesma, line 72: Do you mind explaining what this plugin is needed for? It seems to be related to the images but I am not positive why this is needed. I see that the `.png` assets that you are using are intended for block styles. For a point of reference for you, we have existing precedence of setting the images with the (core mechanism) block styles.

https://github.com/godaddy-wordpress/coblocks/blob/f190e7e7922a95e35e583a8e55ffd3ae1b42c272/src/blocks/opentable/styles/editor.scss#L9

AnthonyLedesma:
I pushed up a few changes to the block.json to remove the `fontSize` and `customFontSize` declarations which are added `automagically` with v2 block supports. Also removed a superfluous style import and merged with master. 

I am seeing some styling issues in TT2.
testimonialsLargeImageGutters


Honestly, it seems strange to me that the testimonials block does not have alignment support enabled. Here is a group block that someone could conceivably use as a testimonial but this one allows for alignment.
testimonialsAlignment

jlemay-godaddy:
> it seems strange to me that the testimonials block does not have alignment support enabled.

It used to have this with Block v1. I'll have a look why it was removed using v2 ...

src/blocks/testimonials/index.js
```diff
@@ -0,0 +1,68 @@
...
+		'coblocks',
+		/* translators: block keyword */
+		__( 'reviews', 'coblocks' ),
+		/* translators: block keyword */
+		__( 'testimonials', 'coblocks' ),
+	],
+	save,
+	supports: {
+		align: [ 'wide' ],
+	},
```
AnthonyLedesma, line 63: These block supports are now defined in block.json. We should remove this declaration here and instead place it in the block.json. Here is an example where the block supports for `align` defined in the block.json. I might advise exploring allowing `none` `wide` and `full` alignments for more aesthetic flexibility. 

https://github.com/WordPress/gutenberg/blob/004a32ec09d5b789be694d85725fe49cca385f08/packages/block-library/src/group/block.json#L10-L14

https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#align

We are not actually doing this correctly in any block as it turns out even the newest block `Counter` that I made. That block I mistakenly defined an `align` attribute and did not opt into the align supports. This is a failure because there is a good chance that this block will eventually need to be deprecated so that we can properly opt-in to the `align` block support. It looks like you don't even need to import any custom components or anything just opt-in 🤦‍♂️

src/blocks/testimonials/save.js
```diff
@@ -0,0 +1,41 @@
...
+	const sanitizedClasses = ( receivedClasses ) => {
+		const classesArray = receivedClasses.split( ' ' );
+		return classesArray.filter( ( classe ) => ! classe.match( new RegExp( 'has-', 'g' ) ) || classe.match( new RegExp( 'font-size', 'g' ) ) );
+	};
+
+	const classes = classnames( sanitizedClasses( saveBlockProps.className ), {
+		'has-columns': columns > 1,
+		'has-responsive-columns': columns > 1,
+		[ `has-${ columns }-columns` ]: columns > 1,
+		[ `has-${ gutter }-gutter` ]: gutter,
```
AnthonyLedesma, line 30: the `has-gutter-gutter` class logic comes from our Gutter Controls component. 
https://github.com/godaddy-wordpress/coblocks/blob/de4b025d4691febe34e37d78b63a485363ce3757/src/components/gutter-control/gutter-wrapper.js#L34

We have a number of examples through the CoBlocks codebase where we use this component and controls.

PR #2235, opened 2022-01-25
WPEX 1690 - refactor post carousel to remove JQuery

snovosel:
### Description

This PR removes jQuery from the post carousel block in an effort to remove jQuery from the coblocks project

### Types of changes

refactor

### How has this been tested?

visually, cypress

### Checklist:
- [x] My code is tested


olafleur-godaddy:
In the frontend, everything looks fine, but it seems misaligned in the editor with Go.
Screen Shot 2022-01-27 at 10 14 27

src/blocks/post-carousel/test/post-carousel.cypress.js
```diff
@@ -34,17 +34,21 @@ describe( 'Test CoBlocks Post Carousel Block', function() {
...
-			cy.get( '[data-type="coblocks/post-carousel"]' ).find( '.slick-slide:not(.slick-cloned)' ).should( 'have.length', numberOfPosts );
+
+			// we do not render the swiper when 1 slide is prensent
+			if ( numberOfPosts > 1 ) {
+				cy.get( '[data-type="coblocks/post-carousel"]' ).find( '.swiper-slide' ).should( 'have.length', numberOfPosts );
+			}
 		} );
 
-		helpers.checkForBlockErrors( 'coblocks/post-carousel' );
+		// helpers.checkForBlockErrors( 'coblocks/post-carousel' );
```
olafleur-godaddy, line 51: commented out code?

olafleur-godaddy (approved):
Looks good to me!

.dev/tests/phpunit/test-class-coblocks.php
```diff
@@ -190,8 +190,6 @@ public function test_final_build_assets_exist() {
 				'dist/js/coblocks-google-recaptcha.js',
 				'dist/js/coblocks-lightbox.js',
 				'dist/js/coblocks-masonry.js',
-				'dist/js/coblocks-slick-initializer-front.js',
-				'dist/js/coblocks-slick-initializer.js',
```
AnthonyLedesma, line 194: I suspect this will break existing blocks that are using the slick carousel for example people who may have updated CoBlocks but have not yet edited an exiting post with an old post-carousel block. That old block would have old markup and should likely still load the old script. 

We will have a need to keep some jQuery around that is used in a deprecated manner.

AnthonyLedesma: That being said. When I was looking at this block on the master branch earlier today the slick carousel is still working but its sure got some quirks like overlapping text for example.

includes/class-coblocks-block-assets.php
```diff
@@ -412,16 +412,9 @@ public function frontend_scripts() {
...
-				COBLOCKS_VERSION,
-				true
-			);
-			wp_enqueue_script(
-				'coblocks-slick-initializer-front',
-				$dir . 'coblocks-slick-initializer-front.js',
-				array( 'jquery' ),
+				'coblocks-post-carousel',
+				$dir . 'coblocks-post-carousel.js',
+				array(),
```
AnthonyLedesma, line 417: Notice where on like 368 of this file I still keep around old jQuery scripts for the masonry block. That is because people who are using old markup of this block but have not yet gone into the editor and deprecated the block will have a broken experience by updating CoBlocks. We must be careful to account for cases of poeple who have not yet deprecated. Perhaps eventually we will be able to competely drop these scripts and hope that most people are updated and deprecated.

src/blocks/post-carousel/edit.js
```diff
@@ -101,6 +103,32 @@ const PostCarousel = ( props ) => {
...
+					swiperWrapper.classList.remove( 'swiper-wrapper' );
+					swiperWrapper.classList.add( 'swiper-wrapper-editor' );
+
+					const swiperBackButton = postCarouselContainer.querySelector( '#wp-coblocks-post-carousel-swiper-prev' );
+					const swiperNextButton = postCarouselContainer.querySelector( '#wp-coblocks-post-carousel-swiper-next' );
+
+					swiperBackButton.style.display = 'none';
+					swiperNextButton.style.display = 'none';
+				}
+			}, 1000 );
```
AnthonyLedesma, line 124: Why does this need a timeout? Could it instead be an async function where we await for the dom elemnts?

src/blocks/post-carousel/post-item.js
```diff
@@ -0,0 +1,106 @@
...
+
+export default compose( [
+	withSelect( ( select, { clientId } ) => {
+		const blockAttributes = select( 'core/block-editor' ).getBlockAttributes( clientId );
+
+		return {
+			attributes: blockAttributes,
+		};
+	} ),
+] )( PostItem );
```
AnthonyLedesma, line 106: I am confused why this is necessary at all. We are making a store call here when this component has access to the execution context where the `attributes` exist. On line `183` or `159` of `src/blocks/post-carousel/edit.js`. Why not simply pass in the existing attributes? You can modify the `PostItem`s like this:
```JavaScript
<PostItem key={ i } post={ post } { ...{ ...props } } />
```

This change will give you full access to the existing props preventing the need for an extra compose and data-store call.

src/components/swiper/index.js
```diff
@@ -176,10 +177,10 @@ const Swiper = ( props ) => {
 	const renderNavigation = useMemo( () => {
 		return (
 			<>
-				<button className={ `nav-button__prev ${ navigation === false && 'no-navigation' }` } id={ `${ uuid }-prev` }>
+				<button className={ navigationClass ? `${ navigationClass }__prev` : `nav-button__prev ${ navigation === false && 'no-navigation' }` } id={ `${ uuid }-prev` }>
 					<svg className="icon" style={ { transform: 'rotate(180deg)' } } />
 				</button>
-				<button className={ `nav-button__next ${ navigation === false && 'no-navigation' }` } id={ `${ uuid }-next` }>
+				<button className={ navigationClass ? `${ navigationClass }__next` : `nav-button__next ${ navigation === false && 'no-navigation' }` } id={ `${ uuid }-next` }>
```
AnthonyLedesma, line 183: This is a little difficult to read now that we are nesting a tertiary as well here. How do you feel about moving this to use the `classnames` package? For the sake of posterity, I think it would help with understanding the conditional.
```javascript
/**
 * External dependencies
 */
import classnames from 'classnames';

const buttonClasses = classnames( {
			[ `${ navigationClass }__next` ]: navigationClass,
			'nav-button__next': ! navigationClass,
			'no-navigation': navigation === false
		} );
```

AnthonyLedesma (changes requested):
I am seeing overlapping styles in the editor as well as on the front end. Also on TT1 there is this weird border circle around the arrows. 

image

image

src/blocks/post-carousel/post-item.js
snovosel: Because the PostItem component is rendered dynamically via the Swiper component (through render-props https://reactjs.org/docs/render-props.html ) , the individual PostItem components are not triggered to re-render via the parent PostEdit. Instead, they are connected to the store and will re-render following any prop update to which they are listening. In the GalleryCarousel a similar pattern is used, though using the Context API.

This way, the Swiper component itself is not re-rendered following any attribute updates, and only the PostItem components are re-rendered specifically in order to minimize any unnecessary re-renders.

src/blocks/post-carousel/edit.js
snovosel: this is triggering specifically the SSR portion of this block, which renders the output. In this case, because the classnames in the editor conflict with the classnames on the front end script, we need to in some way remove the class so that the front end script (which will run regardless) does not invoke the TinySwiper package. I'm not sure what you mean by async function.

src/blocks/post-carousel/post-item.js
AnthonyLedesma: Minimization of excess renders sounds like a positive for sure. I suspect there is something I am missing still. I guess I am not clear about how if these attributes are already available in the parent and should not change dynamically why we would expect excess renders. When the attribute changes on the parent, the children would re-render as expected because the passed attribute and related useEffect function are watching.

AnthonyLedesma: I feel like I am about to learn something lol.

snovosel: if you look at the below function within the Swiper component,
```
const renderList = useMemo( () => {
		return list.map( ( item, index ) => (
			<div className={ `swiper-slide` } id={ index } key={ index }>
				{ children( {
					index,
					item,
				} ) }
			</div>
		) );
	}, [ isDraggable ] );
```

The Swiper component is invoked within the PostEdit and uses a single child which is a function : 
```
{ ( { index: i, item: post } ) => {
    return (
	<PostItem clientId={ clientId } key={ i } post={ post } setAttributes={ setAttributes } />
    );
} }
```

The rendering order would be as follows --> PostEdit ---> Swiper --> PostItem. However, within the Swiper component function `renderList`, the function itself is only recalculated following the update of the `isDraggable` variable since the function is memoized. Therefore, the children function would only be updated if that variable is changed. This separates the concern of the Swiper and the child component that it is rendering and allows us to abstractly use the Swiper component without any information of what it is rendering, and when it should update. It would only update when Swiper specific props are changed. That being said, the children component will have the first render of the attributes if we pass it down as props, though when the attribute changes , the swiper component will not update itself with the updated attributes and so instead we can use redux to listen to any store updates , but within the component that needs that information instead of drilling props down through each level.

snovosel:
> I am seeing overlapping styles in the editor as well as on the front end. Also on TT1 there is this weird border circle around the arrows.
> 
> image
> 
> image

@AnthonyLedesma  I just updated the space between for the slides on the post carousel. that should handle the style overlap

src/components/swiper/index.js
snovosel: updated this in the newest commit

includes/class-coblocks-block-assets.php
snovosel: if the block deprecates correctly why would there need to be any manual effort on the user's part?

snovosel: or this would be for the front end

snovosel: @AnthonyLedesma i just tested this using a post carousel block on master. previewed the front end, and then switched branches to this branch and without loading the editor refreshed the front end and the carousel is still working. i believe because the markup that the post carousel is using a server side rendered markup the markup is automatically updated and the new front end script is controlling the new version of this block.

.dev/tests/phpunit/test-class-coblocks.php
AnthonyLedesma: Thanks for testing this out. Resolving this comment.

src/blocks/post-carousel/edit.js
AnthonyLedesma: Ok so this is meant to prevent the Swiper script from initializing in the editor. I see also the `src/js/coblocks-post-carousel.js` script has a timeout. Is that related to this timeout for the same reason by chance? Or do you mind explaining the timeout there? I am curious if the timeout is needed because there might not be any elements available (like a race condition) also. I saw something similar with the Counter front-end script where the script could load before the related dom elements.

src/blocks/post-carousel/post-item.js
AnthonyLedesma: I appreciate this explation of why. This makes sense then to keep.

AnthonyLedesma:
I pulled down those latest style changes and rebuilt them. I am still on TT1 here. I am still getting some alignment strangeness on other slides just not that first one now. Here is what I am seeing.
postCarouselSizingUpdated


postCarouselSizingUpdatedAlignment
Also it looks like when we are at full alignment the arrows are offscreen a bit. Maybe we can have styles for those arrows that are targeting the `has-align-wide` or whatever the applicable class is.

snovosel:
@AnthonyLedesma  I am still not seeing the issue you have in your video but it does not look like you have the updated changes. There would be spacing between the slides per the newest commit

https://user-images.githubusercontent.com/18115694/152237141-5c6d0aef-e615-4833-bac9-60c1a1f51b5b.mov


regarding the alignment, where in the block settings is this located? I don't see that option

snovosel:
https://user-images.githubusercontent.com/18115694/152237653-7e336f2d-410a-45ae-b092-a1113b0b6ecb.mov

@AnthonyLedesma screen recording of the front end including the newest commit

AnthonyLedesma:
> regarding the alignment, where in the block settings is this located? I don't see that option

That I show in the following video. It's in the block toolbar.

> 
> postCarouselSizingUpdatedAlignment Also it looks like when we are at full alignment the arrows are offscreen a bit. Maybe we can have styles for those arrows that are targeting the `has-align-wide` or whatever the applicable class is.

snovosel:
@AnthonyLedesma updated the styles in front end and editor to account for the full width and wide width

PR #2236, opened 2022-01-25
Remove CSS from enqueue script

jonathanbardo:
### Description

- Remove CSS for extensions JS. The style was duplicated. 
- Move common.css into the style folder for clarity.


includes/class-coblocks-block-assets.php
```diff
@@ -160,14 +160,6 @@ public function editor_assets() {
...
 		$filepath   = 'dist/' . $name;
 		$asset_file = $this->get_asset_file( $filepath );
-		$rtl        = ! is_rtl() ? '' : '-rtl';
-
-		wp_enqueue_style(
-			$name,
-			COBLOCKS_PLUGIN_URL . 'dist/style-' . $name . $rtl . '.css',
-			array(),
-			$asset_file['version']
-		);
```
AnthonyLedesma, line 170: We are removing the RTL logic here? Why? We have RTL files generated and available to enqueue.

src/styles/index.scss
```diff
@@ -68,6 +68,9 @@
 @import "../blocks/share/styles/editor.scss";
 @import "../blocks/social-profiles/styles/editor.scss";
 
+// @import "../formats/**/style.scss";
+@import "../formats/uppercase/style.scss";
```
AnthonyLedesma, line 72: Thank you for the implicit imports.

AnthonyLedesma (changes requested):
Question about the removal of RTL enqueue.

includes/class-coblocks-block-assets.php
olafleur-godaddy: No, it's because the $rtl variable is calculated earlier already

AnthonyLedesma: Oh you know what. I think I see why you did that now. We import that block_assets function and this editor_assets function.

AnthonyLedesma: As a side note - We really need to refactor this file. We have this redundant import going on and it is very confusing to see what exactly is happening.

https://github.com/godaddy-wordpress/coblocks/blob/dcfae001f4b97cc050cda6d3e438b8e9e945631b/includes/class-coblocks-block-assets.php#L47-L49

AnthonyLedesma (approved):
Excellent. Working great!

includes/class-coblocks-block-assets.php
jonathanbardo: @AnthonyLedesma totally agree!

Issue #2251, opened 2022-01-28
ISBAT add custom icons


zitrusblau:
### Describe the bug:
As from your docs, I should be able to add custom icons and place them in the active theme directory. Unfortunately, this is currently not possible w/ most recent versions tested.
https://github.com/godaddy-wordpress/coblocks/blob/master/docs/hooks/icons-block.md

### To reproduce:
Add a custom .svg file (use the example code from the doc) as shown in the doc and add an icon block. The block won't load. Error in JS console:
`315.js:2099 Uncaught (in promise) TypeError: coblocksBlockData.customIcons[o].icon.replace is not a function
    at 315.js:2099:403
    at Array.forEach (<anonymous>)
    at Module.2315 (315.js:2099:336)
    at Function.o (coblocks-8.js?ver=b8d345148b22c477c69a5ff231703701:3072:57480)
    at async coblocks-8.js?ver=b8d345148b22c477c69a5ff231703701:1:211933`

Tried to add different icons ony by one but the example SVG code from your doc should do it I guess. The error above hints to an issue unrelated to the SVG format/code provided.

### Expected behavior:
Load the block and show the icon within it.

### Isolating the problem:
- [x] This bug happens with no other plugins activated
- [x] This bug happens with a default WordPress theme active (twenty twenty one)
- [x] This bug happens **without** the Gutenberg plugin active
- [x] I can reproduce this bug consistently using the steps above

### WordPress version:
WP 5.9 (also happens on 5.8.3)

Hope you can take a look at this. Thanks!

AnthonyLedesma:
Closed with #2256

zitrusblau:
Okay thanks!
Now the JS errors are gone but no SVG Markup is added in the frontend unfortunately when selecting custom icons.
I also tested with your SVG example code again.

jlemay-godaddy:
I can confirm that it's still broken. The icon will display in the editor, but not in the frontend.

olafleur-godaddy:
Fixed in #2282 and deployed in 2.22.0

PR #2256, opened 2022-01-28
Fix custom icons not loading

jlemay-godaddy:
### Description
Fix custom icons crashing the Icons block.
See: https://github.com/godaddy-wordpress/coblocks/issues/2251

### Types of changes
Bug fix

### How has this been tested?
Manually tested. @olafleur we should find a way to cover custom icons in our tests :)

### Checklist:
- [ ] My code is tested

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -314,6 +314,11 @@ commands:
           command: |
             ./vendor/bin/wp theme install go --activate --path=/tmp/wordpress
       - run:
+          name: Create custom icon
+          command: |
+            mkdir -p /tmp/wordpress/wp-content/themes/go/coblocks/icons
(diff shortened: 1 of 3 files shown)


olafleur-godaddy:
I've added a custom icon in the WordPress installation on CircleCI. That was failing on master, but should pass on this branch :)

olafleur-godaddy (approved):
It does. Good job! :)

AnthonyLedesma (changes requested):
I am able to use custom icon without issue but I am not able to get the optional config json to work. I am fine if we need to change this feature and what we support but let's update the doc at the same time. 

https://github.com/godaddy-wordpress/coblocks/blob/master/docs/hooks/icons-block.md#optional-configuration-file

PR #2272, opened 2022-02-09
Drop PHP tests in 7.3 and add them in 8.1

olafleur-godaddy:
### Description
- Drop unit tests in PHP 7.3
- Add unit tests in PHP 8.1
- Switch to Convenience Images (which are the recommended images since December 31st)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ orbs:
 executors:
   node_latest:
     docker:
-      - image: cimg/node:latest
+      - image: cimg/node:16.14
         auth:
           username: $DOCKERHUB_USERNAME
           password: $DOCKERHUB_PASSWORD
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -4,11 +4,11 @@
 	"type": "wordpress-plugin",
 	"license": "GPL-2.0-only",
 	"require": {
-		"php": ">=5.6"
+		"php": ">=7.3"
```
olafleur-godaddy, line 7: Officially, PHP does not support 7.3 anymore, should I even bump it to 7.4 ?
https://www.php.net/supported-versions.php

composer.json
```diff
@@ -4,11 +4,11 @@
...
 	},
 	"config": {
 		"platform": {
-			"php": "5.6.20"
+			"php": "7.3.33"
```
olafleur-godaddy, line 11: Do you know what is the precise default of PHP we use at GoDaddy @fjarrett-godaddy  ? We could use that version (7.4.something) for the platform.

fjarrett-godaddy: @olafleur-godaddy 7.4.27

olafleur-godaddy: Thank you!

fjarrett-godaddy:
@olafleur-godaddy why did we drop 5.6?

olafleur-godaddy:
Because support for PHP 5 has been discontinued since 10 Jan 2019, so I thought we did not need it.

However, if there's a use case for us to keep it, I'll gladly put it back :)

fjarrett-godaddy:
@olafleur-godaddy OK yeah it doesn't matter if support has been dropped, we still claim our stuff will run on PHP 5.6 so we should test it. Perhaps sometime later this year we can drop it.

jonathanbardo (approved):

PR #2275, opened 2022-02-11
Add isset to $_GET['post']

olafleur-godaddy:
### Description
A user mentioned getting a flicker of a blank page on the admin dashboard here : https://wordpress.org/support/topic/undefined-array-errors-after-updating-to-2-21-6/

It seems that these were notices before, but are now warnings in PHP 8.

### DIFF
--- a/includes/class-coblocks-block-assets.php
+++ b/includes/class-coblocks-block-assets.php
@@ -535,7 +535,7 @@ class CoBlocks_Block_Assets {
 			return true;
 		}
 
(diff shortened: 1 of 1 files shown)


jonathanbardo:
@olafleur-godaddy what do you think to add the check to https://github.com/godaddy-wordpress/coblocks/blob/isset-GET-post/includes/class-coblocks-block-assets.php#L546 `isset( $_GET['revision'] )`  as well?

jonathanbardo (approved):

PR #2290, opened 2022-02-23
try fixing that v1 markup on published pages

AnthonyLedesma:
### Description

In cases where users have old V1 masonry markup already saved and on published pages we have broken galleries. This fallback will allow for the old Masonry markup to continue to function as expected.

**Note**
We ended up coming across a few issues and edge cases with regex that will adequately match the V1 masonry block. 

There is one known edge case that will cause users to superfluously load Masonry/jQuery scripts.

**Edge Case**
If a user has V2 Masonry > (Offset Gallery OR Carousel Gallery) > V2 Masonry.
 - This will cause deprecated masonry and jQuery scripts to load. There are no negative side effects aside from superfluous network requests.

### Screenshots

Before
image

After
image

### Types of changes

Re-Introduce jQuery as a fallback :(

### How has this been tested?

Tested manually.


includes/class-coblocks-block-assets.php
```diff
@@ -479,6 +490,21 @@ public function has_coblocks_block( WP_Post $post_object ) {
...
+	/**
+	 * Determine if the given post content contains any v1 Masonry block.
+	 *
+	 * @access public
+	 * @since  2.22.0
+	 *
+	 * @return boolean True when post content contains a v1 Masonry block.
+	 */
+	public function has_masonry_v1_block() {
+		$v1_regex = '/<!-- wp:coblocks\/gallery-masonry.*|\n*(coblocks-gallery--item).*|\n*<!-- \/wp:coblocks\/gallery-masonry -->/m';
```
jrtashjian, line 502: You shouldn't need the multi-line flag (`/m`) here because you're not using the beginning string (`^`) and end of string (`$`) modifiers. Just the slash without any flags would work.

includes/class-coblocks-block-assets.php
```diff
@@ -479,6 +490,21 @@ public function has_coblocks_block( WP_Post $post_object ) {
...
+
+		preg_match_all( $v1_regex, get_the_content(), $matches );
+		return isset( $matches[0] ) && isset( $matches[0][2] ) && ! empty( $matches[0][2] );
```
jrtashjian, line 505: We don't need to store and perform conditions on any of the matches, we just want to know if there is a match. This can be simplified with:
```
return preg_match( $v1_regex, get_the_content() );
```

jrtashjian (changes requested):
Great job figuring this one out! Minor change requested and then you can merge it.

includes/class-coblocks-block-assets.php
jrtashjian: Posting here for reference a possible new regex so we can simplify the function: https://regex101.com/r/xESuJM/1

```
/<!--\s+?wp:coblocks\/gallery-masonry[\s\S]*?coblocks-gallery--item/
```

jrtashjian: This would also work:
```
/<!--\s+?wp:coblocks\/gallery-masonry.*?coblocks-gallery--item/s
```

AnthonyLedesma:
We ended up coming across a few issues and edge cases with regex that will adequately match the V1 masonry block. 

There is one known edge case that will cause users to superfluously load Masonry/jQuery scripts.

**Edge Case**
If a user has V2 Masonry > (Offset Gallery OR Carousel Gallery) > V2 Masonry.
 - This will cause deprecated masonry and jQuery scripts to load. There are no negative side effects aside from superfluous network requests.

Issue #1161, opened 2019-11-22
Undo and Redo editor functions work as expected with the innerBlocks component


jrtashjian:
### Describe the bug
When using the template functionality of the innerBlocks component we can run into issues with how the undo and redo editor functions behave. Specifically if we are setting a block attribute when the template changes we are creating two undo snapshots: one when the attribute is set and another when the innerBlocks are updated.

We solved this issue in the Form block here: https://github.com/godaddy-wordpress/coblocks/pull/1128#issuecomment-557731336

We need to investigate other areas we are using the innerBlocks component to implement the same solution where needed.

### Screenshots
![Kapture 2019-11-22 at 18 20 37](https://user-images.githubusercontent.com/375788/69467719-4dfd8e80-0d57-11ea-9500-49c4173a7270.gif)

### Expected behavior
Clicking undo once should reset the block back to the state before a template was selected. Clicking redo should set the selected template again.

### Isolating the problem:
- This bug happens with no other plugins activated
- This bug happens with a default WordPress theme active
- This bug happens **without** the Gutenberg plugin active

### Additional context
WordPress 5.3

PR #2292, opened 2022-02-24
WPEX-1677 Undo and Redo editor functions work as expected with the innerBlocks component

snovosel:
### Description

This PR fixes some persisting issues with the undo/redo functions on the Row block.

Original Issue - https://github.com/godaddy-wordpress/coblocks/issues/1161 

### Types of changes

Bug fix

### How has this been tested?

Visually

### Checklist:
- [x] My code is tested


src/blocks/row/edit.js
```diff
@@ -183,24 +187,26 @@ const Edit = ( props ) => {
 						: <>
 							<ButtonGroup aria-label={ __( 'Select row layout', 'coblocks' ) } className="block-editor-inner-blocks__template-picker-options block-editor-block-pattern-picker__patterns">
 								<Button
-									icon="exit"
 									className="components-coblocks-row-placeholder__back"
+									icon="exit"
+									label={ __( 'Back to columns', 'coblocks' ) }
 									onClick={ () => {
+										__unstableMarkNextChangeAsNotPersistent();
```
olafleur-godaddy, line 194: I think these are only needed inside `useEffect`. Could you try without the ones that are not inside `useEffect` and see if it still works?

AnthonyLedesma:
For the sake of thoroughly testing the InnerBlocks component to be sure that we do not have the issue on other locations, I would suggest trying to replicate the bug on all of our current blocks that utilize `InnerBlocks`. This issue can occur when choosing from a variation but can also occur when taking action in the block such as adding a new innerBlock.

image

AnthonyLedesma:
Please be sure to link back to the original issue.

snovosel:
@AnthonyLedesma this issue is in reference to template functionality of InnerBlocks . Are there any examples of InnerBlocks issues when adding a new InnerBlock, and if so should we create a separate issue for that?

AnthonyLedesma:
For clarity templates are not variations. If we apply a template to inner blocks then we are adding blocks to the editor and creating an undo point. From my understanding that is what the issue is written for.

snovosel:
@AnthonyLedesma I didn't notice any issues in other blocks aside from the Row block

AnthonyLedesma (approved):

PR #2294, opened 2022-02-25
Allow only images when uploading to the Services and Food & drink blocks

olafleur-godaddy:
### Description
Added a filter to make sure that the user can only select images to upload.

### Types of changes
Enhancement.

### DIFF
--- a/src/blocks/services/service/edit.js
+++ b/src/blocks/services/service/edit.js
@@ -1,8 +1,8 @@
 /**
  * Internal dependencies
  */
-import InspectorControls from './inspector';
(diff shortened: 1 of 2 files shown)


src/blocks/services/service/edit.js
```diff
@@ -170,14 +175,15 @@ const Edit = ( props ) => {
 	const renderPlaceholder = () => {
 		return (
 			<MediaPlaceholder
-				className="wp-block-coblocks-service__figure"
+				accept={ 'image/*' }
```
olafleur-godaddy, line 178: This is the second line change

PR #2295, opened 2022-02-25
what is build difference with more specific cache keys

AnthonyLedesma:
### Checklist:
- [ ] My code is tested

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -339,21 +339,18 @@ commands:
       - restore_cache:
           keys:
             - composer-deps-{{ .Environment.CACHE_VERSION }}-{{ checksum "composer.lock" }}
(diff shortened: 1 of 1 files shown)


olafleur-godaddy:
I think instead the idea would be to remove these lines 
`- composer-deps-{{ .Environment.CACHE_VERSION }}-{{ checksum "composer.lock" }}`
because I think these are the ones that generate more cache (once every time there's a slight difference in the dependencies.

Another thing interesting to try would be to see if there is no cache at all, what is the build time? If there is not a significant  difference, maybe we could do without (or almost), that would reduce cache to zero.

olafleur-godaddy:
Closing as it as been superseded by #2314

PR #2311, opened 2022-03-08
Lint everything on CI

olafleur-godaddy:
### Description
This PR remove the scoping on CSS, JS and PHP linting.

Each of those step will be a little bit longer, but anyway, these steps are executed in parallel to really longer steps (e2e, perf), so the total build time is not changed.

However it brings a benefit because it allow us to make sure that we do not accidentally break something (by updating dependencies, for example, which would not be pick properly the scoping).

### Types of changes
Improve developer experience.

### How has this been tested?
CircleCI

### Acceptance criteria
Linting is not scoped anymore.

### Checklist:
- [X] My code is tested
- [X] I've included developer documentation 
- [X] I've added proper labels to this pull request


.eslintrc.js
```diff
@@ -41,6 +45,11 @@ module.exports = {
 		'react/jsx-sort-props': [ 'warn', { ignoreCase: true } ],
 		'react/no-deprecated': [ 'warn' ],
 		'react/sort-prop-types': [ 'warn', { ignoreCase: true } ],
+		'template-curly-spacing': [ 'warn' ],
+
+		// Deactivation - we should decide if we want them or not
+		'jest/valid-title': 0,
```
olafleur-godaddy, line 51: @jlemay-godaddy will probably need your opinion on these

jonathanbardo: I would go in favor of having one. Unless there is a scenario where titles don't make sense I'm not aware of?

olafleur-godaddy: In that case, I think the errors are caused because we generate the test title from a variable.

Like this : https://github.com/godaddy-wordpress/coblocks/blob/master/src/blocks/services/test/save.spec.js#L16

jonathanbardo: Oh yeah that must be it. Maybe we can exclude that rule on a case per case basis. If we do it across the board maybe then it's worth disabling.

jonathanbardo (approved):

.eslintrc.js
jlemay-godaddy: I mean, it's a title in a Jest file, I don't really have a strong opinion on that, it's only useful to quickly find the failing test. Let's enforce it with excludes if we have a few tests to fix, or disable the whole rule if it's too much to fix

jlemay-godaddy (approved):

PR #2315, opened 2022-03-09
Issue-2198 - handle custom gutter controls within offset block when set to 0

### LINKED ISSUE #2198: Offset Gallery: themes without specified `figure` margins can break gallery gutter settings in editor
### Describe the bug:
On a theme like Twenty Twenty One where there are no theme-specified margins on `figure` elements in the editor, the gutter setting for the gallery block doesn't apply properly.

Without a theme-provided margin, the browser styles kick in applying default margins, which prevents the editor from accurately reflecting the end result that will be seen when the post/page is published.

### To reproduce:

1. Activate Twenty Twenty-One
2. Create a post
3. Insert an Offset Gallery with multiple images
4. Set gutter to None
5. Observe the spacing between the images in the editor
6. Publish and view the post
7. Notice the spacing is correct (no gutter) on the front end
8. Activate a theme like Twenty Twenty, which does set 0 margins on figure elements in the editor
9. Repeat the rest of the steps and compare results

### Expected behavior:
The editor should reflect the gutterless gallery that's presented on the front end, regardless of the theme's styling.

### LINKED ISSUE #2201: ISBAT Spacing between images in stacked block is visible even when set to 0.
Stacked gallery block doesn't remove spacing between images. It is still there even after reducing the gutter value to 0.

### WordPress version:
5.8.2

### Gutenberg version:
2.19.1

snovosel:
### Description

This PR syncs the gutter custom value to the editor and the front end. Previously, custom gutter controls set to 0 would not completely remove the gutter.

Fixes #2198 
Fixes #2201 

### Screenshots

Screen Shot 2022-03-09 at 12 30 27 PM

### Types of changes

Bug fix

### How has this been tested?

visually

### Acceptance criteria

When gutter controls are set to 0 with the gallery offset block, there should be no visible gutter within the editor or the front end 

### Checklist:
- [x] My code is tested


src/blocks/gallery-offset/edit.js
```diff
@@ -223,26 +224,28 @@ const Edit = ( props ) => {
...
-										onSelect={ onSelectImage( index ) }
-										setAttributes={ ( attrs ) => setImageAttributes( index, attrs ) }
-										caption={ img.caption }
 										aria-label={ ariaLabel }
+										caption={ img.caption }
 										captions={ captions }
-										supportsCaption={ true }
+										gutterCustom={ gutterCustom }
+										id={ img.id }
+										imageIndex={ index }
```
AnthonyLedesma, line 233: This does indeed seem like a reasonable solution given the block's current architecture. I see also how you've edited the GalleryImage component to handle the styles. For a fix like this, we should attempt to fix it from within the GutterWrapper which is responsible for applying _gutter_ related styles in the editor as well as the save function. Here is an example of where custom styles are set specifically within the GutterWrapper component.

https://github.com/godaddy-wordpress/coblocks/blob/fac65d4824fb0c4c64c5a5be0ef5a938ea3a46a1/src/components/gutter-control/gutter-wrapper.js#L39-L41

Another aspect to consider here is that the GalleryImage component is being deprecated and removed from our Gallery blocks in favor of core/image inner blocks following the V2 paradigm. So these changes will soon be undone as we move away from the GalleryImage solution and therefore the fix is better applied in the GutterWrapper.

AnthonyLedesma (changes requested):
Good approach but let's try to keep the gutter logic and related styles within the GutterWrapper component.

PR #2320, opened 2022-03-11
Fix Icon block PHPCS warning

olafleur-godaddy:
### Description
Usage of `file_get_contents` is now discouraged. Replace it with the proper way to do it.

### Types of changes
Fix warning

### Acceptance criteria
No more PHPCS warning on icon/index.php file.

### Checklist:
- [X] My code is tested
- [X] I've included developer documentation

### DIFF
--- a/src/blocks/icon/index.php
+++ b/src/blocks/icon/index.php
@@ -78,7 +78,10 @@ function coblocks_render_icon_block( $attrs ) {
 	$icon_path        = is_readable( "$icon_path$icon_style.svg" ) ? "$icon_path$icon_style.svg" : $icon_path . '.svg';
 
 	// Needed for the call to $wp_filesystem->get_contents(...) .
-	require_once ABSPATH . 'wp-admin/includes/file.php';
+	if ( ! class_exists( 'WP_Filesystem' ) ) {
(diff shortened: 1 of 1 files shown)


src/blocks/icon/index.php
```diff
@@ -77,12 +77,15 @@ function coblocks_render_icon_block( $attrs ) {
 	$custom_icon_path = is_readable( "$custom_icon_path$icon_style.svg" ) ? "$custom_icon_path$icon_style.svg" : $custom_icon_path . '.svg';
 	$icon_path        = is_readable( "$icon_path$icon_style.svg" ) ? "$icon_path$icon_style.svg" : $icon_path . '.svg';
 
+	// Needed for the call to $wp_filesystem->get_contents(...) .
+	require_once ABSPATH . 'wp-admin/includes/file.php';
```
jonathanbardo, line 81: @olafleur-godaddy I would use include or simply require as this will throw an error if another plugin already included the file. Maybe also add a check if the wp filesystem class is already in the scope.

jonathanbardo (changes requested):

olafleur-godaddy:
The ignore of the warning is the correct behaviour here, per @jbardo-godaddy ;)

PR #2325, opened 2022-03-11
[WPEX-1865] Experimenting with forcing migration to core blocks.

jrtashjian:
This PR attempts to convert our blocks into core provided blocks where appropriate - starting with the gallery blocks.

## TL;DR

1. Parse blocks in PHP before the post loads in the editor.
2. Migrate all target blocks into dynamic blocks consisting only of block comment delimeters.
3. Force transform the block to the new block after the editor loads.
4. Leave existing blocks untouched until the post is loaded in the editor.

## What problem are we trying to solve?

In the 4 years CoBlocks has been in development the Block Editor in WordPress has improved tremendously to the point that some of our custom blocks are no longer needed. Early on there was a block frenzy and many block libraries were born from it. Today these libraries are less relevant as core blocks contain a significant portion of the needed functionality and simple extensions can acheive the same result.

## What does the Gutenberg project suggest we do?

According to the [Block Editor Handbook](https://developer.wordpress.org/block-editor/) the way we should be handling "block upgrades" are through [deprecations](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-deprecation/). We have a number of deprecations for blocks already, but what about converting one block to an entirely different one?

There isn't a suggested path here or any mention of this possibility so we're in new territory here.

Our initial process for accomplishing this followed the normal pattern of upgrading blocks with a deprecation but altering our Edit function to transform the block and replace it inside the editor (after block validations have occurred). Take our `coblocks/gist` block for example, converted to a `core/embed` block.

**The `coblocks/gist` block edit function**:
```
edit: ( props ) => {
    const { replaceBlocks } = dispatch( 'core/block-editor' );
    replaceBlocks(
        [ props.clientId ],
        switchToBlockType( props, 'core/embed' )
    );
    return null;
},
```

**The `coblocks/gist` block transform**:
```
transforms = {
    to: [
        {
            type: 'block',
            blocks: [ 'core/embed' ],
            transform: ( { url, file, caption, meta } ) => {
                return createBlock( 'core/embed', {
                    providerNameSlug: 'gist',
                    // ...
                } );
            },
        },
    ],
}
```

This works great with one fallback... we have to keep deprecations around which means legacy code is hanging around as well. This is not ideal. We would prefer to purge as much legacy code as possible and not have to run any deprecations in the future for these removed blocks. We also want to prevent the use of these old blocks while maintaining the visual integrity of these blocks on pages that may never be opened in the editor again.

## A new approach...

Hook into the post content before it loads in the editor.

By taking into consideration how the block editor [serializes and parses](https://github.com/WordPress/gutenberg/blob/wp/5.9/docs/explanations/architecture/data-flow.md#serialization-and-parsing) content and how [block validation is triggered](https://github.com/WordPress/gutenberg/blob/wp/5.9/docs/reference-guides/block-api/block-edit-save.md#validation), it becomes clear that we need to somehow force our block into a [dynamic block](https://github.com/WordPress/gutenberg/blob/wp/5.9/docs/how-to-guides/block-tutorial/creating-dynamic-blocks.md) so that only the [block comment delimiters](https://github.com/WordPress/gutenberg/blob/wp/5.9/docs/explanations/architecture/data-flow.md#delimiters-and-parsing-expression-grammar) are rendered to the post_content.

WordPress provides a PHP function to parse the blocks from the post_content called [`parse_blocks()`](https://developer.wordpress.org/reference/functions/parse_blocks/). This function will parse into an array any blocks defined by a single or pair of block comment delimeters and any content within. This makes it easy for us to target our specific blocks and maintain any content within (like innerBlocks). WordPress also provides a PHP function to serialize the array back into the post_content called [`serialize_blocks()`](https://developer.wordpress.org/reference/functions/serialize_blocks/).

Using these two functions we can hook into `the_post` when we are on a block editor admin page. From there we `parse_blocks()` and loop through to find the blocks we're force migrating, run the migration and update the array of parsed blocks with the new content, and finally update the `$post->post_content` with `serialize_blocks()`.

All of this is triggered within [`includes/block-migrate/loader.php`](https://github.com/godaddy-wordpress/coblocks/blob/8591c0e858f60b77b75b760e5907f06fc21ef984/includes/block-migrate/loader.php).

### Migrating a block into a dynamic block

An abstract class ([`CoBlocks_Block_Migration`](https://github.com/godaddy-wordpress/coblocks/blob/8591c0e858f60b77b75b760e5907f06fc21ef984/includes/block-migrate/class-coblocks-block-migration.php)) has been created which can be extended to create individual block migrations. Each class needs to provide the `block_name()` and `migrate_attributes()` methods. The `block_name` method defines what block we are targetting for a migration, and the `migrate_attributes` method returns the new attributes to be saved in the new block comment delimeter.

The `CoBlocks_Block_Migration` abstract class contains methods to assist in parsing classnames and attributes from [DOMElement](https://www.php.net/manual/en/class.domelement.php)'s using [DOMDocument](https://www.php.net/manual/en/class.domdocument.php) and [DOMXPath](https://www.php.net/manual/en/class.domxpath.php).

An example of pulling values from the DOM to build a new attributes array:

```
return array(
	'className' => $this->get_attribute_from_classname( 'has-border-radius-', $this->gallery_wrapper )
		? 'is-style-default'
		: 'is-style-compact',
	'filter'    => $this->get_attribute_from_classname( 'has-filter-', $this->gallery_wrapper ),
	'align'     => $this->get_attribute_from_classname( 'align', $this->block_wrapper() ),
	'lightbox'  => $this->get_attribute_from_classname( 'has-lightbox', $this->block_wrapper() ),
);
```

### Transform our dynamic block into the new core block

Now that we've parsed our old block and saved it as a dynamic block we can transform it into the core block. This part will have to be completed in the editor. We will use the same process we used in the initial approach of forcing a block transform of the block. This requires a simple `edit` function which calls `replaceBlocks()` and `switchToBlockType()`, an empty `save` function (`() => null`), and a single transform to the block you are migrating into.

With all of that in place the attributes defined inside `block.json` can be pruned to only what the class extending `CoBlocks_Block_Migration` defines. It's best to match these attributes to the block we're transforming into to keep the transform code simple. In the case of the `coblocks/gallery-stacked` block we need to loop through the `images` attribute array to create a bunch of `core/image` blocks for the `core/gallery` block.

Example from the [`core/gallery-stacked` block](https://github.com/godaddy-wordpress/coblocks/blob/8591c0e858f60b77b75b760e5907f06fc21ef984/src/blocks/gallery-stacked/index.js):

```
edit: ( props ) => {
    const { replaceBlocks } = dispatch( 'core/block-editor' );
    replaceBlocks(
        [ props.clientId ],
        switchToBlockType( props, 'core/gallery' )
    );
    return null;
},
parent: [], // prevents the block from showing up within the block inserter.
save: () => null,
transforms: {
    to: [
        {
            blocks: [ 'core/gallery' ],
            transform: ( attributes ) => {
                const galleryAttributes = Object.fromEntries(
                    Object.entries( attributes ).filter(
                        ( attribute ) => 'images' !== attribute[ 0 ]
                    )
                );
                return createBlock(
                    'core/gallery',
                    Object.assign( {}, BLOCK_VARIATION_GALLERY_STACKED.attributes, galleryAttributes ),
                    attributes.images.map( ( image ) => {
                        return createBlock( 'core/image', image );
                    } ),
                );
            },
            type: 'block',
        },
    ],
},
```

### Finishing up the migration

All that is required after the migration is to extend the core blocks where appropriate to fully move the features of the old block to the core block (with discretion). This includes creating new [block styles](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-styles/), [block variations](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-variations/), or [block controls](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/block-controls-toolbar-and-sidebar/).

Now what about blocks on published pages that may never be opened in the editor again? The great thing about the block editor is the HTML the block outputs is save in the `post_content` and loaded as is. We only need to ensure the styles we had are still loaded when the block has been detected on the page. This is the only legacy code required to exist

## Conclusion

There you have it. This seems to achieve the ideal scenario while maintaining the integrity of a user's existing content.


src/block-styles/core/gallery/index.js
```diff
@@ -0,0 +1,30 @@
...
+	},
+	{
+		label: __( 'Tiled', 'coblocks' ),
+		name: 'tiled',
+	},
+	{
+		label: __( 'Layered', 'coblocks' ),
+		name: 'layered',
+	},
+].forEach( ( variation ) => registerBlockStyle( 'core/gallery', variation ) );
```
AnthonyLedesma, line 30: 😍

src/extensions/animation/index.js
```diff
@@ -19,9 +19,7 @@ import { useEffect, useRef, useState } from '@wordpress/element';
 const allowedBlocks = [
 	{ animateChildren: false, blockType: 'coblocks/gallery-carousel' },
 	{ animateChildren: true, blockType: 'coblocks/gallery-collage' },
-	{ animateChildren: false, blockType: 'coblocks/gallery-masonry' },
 	{ animateChildren: true, blockType: 'coblocks/gallery-offset' },
-	{ animateChildren: true, blockType: 'coblocks/gallery-stacked' },
```
AnthonyLedesma, line 24: This change needs to be made for the collage block as well correct?

webpack.config.js
```diff
@@ -38,6 +38,8 @@ module.exports = {
 		...coblocksChunks,
 		'coblocks-extensions': path.resolve( process.cwd(), 'src/extensions.js' ),
 		'coblocks-plugin-deactivation': path.resolve( process.cwd(), 'src/components/coblocks-deactivate-modal/index.js' ),
+		'coblocks-styles': path.resolve( process.cwd(), 'src/block-styles.js' ),
+		'coblocks-variations': path.resolve( process.cwd(), 'src/block-variations.js' ),
```
AnthonyLedesma, line 43: I wonder now if we should name this differently for posterity. These are programmatic 'style' extensions as opposed to CSS style-related code but the naming might confuse. Maybe these should instead say 
`coblocks-styles-extension` and `coblocks-variations-extension` for clarity. What do you think @jrtashjian?

AnthonyLedesma (commented):
I'll see if I can track it down but I am getting some errors.

image
image


Also some warnings.
image

jrtashjian:
@AnthonyLedesma I'd try running `yarn clean && yarn setup` to see if that fixes it. I'm not getting any errors.

src/extensions/animation/index.js
jrtashjian: yup! Removed.

jrtashjian:
I've created a new branch we can branch from and merge into for all the incoming migration work. Rather than pushing migrations to the main branch and deploying them adhoc. This way we can ensure we solve any new problems as a whole and have a smoother transition when we do deploy.

New branch name: `the-great-migration`

_I am merging this PR into the new base branch._

PR #2334, opened 2022-03-21
Remove extraneous element from extensions HoC

jrtashjian:
Remove extraFeature from animation as well as its implementation. This has been deemed unnecessary as the toolbar might be expected to shift as the block does.


AnthonyLedesma:
There is a single `extraFeatures` that exists for the animation extension. The extra feature acts to hide the block toolbar while the animation is taking place. I didn't realize the impact of the current design but I think we may want to improve the implementation because the toolbar movement can be jarring. Here is what the extra feature looks like on the master branch.
AnimationTemporaryStatus


And here is the toolbar behavior in this PR. 
AnimationTemporaryStatusDisabled

AnthonyLedesma:
@jrtashjian I just found a bug with the `extraFeature` that I mentioned. Maybe we should consider removing it anyway. I am not sure how we could adjust for this.

**Normal**
animationExtraFeatureToolbarNormal

**Bug if you rush - hides the toolbar forever**
animationExtraFeatureToolbarRushed

AnthonyLedesma:
Tested this in Go, TwentyTwenty, TwentyTwentyOne, and TwentyTwentyTwo.

fjarrett-godaddy:
@AnthonyLedesma it looks like the performance tests indicate a decent degradation, should we be concerned about this?

AnthonyLedesma:
>  I'm wondering if hiding the toolbar is really a benefit. It's obviously built to float around as necessary even if the anchor location changes. In my mind that's intended functionality. I'd prefer we just remove this style override so we have a stable content layout.

At this point, I agree. It would be nice but the truth is that we accomplish it now in a very hacky way. I'll push up the changes that remove this extra feature and its implementation.

jrtashjian:
✅

PR #2337, opened 2022-03-23
Try new editorScript property in block.json

AnthonyLedesma:
### Description

This is take-two at attempting to fix the WordPress.org plugin details page for CoBlocks.

The #Meta team in WordPress slack has identified that this block.json property is one that is required for block registration. Going to give a shot at specifying both paths and slugs to check if there is a difference.

### Types of changes

Removing a [previous modification here](https://github.com/godaddy-wordpress/coblocks/pull/2332) that adds a translation function.

### How has this been tested?

This will need to be deployed in a versioned release in order to test this proposed fix.

### Acceptance criteria

The AC should be resolving the issue with improper block names on the plugin details page. Unfortunately, we will need to test this by releasing a version of CoBlocks.

### DIFF
--- a/src/blocks/accordion/accordion-item/block.json
+++ b/src/blocks/accordion/accordion-item/block.json
@@ -25,5 +25,8 @@
 		"borderColor": {
 			"type": "string"
 		}
-	}
+	},
+	"title": "Accordion Item",
+	"textdomain": "coblocks",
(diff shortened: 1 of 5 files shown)


EvanHerman (approved):
Simple enough. Looks good to me 👍🏼 Hoping this helps fix things!

PR #2347, opened 2022-04-01
Fix OpenTable typo

olafleur-godaddy:
The correct capitalization for this restaurant reservation service is OpenTable.


EvanHerman (approved):
Nice catch 👍

olafleur-godaddy:
Found by someone from the i18n team ;)

PR #2350, opened 2022-04-05
Introduce CoBlocks Labs

AnthonyLedesma:
### Checklist:
- [ ] My code is tested


includes/class-coblocks-site-design.php
```diff
@@ -0,0 +1,407 @@
...
+			'version'      => COBLOCKS_VERSION,
+		);
+
+		// short-circuit.
+		$active_theme = wp_get_theme();
+		if ( 'Go' !== $active_theme->get( 'Name' ) ) {
+			return array();
+		}
+
+		$current_design_style = \Go\Core\get_design_style();
```
olafleur-godaddy, line 127: It seems to me that this will fail badly if we are not in the Go theme, right?

includes/class-coblocks-labs.php
```diff
@@ -0,0 +1,199 @@
...
+	 * Go theme is installed.
+	 *
+	 * @since 2.22.4
+	 * @return string
+	 */
+	public static function is_go_theme_installed() {
+		$go_theme_path   = get_theme_root() . '/go/index.php';
+		$is_go_installed = file_exists( $go_theme_path );
+		return $is_go_installed;
+	}
```
EvanHerman, line 73: I think these two functions can be replaced with a single function. We do something similar in nextgen.

```php
public static function is_go_active() {
	return defined( 'GO_VERSION' ) && 'go' === get_option( 'stylesheet' );
}
```

This should check that Go is both installed and active, although I think these is even a duplicate check. Something like this should be enough to ensure Go is installed and active.

```php
public static function is_go_active() {
	return defined( 'GO_VERSION' );
}
```

includes/class-coblocks-labs.php
```diff
@@ -0,0 +1,199 @@
...
+
+	/**
+	 * Site themes.php page.
+	 *
+	 * @since 2.22.4
+	 * @return string
+	 */
+	public static function get_themes_php_uri() {
+		return admin_url( 'themes.php' );
+	}
```
EvanHerman, line 83: I don't think we need this helper function since it's only being used in one place. I think we can remove a lot of these helpers used inside of `get_coblocks_labs_data()` and just return in the data we need without the helpers:

Also, to check if the launch guide should be enabled, maybe we can just check if the `wpnux_export_data` is empty of not, similar to how we do in nextgen.

```php
! empty( get_option( 'wpnux_export_data' ) );
```

```php
/**
 * Function get_coblocks_labs_data is used to set a global object
 * `coblocksLabs` with relevant data.
 *
 * @since 2.22.4
 * @return array Array of relevant CoBlocks labs data.
 */
public static function get_coblocks_labs_data() {
	return array(
		'isGoThemeActive'      => defined( 'GO_VERSION' ) && 'go' === get_option( 'stylesheet' ),
		'isGoThemeInstalled'  => defined( 'GO_VERSION' ) && 'go' === get_option( 'stylesheet' ),
		'goThemeInstallUri'     => admin_url( 'theme-install.php?theme=go' ),
		'goThemeDetailsUri'   => admin_url( 'themes.php' ) . '?theme=go',
		'launchGuideEligible'  => ! empty( get_option( 'wpnux_export_data' ) ),
		'isLabsEnabled'           => apply_filters( 'coblocks_labs_controls_enabled', false ),
	);
}
```
Or defined them in variables above the return.

```php
/**
 * Function get_coblocks_labs_data is used to set a global object
 * `coblocksLabs` with relevant data.
 *
 * @since 2.22.4
 * @return array Array of relevant CoBlocks labs data.
 */
public static function get_coblocks_labs_data() {
	$is_go_active                       = defined( 'GO_VERSION' ) && 'go' === get_option( 'stylesheet' );
        $is_launch_guide_enabled = get_option( 'wpnux_export_data' ) === false ? false : true;
	
         return array(
		'isGoThemeActive'      => $is_go_active,
		'isGoThemeInstalled'  => $is_go_active,
		'goThemeInstallUri'     => admin_url( 'theme-install.php?theme=go' ),
		'goThemeDetailsUri'   => admin_url( 'themes.php' ) . '?theme=go',
		'launchGuideEligible'  => ! empty( get_option( 'wpnux_export_data' ) ),
		'isLabsEnabled'           => apply_filters( 'coblocks_labs_controls_enabled', false ),
	);
}
```

Do we need to both check `isGoThemeActive` and `isGoThemeInstalled` or can we just check if it's active?

I think we can clean up some of these helpers since they're only used one time in this class and aren't going to be used outside of them. That will just help trim down the length of this file.

includes/class-coblocks-block-assets.php
```diff
@@ -291,6 +291,7 @@ public function editor_assets() {
 				'typographyControlsEnabled'      => $typography_controls_enabled,
 				'animationControlsEnabled'       => $animation_controls_enabled,
 				'localeCode'                     => get_locale(),
+				'baseApiUrl'                     => COBLOCKS_API_NAMESPACE,
```
EvanHerman, line 294: The key here is `baseApiUrl` but the `COBLOCKS_API_NAMESPACE` contains `coblocks`. We should either update the array key here or the key should reference an actual URL.

src/extensions/site-content/index.js
```diff
@@ -0,0 +1,112 @@
...
+						postType={ postType } />
+				) ) }
+			</ComplementaryArea>
+		</>
+	);
+};
+
+/* istanbul ignore next */
+registerPlugin( PLUGIN_NAME, {
+	icon,
```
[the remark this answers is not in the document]
AnthonyLedesma: The new ID is `coblocks-menu-icon-site-design`

.circleci/config.yml
```diff
@@ -350,7 +350,7 @@ commands:
           name: Install dependencies missing from cache
           command: |
             PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
-            yarn install --frozen-lockfile
+            npm ci
```
olafleur-godaddy, line 353: I saw that code in another PR (@EvanHerman 's one I think), is that intended?

.dev/tests/jest/setup-globals.js
```diff
@@ -25,3 +25,244 @@ global.coblocksLayoutSelector = {
...
+
+// Mock site design state for tests.
+// Disable reason: sorting keys for this object is superfluous.
+/* eslint-disable sort-keys */
+global.siteDesign = {
+	availableDesignStyles: {
+		traditional: {
+			slug: 'traditional',
+			label: 'Traditional',
+			url: 'http://localhost:8888/wp-content/themes/go/dist/css/design-styles/style-traditional.css',
```
olafleur-godaddy, line 37: `wp-env` uses `localhost:8889` for tests, right?

src/extensions/coblocks-settings/coblocks-settings-toggle-control.js
```diff
@@ -9,15 +9,17 @@ import { useEntityProp } from '@wordpress/core-data';
...
 	const [ setting, saveSetting ] = useEntityProp( 'root', 'site', settingsKey );
 
 	return (
 		<CoBlocksSettingsModalControl>
 			<CheckboxControl
-				label={ label }
-				help={ help }
 				checked={ !! setting }
+				// With the introduction of CoBlocks labs we now disable the Layout Selector when Go theme is not active.
+				disabled={ shouldDisable ? true : false }
```
olafleur-godaddy, line 20: `disabled={ shouldDisable }` should be enough, right?

.circleci/config.yml
EvanHerman: @olafleur-godaddy We discussed migrating back to `npm` a few weeks ago in Slack so we remain inline with WordPress core. https://godaddy.slack.com/archives/GKQGZCGK1/p1647969507095719

includes/class-coblocks-labs.php
AnthonyLedesma: I forgot that the go theme installation status is an AC because we need to include a conditional URI to either install or activate depending on the install status.

includes/class-coblocks-site-content.php
```diff
@@ -0,0 +1,111 @@
...
+
+	/**
+	 * Retreive the available post types
+	 */
+	private function get_page_nav_post_types() {
+
+		$post_types = get_post_types();
+		$white_list = array(
+			'page' => 'pages',
+			// 'post'    => 'posts', // @codingStandardsIgnoreLine
```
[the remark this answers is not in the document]
AnthonyLedesma: The comments should be removed I suppose. We went a different direction with it I believe and these are remnants.

.dev/tests/jest/setup-globals.js
AnthonyLedesma: Ooo. Hmm, good question.

src/extensions/coblocks-settings/coblocks-settings-toggle-control.js
AnthonyLedesma: Good call.

AnthonyLedesma:
That welcome guide bug turns out to be an issue even if we have the controls rendered conditionally. We use `ifConditon` HOC along with the `registerPlugin` calls for Site Design and Site Contents which prevents render of the component when a PHP filter is not detected. Because however the plugin has been registered with the render function containing a `ComplementaryArea` component we still have this error persisting. I will need to opt-out of these controls in a hacky way instead of using the HOC by preventing the registerPlugin call.

AnthonyLedesma:
Actually, this popup is always around. I am not sure what is going on with that. I don't think we can merge this in until we have this figured out.

olafleur-godaddy:
Would it be interesting to have at least one Cypress test to make sure that Coblocks Labs appear in the choices of the menu?

AnthonyLedesma:
> Would it be interesting to have at least one Cypress test to make sure that Coblocks Labs appear in the choices of the menu?


I think we should indeed have tests for that purpose. However, at this time Labs does not run in the context of E2E where we would likely test labs. That is simply because we have it behind a filter. Not sure about how to get around that in e2e at the moment.

jrtashjian (commented):
Although not required, it would be good to have a clearer layout on mobile. I did a quick layout in Web Inspector with some simple flex layout changes to make it look like this:

**Before**

Screen Shot 2022-04-27 at 16 27 20

**After**

Screen Shot 2022-04-27 at 16 27 10

jrtashjian (commented):
There is an odd interaction when enabling the layout selector. Steps to reproduce:

1. Make sure "layout selector" in Labs is turned off.
2. Create a new "page" post type (not a "post").
3. Without making any edits to the page, open up Labs from the "more menu."
4. Toggle "layout selector" on.
5. The layout selector is immediately opened (because of conditions in place to do so).

We either need to dispatch the proper action to close the layout selector immediately after toggling on the feature, or alter the condition in some other way.

https://user-images.githubusercontent.com/375788/165626968-02eb9b5d-35a6-4608-8f9e-d33642d4fa07.mp4

AnthonyLedesma:
> There is an odd interaction when enabling the layout selector. 

Nice find. I have pushed up a fix.

AnthonyLedesma:
> Although not required, it would be good to have a clearer layout on mobile. 

added some styles for mobile layout

includes/class-coblocks-labs.php
```diff
@@ -0,0 +1,142 @@
...
+		wp_localize_script(
+			'coblocks-editor',
+			'coblocksLabs',
+			array(
+				'isGoThemeActive'     => self::is_go_active(),
+				'isGoThemeInstalled'  => self::is_go_installed(),
+				'goThemeInstallUri'   => admin_url( 'theme-install.php?theme=go' ),
+				'goThemeDetailsUri'   => admin_url( 'themes.php' ) . '?theme=go',
+				'launchGuideEligible' => ! empty( get_option( 'wpnux_export_data' ) ),
+				'isLabsEnabled'       => apply_filters( 'coblocks_labs_controls_enabled', false ),
```
jrtashjian, line 135: Do we really need this? The AC states that each individual feature should be off by default but I believe we want CoBlocks Labs to be available at all times. I would remove this condition throughout the codebase or if it's legitimately needed, default to `true`.

jrtashjian (changes requested):

includes/class-coblocks-labs.php
AnthonyLedesma: Labs should be disabled by default at this time except unless explicitly enabled with a PHP filter.

AnthonyLedesma: Oh wait. I think I misunderstood that original ask. (then failed to mark it as an AC.) let me go ahead and remove this whole filter thing now.

package.json
```diff
@@ -83,11 +83,12 @@
 		"@jest/core": "^27.4.5",
 		"@react-google-maps/api": "^2.7.0",
 		"@testing-library/jest-dom": "^5.16.1",
+		"@testing-library/react-hooks": "^7.0.2",
 		"@wordpress/api-fetch": "^5.2.4",
 		"@wordpress/base-styles": "^4.0.2",
 		"@wordpress/blob": "^3.2.1",
 		"@wordpress/block-editor": "8.0.11",
-		"@wordpress/block-library": "^6.0.2",
+		"@wordpress/block-library": "6.0.2",
```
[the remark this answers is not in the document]
AnthonyLedesma: Yes, that is the case. I forget the exact error but it was something like missing deps error.

PR #2357, opened 2022-04-11
Introduce godaddy global styles

jrtashjian:
Requires and initializes [godaddy-wordpress/styles](https://github.com/godaddy-wordpress/styles/pull/2) as a dependency. This PR is currently pulling in the remote repositorie's `add-wp-env` branch. A PR will be submitted to change this once the remote branch is merged.


class-coblocks.php
```diff
@@ -188,6 +194,8 @@ public function is_amp() {
 	}
 endif;
 
+// require_once dirname( __FILE__ ) . '/vendor/autoload.php';
+
```
AnthonyLedesma, line 198: Should this be commented out like this?

class-coblocks.php
```diff
@@ -134,6 +134,12 @@ private function includes() {
 				require_once COBLOCKS_PLUGIN_DIR . 'includes/admin/class-coblocks-install.php';
 				require_once COBLOCKS_PLUGIN_DIR . 'includes/admin/class-coblocks-crop-settings.php';
 			}
+
+			// StylesLoader.
+			require_once COBLOCKS_PLUGIN_DIR . 'includes/Dependencies/GoDaddy/Styles/StylesLoader.php';
+			GoDaddy\WordPress\Plugins\CoBlocks\Dependencies\GoDaddy\Styles\StylesLoader::getInstance()->setBasePath( COBLOCKS_PLUGIN_DIR . 'includes/Dependencies/GoDaddy/Styles/' );
+			GoDaddy\WordPress\Plugins\CoBlocks\Dependencies\GoDaddy\Styles\StylesLoader::getInstance()->setBaseUrl( COBLOCKS_PLUGIN_URL . 'includes/Dependencies/GoDaddy/Styles/' );
+			add_action( 'plugins_loaded', array( GoDaddy\WordPress\Plugins\CoBlocks\Dependencies\GoDaddy\Styles\StylesLoader::getInstance(), 'boot' ) );
```
AnthonyLedesma, line 142: Will this dependency be available in the built version of CoBlocks? Just thinking about the final product there.

AnthonyLedesma (approved):
One comment about a line that is commented out.

Other than that I guess I am not sure what all to expect here. Composer commands seem to run as expected. Is there anything here that will be changing? Or is this just allowing access to this new dependency?

jrtashjian:
Yes, just the new stylesheet is loaded so that `.godaddy-styles` is available to apply our overrides. The new "StylesLoader" dependency handles all the conditional loading of the script when multiple plugins are trying to enqueue the same style handle. It also handles loading the proper stylesheet for the currently installed version of WP (should we need that granularity). All that logic will be global and handled inside [@godaddy-wordpress/styles](https://github.com/godaddy-wordpress/styles) repo.

class-coblocks.php
jrtashjian: Removed.

PR #2365, opened 2022-04-25
Add a validation for the presence of the minor part in the version

olafleur-godaddy:
### Description
When executing this code on betas or RCs version, there is not necessarily a minor part in the version, so there was a warning message that shown (and made all the Cypress tests fail when testing with 6.0-beta2 ;) )

### Screenshots
Before : 
Screen Shot 2022-04-24 at 22 55 52

After :
Screen Shot 2022-04-24 at 22 56 57


olafleur-godaddy:
This code needs to be in GoDaddy styles, see : https://github.com/godaddy-wordpress/styles/pull/4

olafleur-godaddy:
This code needs to be in GoDaddy styles, see : https://github.com/godaddy-wordpress/styles/pull/4

PR #2367, opened 2022-05-02
Migrate posts block to core

AnthonyLedesma:
### Description

Migrate the `coblocks/posts` block conditionally into core blocks. External feeds are mapped into the `core/rss` block. Internal post feeds are mapped into the `core/query` block. 

Traditional layouts offered by RSS and Query blocks are varied enough not to warrant the use of additional styles/variations. I propose the current migrations are close enough aesthetically to be good as-is.

### Screenshots

image

image

### Types of changes

JS and PHP changes. Changes here for the migration into core/query or core/rss depending on attribute settings.

### Acceptance criteria

Should migrate any use of `coblocks/posts` either into `core/query` OR `core/rss`.

### Checklist:
- [x] My code is tested


jrtashjian (changes requested):
This is fantastic and works great!

The only change we need is to keep the posts block `index.php` file around for the server rendering to continue working with existing blocks. Other than that, it's a nice transition to the core blocks.

includes/block-migrate/class-coblocks-block-migration.php
```diff
@@ -111,7 +110,11 @@ protected function block_wrapper() {
 	 *
 	 * @return string attribute value.
 	 */
-	protected function get_attribute_from_classname( $classname_prefix, DOMElement $element ) {
+	protected function get_attribute_from_classname( $classname_prefix, $element ) {
+		if ( empty( $element ) ) {
+			return '';
+		}
+
```
jrtashjian, line 117: What is the reason for removing the `DOMElement` type hint? We don't want to accept any value here since we expect the attributes property to exist with the `getNamedItem` method.

AnthonyLedesma: Okay reverted this change. I think this was something that I changed early on while I was testing different approaches but this is not needed now.

PR #2369, opened 2022-05-03
WPEX-1925 - adjust focus color on accordion block

### LINKED ISSUE #2303: Accordions - Keyboard focus indication very subtle and hard to see
### Describe the bug:
Accessibility issue for sighted keyboard-only users.

I've added Accordion blocks to a page, and viewed the page in my browsers (Firefox and Chrome). When tabbing around the page using keyboard, the focus indication on the `<summary>` element is very subtle and hard for users to see. There are no settings within the Gutenberg editor to allow users to set a more obvious focus style.

### To reproduce:

1. Add Coblocks Accordions to a page
2. Choose a selection of colours
3. Publish or Update the page
4. View page in browser and tab down page to real the Accordions
5. Note that the focus indication is very feint.

### Expected behavior:

To help users who are forced to use keyboard instead of a mouse or other pointing device - perhaps because of a motor impairment - a more obvious focus indication should be provided - similar perhaps to the links/buttons that are added to Hero block. 

### Screenshots:

image
This screenshot shows two accordions. The second one has keyboard focus.

image
This screenshot shows better focus indication on links in Hero block.
### Isolating the problem:

- [x] This bug happens with a default WordPress theme active - it's Twenty Twenty-One

### WordPress version:

WP 5.9.1

### Gutenberg version:

Not using the Gutenberg plugin.

snovosel-godaddy:
### Description

adjusts the outline color of the summary tag within the Accordion block for better visibility to keyboard users.

fixes #2303 

### Screenshots

Before:
Screen Shot 2022-05-03 at 11 58 39 AM

After:
Screen Shot 2022-05-03 at 11 57 41 AM

### Types of changes

bug fix / css change

### How has this been tested?

visually

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/accordion/accordion-item/styles/style.scss
+++ b/src/blocks/accordion/accordion-item/styles/style.scss
@@ -36,7 +36,7 @@
 		}
 
 		&:focus {
-			outline: 1px dotted #a2aab2;
+			outline: 1px dotted;
 			outline-offset: -4px;
 		}
 	}


EvanHerman (approved):
This looks good to me. Even with a black background the outline is still visible. Good work 👍 
image
image

PR #2371, opened 2022-05-03
Migrate dynamic HR to core/separator

EvanHerman:
### Description
Migrate dynamic HR block to `core/separator`.

### Types of changes

### How has this been tested?
Manually tested each block and attribute to ensure the blocks migrated without any breaking changes. The new core/separator block height differed from the dynamic HR block, so there are `core/spacer` blocks added to try and mimic the same heights. There is a slight difference in spacing due to block margins in the editor.

The tests were removed, since the dynamic HR block will no longer be present.

### Checklist:
- [x] My code is tested


jrtashjian (approved):
Looks good! I restored the front-end styles. We need to keep those around so that existing blocks continue to look the same until they are migrated.

PR #2394, opened 2022-06-06
WPEX 1973  - migrate features to core

snovosel-godaddy:
### Description

Migrate Features block to a pattern that implements core blocks and `coblocks/icon` block

### Types of changes

enhancement/refactor

### How has this been tested?

manually/visually

### Acceptance criteria

Block should be interchangeable with previous block versions and should be able to add new versions of the block without issue

### Checklist:
- [x] My code is tested


AnthonyLedesma (changes requested):
Looks like we have a similar situation here where the attributes for alignment and color are not being carried over. We may need to introduce block supports like we did with the Row block.
featuresMigration

AnthonyLedesma (commented):
This thing is so close right now. I set up a whole bunch of Features blocks with a lot of attributes set. Everything seems to be coming through except for color settings. I am pretty sure we managed to preserve these color settings in the Row block migration.
featuresMigration

PR #2404, opened 2022-07-05
WP 6.1 RC6 Status

olafleur-godaddy:
### Description
A draft PR to see the effect of having WP nightly, Gutenberg nightly and PHP 8.1 on CoBlocks


olafleur-godaddy:
Not needed anymore, as WP 6.1 was released.

PR #2406, opened 2022-07-06
WPEX 1978  migrate food and drinks to pattern

snovosel-godaddy:
### Description

Handle the migration for the Food and Drinks block to core blocks

### Types of changes

New feature / breaking change

### How has this been tested?

Manually

### Checklist:
- [x] My code is tested


jrtashjian (changes requested):
Initial reaction when loading this up is that this seems to be a bit excessive and I believe causing slowdowns when working in the editor. Could we try to simplify the pattern a bit?

image

I'm also getting some oddities in the migration from what was designed before the migration. Please make sure to test the barebones styles are mirrored as close as we can with core blocks.

Here are some from our template gallery:
- https://wpnux.godaddy.com/v3?page_id=38&template=lucy&language=en_US&country=US
- https://wpnux.godaddy.com/v3?page_id=11&template=lucca&language=en_US&country=US
- https://wpnux.godaddy.com/v3?page_id=21&template=barista&language=en_US&country=US

The old theme styles are also missing and we do need to keep these around if blocks haven't been migrated yet.

image

src/styles/style.scss
```diff
@@ -47,8 +47,6 @@
 @import "../blocks/faq/faq-item/styles/style.scss";
 @import "../blocks/faq/styles/style.scss";
 @import "../blocks/features/styles/style.scss";
-@import "../blocks/food-and-drinks/food-item/styles/style.scss";
-@import "../blocks/food-and-drinks/styles/style.scss";
```
jrtashjian, line 51: These front-end styles need to be restored to maintain the styles of the original block. The editor styles are the only ones not needed anymore.

jrtashjian (changes requested):

PR #2408, opened 2022-07-14
Drop support for PHP 5.6

olafleur-godaddy:
### Description
Drop support for PHP 5.6 as it not officially supported anymore.


eherman-godaddy (approved):
Looks good! 👍

PR #2414, opened 2022-07-28
Introduce framework for e2e testing migrations

AnthonyLedesma:
### Description

Introduced node scripts that will install post content from fixtures. This framework should run from the cypress config and prepare all fixtures for all tests before any tests run. To handle CLI path to WordPress I have introduced an env variable in CI.

To add new block migration tests using this framework you should do the following.
1. Introduce fixture files.
  - I would advise a number of variations. 
  - Ideal scenario would be to have tests on every feature for every block.
2. Introduce custom migration function using `setupPostsForE2E` helper utility and pass in block name.
  - Core helper function has extensive comments on usage - See: `setupPostsForE2E` comments.
  - Custom function should return an async function expression awaiting the helper return.
3. Edit cypress.config.js in the project root and await the custom function. 
4. Cypress spec file should verify quantity and name.

### Types of changes

Upgrade cypress to v10.
New utilities and structure for e2e testing migrations.

### How has this been tested?

CICD setup to run this new `coblocks/post` test.
Tests will function locally if you already have composer vendor files installed.

### Acceptance criteria

Test that migrations function as expected.
We should have fixture files in HTML format with several expected variations of block migrations. 
These should include different settings on the variations functionalities of said block. 
In this PR we test the Posts block has migrated to both 'core/query' and 'core/rss'.

### Checklist:
- [x] My code is tested


AnthonyLedesma:
Suffers from some existing issues with Lightbox apparently as well as some PHP linting needed.

src/utils/testing-utils.js
```diff
@@ -0,0 +1,295 @@
...
+ * const { name } = require('../block.json');
+ *
+ * const postsPrepare = () => ( async () => await setupPostsForE2E( name ) )();
+ *
+ * module.exports = {
+ *     postsPrepare
+ * }
+ * ```
+ *
+ * Within the cypress.config.js file we need to modify setupNodeEvents anonyomus function expression to include the block name helper.
```
olafleur-godaddy, line 29: anonymous 😉

src/blocks/posts/test/posts.cypress.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Include our constants
+ */
+import * as helpers from '../../../../.dev/tests/cypress/helpers';
+
+describe( 'Test CoBlocks Posts Block', function() {
+	/**
+	 * Test that we can add an Posts block that migrates into the core/query block or core/rss block respectively.
```
olafleur-godaddy, line 8: an Posts? a Posts? not sure here

olafleur-godaddy (approved):
Nice job! This is quite an undertaking! Yay for the update to Cypress 10! :) 
You can merge when the build is green!

PR #2429, opened 2022-09-28
Update FAQ block to default to h2 title

EvanHerman:
### Description
WPEX-2419

### Types of changes
Set the FAQ title to h2, and updated the pa11y.sql test file with the same changes.

### How has this been tested?
pa11y run locally.

### Acceptance criteria
- Switch the default FAQ block title element to h2 instead of h3
- Update pa11y.sql file with changes from above.

### Checklist:
- [x] My code is tested


jkopepasah-godaddy (approved):

AnthonyLedesma:
Would be cool to try and get this merged in. Will check out what the conflict looks like.

AnthonyLedesma:
I am not 100% clear on what should be removed from the conflict. That does not seem likely to affect deprecations given it's a default template but it is a possibility. Did we happen to confirm that no deprecations are required here?

kopepasah:
@AnthonyLedesma may be best to let @EvanHerman resolve the conflicts, as he is the original committer. Likely we will just want to take the most recent version of the SQL test file, but Evan will know best.

EvanHerman:
I think this is actually not relevant anymore after we merged in the other pa11y updates. I am pretty sure I made the necessary changes in https://github.com/godaddy-wordpress/coblocks/pull/2434

And our pa11y builds are still passing on master, so I think we can close out this PR.

PR #2430, opened 2022-09-30
Enable a11y testing on CircleCI.

jkopepasah-godaddy:
This PR covers fixes for Pa11y, which are each being committed in individual pull requests to this source branch. Once completed, we will merge this source branch into `master`, as long as the tests are passing.


kopepasah:
@EvanHerman please merge the master branch into this branch, resolving the conflicts, then let @olafleur know when it is ready.

kopepasah (approved):

EvanHerman:
@jkopepasah-godaddy The builds are failing here because the `perf_tests_current` job generates a `.json` file with the results using the branch name (`fix/pa11y-median-results.json`). In this case the branch name has a `/` in it, so it thinks that it is a path and it's saying the file doesn't exist, thus why the build consistently fails. Going to merge this but moving forward we can either not use slashes in branch names or update the build to convert slashes to dashes or something.

PR #2434, opened 2022-10-11
pa11y Updates

EvanHerman:
### Description
Update carousel to pass a11y tests.

### Types of changes
`a11y` tweaks (non-breaking change which fixes an issue)

# Notes:
- Carousel block needs deprecation to remove the ID attribute (to prevent duplicate carousel IDs on the page).

### How has this been tested?
pa11y.js

### Checklist:
- [x] My code is tested


src/js/coblocks-services.js
```diff
@@ -0,0 +1,15 @@
...
+		for ( let i = 0; i < servicesBlocks.length; i++ ) {
+			const currentServiceBlock = servicesBlocks[ i ];
+			const hasImage = currentServiceBlock.getElementsByTagName( 'figure' ).length;
+
+			if ( 0 === currentServiceBlock.innerText.length && ! hasImage ) {
+				currentServiceBlock.remove();
+			}
+		}
+	} );
+}() );
```
EvanHerman, line 15: The way that the services block is written, there is always an empty service block in the InnerBlocks.Content on save. As far as I can tell, there is no way around this without re-writing the services block to not inject a new service when the previous one contains content.

This script loads on the front of site and removes all empty service blocks on the page.

src/blocks/gallery-carousel/save.js
```diff
@@ -130,7 +130,7 @@ const save = ( props ) => {
 	return (
 		<div aria-label={ __( `Carousel Gallery`, 'coblocks' ) } >
 			<div className={ innerClasses }>
-				<div className={ swiperClasses } data-swiper={ JSON.stringify( swiperOptions ) } id={ uuid } style={ swiperStyles } >
```
kopepasah, line 133: Do we not need an ID on the carousel?

src/blocks/gallery-carousel/deprecated.js
```diff
@@ -1181,6 +1181,188 @@ const deprecated =
...
+												</figure>
+												<RichText.Content className={ captionClasses } tagName="figcaption" value={ image.caption } />
+											</div>
+										</div>
+									);
+								} ) }
+							</div>
+							{ prevNextButtons && (
+								<>
+									<button className={ `nav-button__prev` } id={ `${ uuid }-prev` } >
```
kopepasah, line 1335: If someone has two carousels on the page, the accessibility tests will still fail.

src/blocks/services/service/test/service.cypress.js
```diff
@@ -28,19 +28,53 @@ describe( 'Test CoBlocks Service Block', function() {
 	it( 'Test the service block custom classes.', function() {
 		helpers.addBlockToPost( 'coblocks/services', true );
 
-		helpers.addCustomBlockClass( 'my-custom-class', 'service' );
+		cy.get( '.wp-block-coblocks-services [data-type="coblocks/service"]:first-child' ).click( { force: true } );
+
+		helpers.upload.imageToBlock( 'coblocks/service' );
+
+		cy.get( '.block-editor-block-inspector__advanced' ).scrollIntoView().find( 'button' ).then( ( $btn ) => {
+			const isOpen = $btn.attr( 'aria-expanded' );
```
kopepasah, line 36: This code is reused three times, which is a good reason to add as a helper function.

includes/class-coblocks-block-assets.php
```diff
@@ -464,6 +495,18 @@ public function frontend_scripts() {
 			);
 		}
 
+		// Services block.
+		if ( $this->is_page_gutenberg() || has_block( 'coblocks/services' ) ) {
```
kopepasah, line 499: Rather than checking for the block then enqueueing this script, it needs to be added as a dependency of the Services block. So each time the services block renders, this script gets called.

src/js/coblocks-services.js
kopepasah: See my comment regarding adding this script as a dependency of the block.

kopepasah (changes requested):
Great work @EvanHerman! I added some feedback for a few items and looking forward to see what's next!

.dev/tests/a11y/pa11y.js
```diff
@@ -32,16 +32,21 @@ if ( process.argv[ 2 ] ) {
 // Set up the pa11y config options
 const config = {
 	standard: packageJson.testing.accessibility.compliance,
-	hideElements: '#wpadminbar',
+	hideElements: '#wpadminbar, #gist5332801 > div > div:nth-child(2)',
```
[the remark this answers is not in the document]
EvanHerman: This is the gist block navigation at the bottom. The embed does not pass a11y tests. Since the frontend markup is out of our control, I am just hiding it in the test.
image

.dev/tests/phpunit/includes/test-coblocks-form.php
```diff
@@ -319,7 +319,7 @@ public function test_render_field_email() {
 	 */
 	public function test_render_field_textarea() {
 
-		$this->expectOutputRegex( '/<textarea name="field-message-2\[value\]" id="message-2" class="coblocks-field coblocks-textarea" rows="20" ><\/textarea>/' );
+		$this->expectOutputRegex( '/<textarea name="field-message-2\[value\]" aria-label="Message" id="message-2" class="coblocks-field coblocks-textarea" rows="20" ><\/textarea>/' );
```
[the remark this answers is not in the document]
EvanHerman: This label as well as the Email aria-label are both translatable. The aria-label attribute matches the field label from the editor.

.dev/tests/phpunit/includes/test-coblocks-form.php
```diff
@@ -308,7 +308,7 @@ public function test_render_field_name_has_last_name() {
 	 */
 	public function test_render_field_email() {
 
-		$this->expectOutputRegex( '/<input type="email" id="email" name="field-email\[value\]" class="coblocks-field coblocks-field--email"  \/>/' );
+		$this->expectOutputRegex( '/<input type="email" id="email" aria-label="email" name="field-email\[value\]" class="coblocks-field coblocks-field--email"  \/>/' );
```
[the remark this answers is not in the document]
EvanHerman: See comment for "Message" aria-label below. (Although this field should be Email (`$label`) and not the field slug.

src/blocks/gallery-carousel/deprecated.js
EvanHerman: You are right. These ID's aren't used by use so I don't think they actually need to be present.

src/blocks/gallery-carousel/save.js
EvanHerman: No, the ID isn't used on our end. We initialize the carousels using the class name.

src/js/coblocks-services.js
jrtashjian: Although I'm not a fan of this we are migrating away from the services block so I'm ok with this existing until then since it resolves [WPEX-2415](https://godaddy-corp.atlassian.net/browse/WPEX-2415). It's not worth re-writing the block if we're removing it.

I would consider splitting this out into it's own PR so we can merge it independently if the 5 issues being resolved in this PR are not ready to merge.

includes/class-coblocks-block-assets.php
kopepasah: Okay, I better understand how these block scripts are being call. Not the most intuitive method, but it's clear you are following what already exists.

I was under the impression that block scripts were enqueued automatically, but I can see now it is a manual enqueue, in this method.

src/blocks/services/service/test/service.cypress.js
EvanHerman: Forgot about breaking this out into a helper. Doing that now.

kopepasah:
@EvanHerman once you merge in the latest changes from the default branch, these checks will pass. Approved.

kopepasah (approved):

PR #2435, opened 2022-10-13
Fix broken tests in Accordion block in WP 6.1 RC1

olafleur-godaddy:
### Description
Fix tests that failed because of settings panel changes that affected (at least) the Accordion block.

### DIFF
--- a/.dev/tests/cypress/helpers.js
+++ b/.dev/tests/cypress/helpers.js
@@ -363,7 +363,10 @@ export function setColorSettingsFoldableSetting( settingName, hexColor ) {
 	cy.get( '.block-editor-panel-color-gradient-settings__dropdown' ).contains( settingName, { matchCase: false } ).click();
 	cy.get( '.components-color-palette__custom-color' ).click();
 
-	cy.get( '[aria-label="Show detailed inputs"]' ).click();
+	if ( Cypress.$( '.branch-6-0' ).length > 0 ) {
+		cy.get( '[aria-label="Show detailed inputs"]' ).click();
+	}
+
 	cy.get( '.components-color-picker' ).find( '.components-input-control__input' ).click().clear().type( formattedHex );
 
 	cy.get( '.block-editor-panel-color-gradient-settings__dropdown' ).contains( settingName, { matchCase: false } ).click();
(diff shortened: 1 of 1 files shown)


.dev/tests/cypress/helpers.js
```diff
@@ -510,13 +546,7 @@ export function hexToRGB( hex ) {
...
-	return cy.get( containerClass + ' iframe' ).its( '0.contentDocument' ).should( 'exist' );
-}
-
-export function getIframeBody( containerClass ) {
-	return getIframeDocument( containerClass ).its( 'body' ).should( 'not.be.undefined' )
-		// wraps "body" DOM element to allow
-		// chaining more Cypress commands, like ".find(...)"
-		.then( cy.wrap );
+// Mailserver is usually setupped on CI
+export function MailServerSetupped() {
```
kopepasah, line 550: @olafleur while I wish Setupped was a work, sadly it is not. 😆

Let's use something like `MailServerIsSetup` or `MailServerIsReady`.

.dev/tests/cypress/helpers.js
```diff
@@ -510,13 +546,7 @@ export function hexToRGB( hex ) {
...
+	return Cypress.env( 'testURL' ) !== 'http://localhost:8889';
```
kopepasah, line 551: Are we use this helper function name needs to include MailServer? I am not seeing the relation between the output and Mail Server. Its really just checking the testURL, and nothing else. 

Maybe something like `IsNotWPTestEnv` or something like that would make more sense?

.dev/tests/cypress/helpers.js
kopepasah: Please also see my comment, regarding Mail Server.

kopepasah (changes requested):
Looks good, but left a few comments.

One other general question I have is regarding version support. I see you move some tests related to 5.9, and added a few conditionals specific to 6.1. 

1. Are we not supporting tests in 5.9?
2. What occurs when a release past 6.1 occurs to the conditionals?

.dev/tests/cypress/helpers.js
olafleur-godaddy: Renamed it to a more relevant name, good idea.

.dev/tests/cypress/helpers.js
olafleur-godaddy: Ha! Ha! Thanks, I was sure `setupped` existed ;)

olafleur-godaddy:
> 1. Are we not supporting tests in 5.9?

We do, until this PR is merged ;)
When we enter a transition of version, our focus is switching ( in that case, 5.9-6.0 to 6.0-6.1). On the CI, we only run on the current version (6.0), so it won't break anything. Our goal is to have a seamless transition on the CI from 6.0 to 6.1 on November 1st.

> 2. What occurs when a release past 6.1 occurs to the conditionals?

Great question. I see these conditionals as temporary scaffolding to be removed when we will start working on compatibility with WP 6.2 (because at this point, we will remove the 6.0 branches from the tests). But yeah, good point, I changed it so that it also works for any 6.X version of WordPress, so, it gives us a little bit more time than only during 6.1 :)

kopepasah (approved):
👍🏻

PR #2441, opened 2022-10-30
Make sure that the Welcome Guide does not show in the Cypress tests

olafleur-godaddy:
### Description
A few tweaks (with a lot of trial and error ;) ) to make the Cypress tests pass with WP 6.1 by fixing the last problem we have for the moment : the welcome guide showing up unexpected and making our tests fail :)

### DIFF
--- a/.dev/tests/cypress/helpers.js
+++ b/.dev/tests/cypress/helpers.js
@@ -93,10 +93,8 @@ export function disableGutenbergFeatures() {
 		}
 
 		if ( safeWin.wp.data.select( 'core/edit-post' ).isFeatureActive( 'welcomeGuide' ) ) {
-			cy.log( 'Was activated - toggle it' );
 			safeWin.wp.data.dispatch( 'core/edit-post' ).toggleFeature( 'welcomeGuide' );
 		} else {
-			cy.log( 'Is not activated' );
 		}
 
 		safeWin.wp.data.dispatch( 'core/editor' ).disablePublishSidebar();
--- a/.dev/tests/cypress/helpers.js
+++ b/.dev/tests/cypress/helpers.js
@@ -35,7 +35,7 @@ export function addFormChild( name ) {
  * Login to our test WordPress site
  */
 export function loginToSite() {
-	goTo( '/wp-admin/post-new.php?post_type=post' )
+	return goTo( '/wp-admin/post-new.php?post_type=post' )
 		.then( ( window ) => {
 			if ( window.location.pathname === '/wp-login.php' ) {
 			// WordPress has a wp_attempt_focus() function that fires 200ms after the wp-login.php page loads.
@@ -46,9 +46,8 @@ export function loginToSite() {
 				cy.get( '#user_pass' ).type( Cypress.env( 'wpPassword' ) );
 				cy.get( '#wp-submit' ).click();
 			}
-		} );
-
-	cy.get( '.block-editor-page' ).should( 'exist' );
+		} )
+		.get( '.wp-block-post-title' ).should( 'exist' );
 }
 
 /**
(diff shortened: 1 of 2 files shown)


AnthonyLedesma (commented):
Do we want to keep the logging in place here beyond local testing purposes?

.dev/tests/cypress/helpers.js
```diff
@@ -87,17 +86,19 @@ export function getWindowObject() {
...
 			safeWin.wp.data.dispatch( 'core/edit-post' ).toggleFeature( 'fixedToolbar' );
 		}
 
-		if ( ! safeWin.wp.data.select( 'core/edit-post' ).isFeatureActive( 'welcomeGuide' ) ) {
-			return;
+		if ( safeWin.wp.data.select( 'core/edit-post' ).isFeatureActive( 'welcomeGuide' ) ) {
+			cy.log( 'Was activated - toggle it' );
+			safeWin.wp.data.dispatch( 'core/edit-post' ).toggleFeature( 'welcomeGuide' );
+		} else {
+			cy.log( 'Is not activated' );
```
AnthonyLedesma, line 99: This is the logging I am referring to.

.dev/tests/cypress/support/commands.js
```diff
@@ -1,8 +1,10 @@
 import { disableGutenbergFeatures, loginToSite } from '../helpers';
 
 before( function() {
-	loginToSite();
-	disableGutenbergFeatures();
+	loginToSite().then( () => {
+		cy.wait( 10000 );
```
AnthonyLedesma, line 5: I do not love to see an arbitrary wait here. Do you know what we are waiting for exactly here? Perhaps we can instead assert something expected that only shows after this action is complete?

.dev/tests/cypress/helpers.js
olafleur-godaddy: Yeah, will remove it.

.dev/tests/cypress/support/commands.js
olafleur-godaddy: I also do not love it ;) , but I've tried to wait for a lot of things, and there is always an error somehow. Never figured out the right thing to wait on. So yeah, I know it's suboptimal, but I would merge it for now (as release of WP 6.1 is today) and try to find a better way to do it after release.

AnthonyLedesma: Sounds good to me. Should we add a quick comment there perhaps so we can easily spot it? Otherwise, we can make an effort in the future to investigate all uses of `wait` and fix it then.

olafleur-godaddy: Yeah, will add a comment, good idea.

.dev/tests/cypress/helpers.js
```diff
@@ -46,9 +46,8 @@ export function loginToSite() {
 				cy.get( '#user_pass' ).type( Cypress.env( 'wpPassword' ) );
 				cy.get( '#wp-submit' ).click();
 			}
-		} );
-
-	cy.get( '.block-editor-page' ).should( 'exist' );
+		} )
+		.get( '.wp-block-post-title' ).should( 'exist' );
```
jrtashjian, line 50: Smart. I'm assuming this will make Cypress wait until the editor is actually loaded rather than the DOM surrounding the React app.

olafleur-godaddy: Yes, it makes it wait a little longer :)

.dev/tests/cypress/support/commands.js
jrtashjian: This is so hard to do and we've gone through many solutions that failed in the past. Cypress has some weird ways to handle different paths like this but again, it's inconsistent.

I'm not sure where we're at with moving towards Storybook so it might not be worth going any further than the arbitrary wait. The change we can make knowing a wait is in our tests now is to make sure we're not running the `loginToSite` code on each test like we did in the past. There's an old PR I closed recently where I was [optimizing our tests](https://github.com/godaddy-wordpress/coblocks/pull/1319) for this reason.

jrtashjian (approved):
This looks good to me! You can merge it when you're ready.

AnthonyLedesma (approved):

.dev/tests/cypress/support/commands.js
olafleur-godaddy: Yeah, the fact that now we are heavily parallelizing the tests is that it is adding basically 10s to the overall build time instead of 10s for each test. So, barely noticeable ;)

PR #2442, opened 2022-11-01
Relocate styles used for Site Design controls.

AnthonyLedesma:
### Description

This PR is a measure to resolve issues with cascading styles. 
Companion PR on Go theme for style consistency.
https://github.com/godaddy-wordpress/go/pull/814

The summary of the overall issue is this: 
Styles on-load are persisted by PHP into the header element where we generate a custom style tag. After a user interacts with the Site Design component we previously modified this custom style element which resides in the header. An issue with this approach is that other styles used for the Site Design component live in the body and thus have higher specificity.

To fix this issue we use the same `useEffect` call that modified our custom style element to relocate the element into the proper hierarchal order thus allowing styles to cascade as expected. 

### Screenshots

designStylesMoved

### Types of changes

JavaScript. Also included with this story are some simple modifications to Go theme for style consistency.
Go PR: https://github.com/godaddy-wordpress/go/pull/814

### How has this been tested?

Tested manually.

### Acceptance criteria

Should allow font labels to change transform style as expected.

### Checklist:
- [x] My code is tested

### DIFF
--- a/includes/class-coblocks-site-design.php
+++ b/includes/class-coblocks-site-design.php
@@ -83,29 +83,6 @@ class CoBlocks_Site_Design {
 
 		add_action( 'wp_ajax_site_design_update_design_style', array( $this, 'update_design_style' ) );
 		add_action( 'rest_api_init', array( $this, 'design_endpoint' ) );
-
-		/**
-		 * Add the shared styles to the editor
-		 */
-		add_action(
-			'admin_init',
-			function() {
-				$suffix = SCRIPT_DEBUG ? '' : '.min';
-				$rtl    = ! is_rtl() ? '' : '-rtl';
-				// Enqueue  shared editor styles.
-				add_editor_style(
-					"dist/css/style-editor{$rtl}{$suffix}.css"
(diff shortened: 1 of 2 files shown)


includes/class-coblocks-site-design.php
```diff
@@ -83,29 +83,6 @@ public function __construct() {
...
-			'admin_init',
-			function() {
-				$suffix = SCRIPT_DEBUG ? '' : '.min';
-				$rtl    = ! is_rtl() ? '' : '-rtl';
-				// Enqueue  shared editor styles.
-				add_editor_style(
-					"dist/css/style-editor{$rtl}{$suffix}.css"
-				);
-			}
-		);
```
AnthonyLedesma, line 100: We have duplicate styles! So this is superfluously loaded from here as well as from Go. Considering Site Design requires Go we can remove this functionality from CoBlocks. Also logically speaking loading this sheet from Go makes more sense anyway.

jlemay-godaddy (approved):

PR #2449, opened 2022-11-18
Update form button class attribute name

### LINKED ISSUE #2218: Form Block Contact - Button Style outlined as filled
Using latest of Wordpress (5.8.2), CoBlocks Plugin (2.19.1) and Go-Theme (1.5.1)

As i noticed today the Buttons Styles of the Form Block when using the Contact-Template are not correct. I selected Button-Style "outlined" but Button is displayed as "filled". Same for Event-Template.

The Single-Button Block works well. I tried it with the Theme "Tabor", too - same issue there.

EvanHerman:
Resolves #2218 

### Description
The form block submit button class attribute name changed from `` to ``, but was not updated in the .php file that renders the button, so the form button styles did not reflect selected style on the front of site. The button would always show as the default 'filled' style, and not have the proper classes assigned to it.

### Screenshots
Master:
image

PR:
image

### Types of changes

Updated `$atts['submitButtonClasses']` to `$atts['className']` when the submit button is rendered.

### How has this been tested?
Manually tested.

### Acceptance criteria
Originally reported in #2218 

### Checklist:
- [x] My code is tested

### DIFF
--- a/.dev/tests/phpunit/includes/test-coblocks-form.php
+++ b/.dev/tests/phpunit/includes/test-coblocks-form.php
@@ -590,7 +590,7 @@ class CoBlocks_Form_Tests extends WP_UnitTestCase {
 		$this->expectOutputRegex( '/<button type="submit" class="wp-block-button__link custom-button-class" style="background-color: #B4D455; color: #333333;">Submit<\/button>/' );
 
 		$atts = [
-			'submitButtonClasses'         => 'custom-button-class',
+			'className'                   => 'custom-button-class',
 			'customBackgroundButtonColor' => '#B4D455',
 			'customTextButtonColor'       => '#333333',
 		];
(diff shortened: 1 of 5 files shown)


aledesma-godaddy:
Great fix. I think this is pretty close. I noticed that some of our style selectors are not working properly here however. We might want to update these selectors shown in this image so that the background color is not superfluously applied. 
Screen Shot 2022-11-30 at 8 48 43 AM

This is specifically related to the selector: This selector forces the outline style to use a background making the text unreadable. 
```css
.wp-block-button:not(.is-style-outline) .wp-block-button__link:not(.has-background) {
    background-color: var(--go-button--color--background, var(--go--color--primary));
}
```

EvanHerman:
@AnthonyLedesma Good call. I'll get that updated and ready or another review 👍

PR #2451, opened 2022-11-24
GitHub Actions Migration

kopepasah:
This pull request handles the migration from CircleCI to GitHub Actions.


kopepasah:
@olafleur while we are still completing the deploy to wp.org portion of this task, please go ahead and start to review, as there are some big changes included.

.dev/tests/cypress/helpers.js
```diff
@@ -35,69 +35,55 @@ export function addFormChild( name ) {
...
-			throw new Error( 'Check the previous test, window property was invoked outside of Editor.' );
-		}
-
-		if ( ! win?.wp ) {
-			throw new Error( 'Window property was invoked within Editor but `win.wp` is not defined.' );
-		}
-
-		return win;
+export function getWPDataObject() {
+	return cy.window().its( 'wp.data' ).then( ( data ) => {
```
olafleur-godaddy, line 67: Nice! I did not know about `its`

.dev/tests/cypress/support/commands.js
```diff
@@ -1,14 +1,11 @@
...
 	loginToSite().then( () => {
-		if ( isNotWPLocalEnv() ) {
-			// Waiting to see if the Welcome Guide will show up. Could probably be improved, but
-			// for the moment, it seems hard to tie the wait to something else
-			cy.wait( 10000 );
-		}
-
-		disableGutenbergFeatures();
+		goTo( '/wp-admin/post-new.php?post_type=post' ).then( () => {
+			cy.wait( 2000 );
```
olafleur-godaddy, line 6: Is there a reason why you still execute it in local now? In my tests, the wait was not needed, so that is a 2 seconds which is useless in that case.

.github/workflows/deploy-create-artifact.yml
```diff
@@ -0,0 +1,62 @@
...
+
+# First build our plugin, then create and store the artifact.
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        # with:
```
olafleur-godaddy, line 24: Commented out code?

.github/workflows/test-e2e-cypress.yml
```diff
@@ -0,0 +1,145 @@
...
+        type: string
+        default: go
+      wpVersion:
+        required: false
+        type: string
+        default: ""
+      phpVersion:
+        required: false
+        type: string
+        default: "8.1"
```
olafleur-godaddy, line 21: Unfortunately, I think the default should still be 7.4 for the moment, in our specific case.

.github/workflows/test-e2e-cypress.yml
```diff
@@ -0,0 +1,145 @@
...
+
+      - name: Build plugin
+        run: |
+          yarn build
+
+      - name: Prepare Go
+        if: ${{ inputs.theme == 'go' }}
+        run: |
+          cd $(wp-env install-path)/go
+          mkdir -p coblocks/icons
```
olafleur-godaddy, line 119: This custom SVG icon was not Go-specific, but CoBlocks specific (the Icon block), so it should be executed for any theme.

cypress.config.js
```diff
@@ -2,7 +2,7 @@ const { defineConfig } = require( 'cypress' );
 
 module.exports = defineConfig( {
 	chromeWebSecurity: false,
-	defaultCommandTimeout: 120000,
+	defaultCommandTimeout: 20000,
```
olafleur-godaddy, line 5: Oh, it's nice if we can reduce that time and not have flaky tests :)

olafleur-godaddy (approved):

olafleur-godaddy:
So, if I understand correctly, we don't test for variations of themes or browsers for the moment, right?

.dev/tests/cypress/support/commands.js
kopepasah: Yeah, I could not get this working in GitHub Actions consistently with Cypress. Seems the code executes too fast, even though it is waiting on the data object. We would possibly add some additional check for the featureActive portion, but will suffice until we can complete the Playwright migration.

kopepasah:
> So, if I understand correctly, we don't test for variations of themes or browsers for the moment, right?

Correct. We can add variations for themes, pretty quickly. I'll have to spend some time debugging why Firefox is not working in GitHub Actions, but let's tackle that in an alternate task.

.github/workflows/deploy-create-artifact.yml
kopepasah: It was commented out, until we are ready to run the release. Fixed now.

.github/workflows/test-e2e-cypress.yml
kopepasah: I added a test for PHP7.4, but did not change the default.

.github/workflows/test-e2e-cypress.yml
kopepasah: We may need to circle back to this, then. As we will need to find an alternate way to use WP-ENV and add these into any theme within tests.

.dev/tests/cypress/support/commands.js
olafleur-godaddy: Yeah, no problem on the waiting on the CI (we dit 10s before, so 2s is an improvement). I was more wondering why you removed the if that made the tests not wait in local (as it is not needed).

.github/workflows/test-e2e-cypress.yml
olafleur-godaddy: Ok, but just to give a little context, we added this last year because we realized that when we added a custom icon (like specified here : https://github.com/godaddy-wordpress/coblocks/blob/master/docs/hooks/icons-block.md ) it would break the Icon block. So, the idea is just to make sure that we don't break that functionality again in CoBlocks.

kopepasah: Okay, I've got this covered in the latest change: https://github.com/godaddy-wordpress/coblocks/pull/2451/commits/1bed6e1816396589a47363b11ca6fda3c45b5539

PR #2454, opened 2022-12-02
WPEX-2211 - add required filter for padding controls

snovosel-godaddy:
### Description

This PR adds a missing filter to the padding-controls. Without this filter the padding controls will not persist to the front end during the save post method.

### Checklist:
- [ ] My code is tested


AnthonyLedesma (changes requested):
The fix does not appear to be so simple.
paddingControlsFront-End

In this PR I see the core controls working as long as the CoBlocks controls have not been used. I also see front-end reflect correctly with the core controls but not with the CoBlocks controls.

PR #2467, opened 2023-01-17
Add a null check in class-coblocks-generated-styles.php

EvanHerman:
### Description
Originally reported in https://wordpress.org/support/topic/small-mistake-in-class-coblocks-generated-styles-php/

In WordPress 6.1 and PHP 8.1, there is a PHP warning when a null value is passed to `wp_add_inline_style`.

### Types of changes

### How has this been tested?
Manually tested.

### Checklist:
- [x] My code is tested


AnthonyLedesma (approved):
styles() depends on $post global. That is known to be `null` unless inside of a post. This is the correct logic to prevent this issue when the value is null.

PR #2472, opened 2023-01-25
Start Enzyme Migration to React Testing Library

olafleur-godaddy:
### Description
- Install React Testing Library dependencies
- Convert a first test file from Enzyme to RTL


kopepasah:
This looks good, but my only consideration (with may be due to legacy) is that we are using different coding practices for each test. One uses only `beforeEach()` to set the `wrapper`, while the other is using a `setup()` function. While I know it can be challenging in practice, we do need to solidify a set standard for writing these tests, so others can understand a test more easily.

However, I could also be misunderstanding how the `setup()` function works in the testing library.

Otherwise, I think this is good to go.

kopepasah (approved):

olafleur-godaddy:
Your comment is noted. Yes, agreed, we should aim to make a passthrough after to cleanup the way we initialize the objects we are testing.

PR #2499, opened 2023-02-24
Build the plugin and extract from correct dir.

AnthonyLedesma:
### Description

The deploy to WordPress workflow is now fixed, and can be run manually at any time by heading to https://github.com/godaddy-wordpress/coblocks/actions/workflows/deploy-wordpress.yml and clicking 'Run Worklow' in the top right. You will just need to enter the version being deployed (should match the plugin header in `master`).

A second workflow was added that will run whenever a tag is committed to GitHub. The tag is then built and deployed to WordPress.org SVN tags. The tag _will not_ be copied to trunk. The only time trunk/tags are both updated is when the `Deploy / WordPress` workflow mentioned above is manually triggered.

### Screenshots
image

### Types of changes

### How has this been tested?
Ran the workflow on GitHub actions.

### Acceptance criteria
- Resolve failing deploy workflow.
- Introduce a new workflow that deploys the tag to the appropriate WordPress.org tags directory.

### Checklist:
- [x] My code is tested


EvanHerman:
I believe @aledesma-godaddy is fixing the e2e errors in https://github.com/godaddy-wordpress/coblocks/pull/2501

.distinclude
```diff
@@ -1,17 +1,21 @@
 # List all files and directories to be deployed with the Go thene.
```
olafleur-godaddy, line 1: Is it a copy-paste error of talking about Go or it's talking about the Go install that CoBlocks is using?
Also, it should say `theme`

olafleur-godaddy (approved):
Looking good!

olafleur-godaddy:
> I believe @aledesma-godaddy is fixing the e2e errors in #2501

Ok, great! The quicker we can get to green, the better ;)

.distinclude
EvanHerman: Yeah, you are right. This should just be "deployed with CoBlocks". I've updated that, good catch 👍

AnthonyLedesma (commented):
I created the PR so I can not put approval on there but this looks good. Glad to see tests passing. In what ways did you test this? I assume you may have set up a tag deployment. Did you happen to try any other part of the deployment?
@EvanHerman

EvanHerman:
@AnthonyLedesma I tested it by running steps except the actual svn commit. When I was happy with where it was I re-added that last command in there

PR #2509, opened 2023-03-22
Add a check to see if a coblocks-animate is added to an element on the page

EvanHerman:
### Description
Determine if an element on the page has the `.coblocks-animate` class, and if so load the styles where the animations are defined.

Background: Adding an animation to a core element, when it is the only element on a page, causes the animation styles to not load.

To reproduce, you can just add a single element to the page (I tested and confirmed when using the `preformatted` block.) and then add an animation to the block. This works fine in the editor, but publish/update the page and look at it on the front of site and you will see the animation does not work.

### Types of changes

### How has this been tested?
Manually tested these changes.

### Checklist:
- [x] My code is tested

### DIFF
--- a/includes/class-coblocks-block-assets.php
+++ b/includes/class-coblocks-block-assets.php
@@ -143,7 +143,7 @@ class CoBlocks_Block_Assets {
 			}
 		}
 
-		if ( ! $has_coblock && ! $this->is_page_gutenberg() ) {
+		if ( ! $has_coblock && ! $this->is_page_gutenberg() && ! $this->has_coblocks_animation() ) {
 			return;
 		}
 
@@ -654,6 +654,15 @@ class CoBlocks_Block_Assets {
 			}
 		}
 		return false;
+	}
+
+	/**
+	 * Determine if the page content contains an element with a coblocks-animate class.
+	 *
+	 * @return boolean True when an element on the page has .coblocks-animate class, else false.
+	 */
+	public function has_coblocks_animation() {
+		return false !== strpos( get_the_content(), 'coblocks-animate' );
 	}
 }
 


includes/class-coblocks-block-assets.php
```diff
@@ -655,6 +655,15 @@ protected function is_page_gutenberg() {
...
 	}
+
+	/**
+	 * Determine if the page content contains an element with a coblocks-animate class.
+	 *
+	 * @return boolean True when an element on the page has .coblocks-animate class, else false.
+	 */
+	public function has_coblocks_animation() {
+		return false !== strpos( get_the_content(), 'coblocks-animate' );
+	}
```
AnthonyLedesma, line 666: Do you think we should use similar logic here for the content as your Timber fix? I guess it would only affect those on the custom page types we have discussed before.

EvanHerman: Good call. Done in the latest commit.

includes/class-coblocks-block-assets.php
```diff
@@ -655,6 +655,24 @@ protected function is_page_gutenberg() {
...
+	 */
+	public function has_coblocks_animation() {
+		global $post;
+
+		/**
+		 * Resolves a fatal error bug on PHP 8+ with Timber.
+		 *
+		 * @see https://wordpress.org/support/topic/the-method-has_masonry_v1_block-produces-a-fatal-error-on-php-8-0-22-and-above/
+		 */
+		$post_content = ! empty( $post ) ? $post->post_content : get_the_content();
```
jrtashjian, line 672: We need to make sure `the_content` filters are applied so that we are working with the output of any server-rendered blocks that have the `coblocks-animate` class applied.

```
$post_content = apply_filters( 'the_content', $post_content );
```

jrtashjian (changes requested):

jrtashjian (approved):

PR #2510, opened 2023-03-24
Update CoBlocks_Site_Design:short_circuit_check to prevent fatal error on WooCommerce system status page on PHP8

EvanHerman:
### Description
As reported, when on PHP 8, and having the WooCommerce + WooCommerce Subscriptions plugins installed, the WooCommerce system status fatal errors and is inaccessible.

### Screenshots
image

### Types of changes

### How has this been tested?
This was manually tested against PHP 7.4, and PHP 8. I tested with a few variations of plugins installed/activated and was consistently able to reproduce the issue with only PHP 8, CoBlocks, WooCommerce and WooCommerce Subscriptions.

### Acceptance criteria
Ensure that the WooCommerce system status is accessible and no fatal error exists.

### Checklist:
- [x] My code is tested

### DIFF
--- a/includes/class-coblocks-site-design.php
+++ b/includes/class-coblocks-site-design.php
@@ -99,8 +99,8 @@ class CoBlocks_Site_Design {
 	 * @return boolean
 	 */
 	public static function short_circuit_check() {
-		$active_theme = wp_get_theme( 'go' );
-		return 'Go' !== $active_theme->get( 'Name' );
+		$active_theme = get_stylesheet();
+		return 'go' !== $active_theme;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/class-coblocks-site-design.php
```diff
@@ -99,8 +99,7 @@ function() {
 	 * @return boolean
 	 */
 	public static function short_circuit_check() {
-		$active_theme = wp_get_theme();
-		return 'Go' !== $active_theme->get( 'Name' );
+		return 'go' !== get_option( 'stylesheet' );
```
sc0ttkclark, line 102: We might want to use `get_stylesheet()` here instead

EvanHerman: @sc0ttkclark Good call

sc0ttkclark (approved):

PR #2519, opened 2023-03-31
WPEX-2801 - fix gallery carousel in the site editor

snovosel-godaddy:
### Description

Previously, within the site-editor, the Gallery carousel was not working at all. This was happening because the TinySwiper package was using a DOM query that was intended for the Gutenberg DOM structure. This PR adds an additional layer of logic to determine if the block is being used within the Gutenberg editor, or specifically the site-editor. 

### Types of changes

Bug Fix

### How has this been tested?

Manually

### Acceptance criteria

The block should be workable in both the Gutenberg editor, and site-editor. Including both the editors and the respective front ends.

### Checklist:
- [x] My code is tested


src/components/swiper/index.js
```diff
@@ -56,6 +56,21 @@ const Swiper = ( props ) => {
...
+	const determineDocumentObject = () => {
+		const siteEditorIframeContainer = document.querySelector( '.edit-site-visual-editor__editor-canvas' );
+
+		let currentDocument = document;
+
+		if ( siteEditorIframeContainer ) {
+			currentDocument = siteEditorIframeContainer.contentDocument;
+		}
+
+		return currentDocument;
```
AnthonyLedesma, line 71: We should remove this variable and instead return the correct value. On line 68 we should return the `siteEditorIframeContainer.contentDocument` if truthy as expected here and on line 71 we should return `document`. This will avoid the extra variable and also it allows the variable on line 98 to be more contextually relevant.

src/components/swiper/index.js
```diff
@@ -80,10 +95,12 @@ const Swiper = ( props ) => {
 
 	useEffect( () => {
 		try {
-			const swiperContainer = document.getElementById( uuid );
+			const currentDocument = determineDocumentObject();
```
AnthonyLedesma, line 98: This is fine to have but recommend we simplify the logic around `determineDocumentObject`.

AnthonyLedesma (changes requested):
I have verified this fixes the original bug report. Just a few minor code change recommendations.

AnthonyLedesma (approved):
We can merge this. 

I wanted to find a way to programmatically determine if block is used on a template for 'site editing' but there does not yet seem to be a reliable way to determine this. We may have to load superfluous styles until we have a programmatic method to determine this.

PR #2521, opened 2023-04-06
Register all blocks in PHP

AnthonyLedesma:
### Description

The many changes are a first pass to convert the blocks to register in PHP. The reason for this change has several benefits. 
**By registering in PHP**
- We can finally do the work to split our block code. That will lead to significant performance benefits. 
- This allows us to load several assets on-demand only when the related blocks are rendered. For example the `lightbox, services-script, counter-script, gist-script, events-script, post-carousel-script, tiny-swiper-sciprs, masonry and animation scripts`. While a number of these scripts had attempted to render 'per-block', in many cases the whole script would render regardless. This change aligns us with core practice in that regard.
- It allows us to properly parse block attributes in various contexts. Previously, the server (PHP) had no knowledge of the block attributes which can affect block parsing in a number of ways including missing attributes.

### Checklist:
- [ ] My code is tested

### DIFF
--- a/includes/class-coblocks-register-blocks.php
+++ b/includes/class-coblocks-register-blocks.php
@@ -109,7 +109,7 @@ class CoBlocks_Register_Blocks {
 	}
 
 	/**
-	 * Load block manifest from block-manifest.json file.
+	 * Build a block manifest. The resulting array is used to register blocks and parses block.json files.
 	 * For now we need to collect the `render` property from the block.json files and require manually.
 	 *
 	 * @return array The block manifest or an empty array if no content is found.


.dev/tests/jest/helpers.js
```diff
@@ -45,42 +46,45 @@ import { createBlock, getBlockTransforms, parse, registerBlockType, serialize, u
...
-	const v2Masonry = stackedMeta?.apiVersion === 2 ? getV2Settings( masonryMeta, masonrySettings ) : {};
-	const v2Offset = stackedMeta?.apiVersion === 2 ? getV2Settings( offsetMeta, offsetSettings ) : {};
-	const v2Stacked = stackedMeta?.apiVersion === 2 ? getV2Settings( stackedMeta, stackedSettings ) : {};
-
-	registerBlockType( carouselName, { category: 'common', ...carouselSettings, ...v2Carousel } ); // Register carousel block
-	registerBlockType( collageName, { category: 'common', ...collageSettings, ...v2Collage } ); // Register collage block
-	registerBlockType( masonryName, { category: 'common', ...masonrySettings, ...v2Masonry } ); // Register masonry block
-	registerBlockType( offsetName, { category: 'common', ...offsetSettings, ...v2Offset } ); // Register offset block
-	registerBlockType( stackedName, { category: 'common', ...stackedSettings, ...v2Stacked } ); // Register stacked block
+	// const getV2Settings = ( blockMeta, blockSettings ) => {
```
olafleur-godaddy, line 49: If you don't need that code anymore, you can remove it ;)

includes/class-coblocks-block-assets.php
```diff
@@ -456,90 +445,139 @@ public function frontend_scripts() {
...
+	 *
+	 * @param string $block_content The block content to be rendered.
+	 * @param array  $block         The block attributes.
+	 *
+	 * @return string The original block content.
+	 */
+	public function coblocks_enqueue_scripts_for_core_blocks( $block_content, $block ) {
+
+		$block_name       = $block['blockName'];
+		$block_attributes = $block['attrs'];
```
olafleur-godaddy, line 519: I have a warning on my website : 
`Warning: Undefined array key "attrs" in /var/www/html/wp-content/plugins/coblocks/includes/class-coblocks-block-assets.php on line 519`

olafleur-godaddy (approved):
This is awesome work, good job!

includes/class-coblocks-block-assets.php
AnthonyLedesma: Good call! I've added some bail logic for if the block does not have attributes, or a name defined.

PR #2524, opened 2023-04-08
Disable Site Design controls in Go Site Editing.

AnthonyLedesma:
### Description

The Site Design controls were built with classic theme functionality in mind. The Site Editor controls now replicate the behavior of the Site Design controls inside of CoBlocks and we no longer need this tool with GSE. 

With this PR the controls will remain as backward compatible.

### Screenshots

Screen Shot 2023-04-07 at 5 38 45 PM

### Types of changes

PHP and JS mods to allow conditional disable for GSE context.

### How has this been tested?

Tested manually. Needs tests still.

### Acceptance criteria

Should disable when Go is updated to GSE. 
Should have a test in place to simulate conditions.

### Checklist:
- [ ] My code is tested


AnthonyLedesma:
Maybe we should explore hiding the controls altogether. Right now things are still disabled.  Having controls disabled here has the benefit of showing, even if a user is on another theme, it still points them to get GSE where they have baked-in site design controls.

AnthonyLedesma:
Even better if we can deep-link into the GSE design controls. @jrtashjian  Do you think it is reasonable to deep-link straight to the Site Editor in Go? See the image for details - we can have a link.

includes/class-coblocks-labs.php
```diff
@@ -54,7 +54,21 @@ public function __construct() {
...
 	 */
 	public static function is_go_active() {
-		return defined( 'GO_VERSION' ) && 'go' === get_option( 'stylesheet' );
+		return 'go' === get_option( 'stylesheet' );
+	}
+
+	/**
+	 * Checks if the Go theme's get_design_style() method is accessible, otherwise assumes FSE context.
+	 *
+	 * @since 3.1.0
```
EvanHerman, line 63: I think some of these versions are wrong. They should be `@since 2.0.0`

includes/class-coblocks-labs.php
```diff
@@ -115,6 +129,13 @@ public function propagate_settings() {
 			update_option( 'coblocks_site_design_controls_enabled', 0 );
 			update_option( 'coblocks_layout_selector_controls_enabled', 0 );
 		}
+
+		// Site design is disabled when in GSE context.
+		$is_go_se = self::is_go_se();
```
EvanHerman, line 134: Should this variable and function be `is_go_fse`? Or is `se` correct?

includes/class-coblocks-labs.php
```diff
@@ -115,6 +129,13 @@ public function propagate_settings() {
...
+
+		if ( $is_go_se ) {
```
EvanHerman, line 136: Nit picking, but I don't think that the variables are necessary here, since they are only used in this method. We can probably drop the two lines where the variables are assigned.

```php
if ( ! self::is_go_active() ) {
	update_option( 'coblocks_site_design_controls_enabled', 0 );
	update_option( 'coblocks_layout_selector_controls_enabled', 0 );
}

if ( self::is_go_se() ) {
	update_option( 'coblocks_site_design_controls_enabled', 0 );
}
```

includes/class-coblocks-labs.php
AnthonyLedesma: > Site Editing (formerly known as Full Site Editing or FSE)
https://wordpress.com/support/site-editing/

FSE is dead now and SE is the new norm. I have been pushing for GSE as a reference for that reason.

includes/class-coblocks-labs.php
```diff
@@ -54,7 +54,21 @@ public function __construct() {
...
+	 *
+	 * @since 3.1.0
+	 * @return boolean
+	 */
+	public static function is_go_se() {
+		if ( ! self::is_go_active() ) {
+			return false;
+		}
+		// No Method = FSE || Go classic.
+		return ! ( function_exists( '\Go\Core\get_design_style' ) );
```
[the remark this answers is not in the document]
AnthonyLedesma: The `GO_VERSION` constant will not be set in the new version, going by the current working GSE.
I am open to other suggestions regarding checking if it is a block theme.

PR #2525, opened 2023-04-10
WPEX-2801 - fix Carousel block on WP 6.2

snovosel-godaddy:
### Checklist:
- [ ] My code is tested

### DIFF
--- a/src/components/swiper/index.js
+++ b/src/components/swiper/index.js
@@ -72,7 +72,9 @@ const Swiper = ( props ) => {
 			swiperWrapperRef.current.addEventListener( ( 'mousedown' ), ( e ) => e.stopPropagation() );
 
 			return () => {
-				swiperWrapperRef.current.removeEventListener( ( 'mousedown' ), ( e ) => e.stopPropagation() );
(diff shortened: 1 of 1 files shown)


src/components/swiper/index.js
```diff
@@ -72,7 +72,9 @@ const Swiper = ( props ) => {
 			swiperWrapperRef.current.addEventListener( ( 'mousedown' ), ( e ) => e.stopPropagation() );
 
 			return () => {
-				swiperWrapperRef.current.removeEventListener( ( 'mousedown' ), ( e ) => e.stopPropagation() );
+				if ( swiperWrapperRef && swiperWrapperRef.current ) {
+					swiperWrapperRef.current.removeEventListener( ( 'mousedown' ), ( e ) => e.stopPropagation() );
+				}
```
AnthonyLedesma, line 77: What about using the optional chaining operator? We can do this instead:
```suggestion
swiperWrapperRef?.current?.removeEventListener( ( 'mousedown' ), ( e ) => e.stopPropagation() );
```
Where the `?.` operator will only execute if exists otherwise is equivalent to `null`. This is basically the same as checking the property first.

PR #2540, opened 2023-07-27
Conditionally set auto_detect_line_endings when lower than PHP 8.1

EvanHerman:
### Description
Conditionally set `auto_detect_line_endings` in the ical parser when on PHP < 8.1.

> Since PHP 8.1, PHP INI directive auto_detect_line_endings is deprecated. Setting it to true emits a PHP deprecation notice. This INI directive will be removed in PHP 9.0.

### Types of changes
- Bug fix (non-breaking change which fixes an issue)
- PHP 8 compatibility

### Checklist:
- [x] My code is tested


AnthonyLedesma (approved):

AnthonyLedesma:
Kinda, related here is that it seems like the Events block ical parser is completely broken now with PHP 7.4. I think we should disable the feature when running PHP 7.4 at this point. We cannot even download and run 7.4 due to security/support issues now. Is still around one some hosts but we should still work toward non-support.

AnthonyLedesma (changes requested):
Damn. It looks like the Events block ICS parser has been silently failing for a while. This is confirmed not to be a failure with 6.3. I get a visual error when running locally with `SCRIPT_DEBUG` and `WP_DEBUG` running PHP 8.0.28 with wp-env. 

Our E2E test is giving a false positive because it is checking for an error state, but the ICS parser giving a silent failure. It seems to properly parse the objects, however, that is not happening in reality.

Silent fail in prod, and apparently in some CI instances.
Screenshot 2023-08-08 at 5 33 56 AM

Fail seen in 7.4, and running 8.0 locally with wp-env.
Screenshot 2023-08-08 at 5 35 05 AM

```js
api-fetch.js?ver=c6922e5e289e31508e9e:727     GET http://localhost:8888/index.php?rest_route=%2Fwp%2Fv2%2Fblock-renderer%2Fcoblocks%2Fevents&context=edit&attributes%5BexternalCalendarUrl%5D=https%3A%2F%2Fcalendar.google.com%2Fcalendar%2Fical%2F8hohgb8qv19fgvjbbkcehe0ce0%2540group.calendar.google.com%2Fpublic%2Fbasic.ics&attributes%5BeventsRange%5D=all&attributes%5BeventsToShow%5D=5&attributes%5BshowCarousel%5D=true&post_id=4&_locale=user 500 (Internal Server Error)
```

AnthonyLedesma:
I realize now, that I forgot about this bug! What a bummer. Let's hold off on making any more changes to the events block for now I think.

PR #2575, opened 2023-10-23
[Urgent] Performance: don't subscribe all blocks to selected block

ellatrix:
### Description

Coblocks extends all blocks (through the `editor.BlockListBlock` filter) with a `withSelect` HoC, subscribing _all_ blocks to the currently selected block. This means that on typing, the selected block updates, and _all_ blocks on the page re-render because they are all subscribed to changes to the selected block.

I went through all functions in `addAllEditorProps` to which `props` is being passed to, but none of them seem to be using `props.selected`, so it looks like this subscription can be removed without further changes. 

### Types of changes

Use the React Dev Tools and enable highlight on re-render. Type in a block and watch all blocks re-render.

image

### Checklist:
- [ ] My code is tested

### DIFF
--- a/src/extensions/apply-extensions.js
+++ b/src/extensions/apply-extensions.js
@@ -84,7 +84,6 @@ const enhance = compose(
 		return {
 			getBlocks: select( 'core/block-editor' ).getBlocks,
 			select,
-			selected: select( 'core/block-editor' ).getSelectedBlock(),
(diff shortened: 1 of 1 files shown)


AnthonyLedesma (approved):
Excellent find. Thank you so much for looking into this issue!

PR #2607, opened 2024-04-25
Sanitize the iCal URL before initializing

EvanHerman:
### Description
- [x] Escape the iCal URL before initializing the events calendar.
- [x] Prevent `localhost` or `127.0x` URLs from being used for external calendar URLs.

### Types of changes

### How has this been tested?
Tested by running `nc -l 127.0.0.1 9000`. Adding a `Events` block to the page, ticking off `Link a calendar` and adding the URL `http://127.0.0.1:9000` to the calendar URL and ensuring an error is returned.

### Checklist:
- [x] My code is tested

### DIFF
--- a/src/blocks/events/index.php
+++ b/src/blocks/events/index.php
@@ -19,6 +19,11 @@ function coblocks_render_coblocks_events_block( $attributes, $content ) {
 		return $content;
 	}
 
+	// If externalCalendarUrl contains a localhost URL, return an error message.
+	if ( strpos( $attributes['externalCalendarUrl'], 'localhost' ) !== false || strpos( $attributes['externalCalendarUrl'], '127.0' ) !== false ) {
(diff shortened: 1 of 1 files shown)


src/blocks/events/index.php
```diff
@@ -34,7 +39,7 @@ function coblocks_render_coblocks_events_block( $attributes, $content ) {
 				'use_timezone_with_r_rules'     => false,
 			)
 		);
-		$ical->init_url( $attributes['externalCalendarUrl'] );
+		$ical->init_url( sanitize_url( $attributes['externalCalendarUrl'] ) );
```
olafleur-godaddy, line 42: The linting system thinks you should use `esc_url_raw()` instead, since `sanitize_url` is deprecated.

olafleur-godaddy: Oh yeah, you're right, it seems it is "un-deprecated". Our linting system is probably not up to date, then ;)

PR #2612, opened 2024-05-29
[WPEX-3724] Resolve XSS security vulnerability

mtashjianjr-godaddy:
Resolves an issue with improper escaping of HTML output in the Social Profiles block. The solution implemented (other than ensuring output is properly escaped) is to use `wp_kses` with a defined set of html elements and attributes expected which removes all extraneous attributes.


EvanHerman (approved):
Looks good to me. Works well on my end. Nice work! 👍

PR #2624, opened 2024-12-20
Improve nonce handling and permissions in crop settings API

mtashjianjr-godaddy:
Enhance nonce validation and permission checks for crop settings, ensuring only authorized users can access and modify images. Additionally, conditionally include nonce values in localized script data.

### DIFF
--- a/includes/admin/class-coblocks-crop-settings.php
+++ b/includes/admin/class-coblocks-crop-settings.php
@@ -86,26 +86,22 @@ class CoBlocks_Crop_Settings {
 	 * Retrieve the original image.
 	 */
 	public function get_original_image() {
-		$nonce = filter_input( INPUT_POST, 'nonce' );
-
-		if ( ! $nonce ) {
-
-			wp_send_json_error( 'No nonce value present.' );
-
+		if ( ! wp_verify_nonce( filter_input( INPUT_POST, 'nonce' ), 'cropSettingsOriginalImageNonce' ) ) {
+			wp_send_json_error( 'Invalid nonce value.', 403 );
 		}
 
-		if ( ! wp_verify_nonce( htmlspecialchars( $nonce ), 'cropSettingsOriginalImageNonce' ) ) {
-
-			wp_send_json_error( 'Invalid nonce value.' );
-
+		if ( ! current_user_can( 'upload_files' ) ) {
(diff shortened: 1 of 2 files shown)


includes/admin/class-coblocks-crop-settings.php
```diff
@@ -86,26 +86,22 @@ public function hide_cropped_from_library( $query ) {
 	 * Retrieve the original image.
 	 */
 	public function get_original_image() {
-		$nonce = filter_input( INPUT_POST, 'nonce' );
-
-		if ( ! $nonce ) {
-
-			wp_send_json_error( 'No nonce value present.' );
-
+		if ( ! wp_verify_nonce( filter_input( INPUT_POST, 'nonce' ), 'cropSettingsOriginalImageNonce' ) ) {
```
eherman-godaddy, line 89: I'm wondering if we should wrap this filter_input with a `sanitize_text_field()` just to ensure it's sanitized before checking. This has been getting flagged in plugin reviews, and the security team might flag it if we don't.

```php
if ( ! wp_verify_nonce( sanitize_text_field( filter_input( INPUT_POST, 'nonce' ) ), 'cropSettingsOriginalImageNonce' ) ) {
```

includes/admin/class-coblocks-crop-settings.php
```diff
@@ -127,18 +123,22 @@ public function get_original_image() {
 	 * Cropping.
 	 */
 	public function api_crop() {
-		$nonce = filter_input( INPUT_POST, 'nonce' );
-
-		if ( ! $nonce ) {
-
-			wp_send_json_error( 'No nonce value present.' );
+		if ( ! wp_verify_nonce( filter_input( INPUT_POST, 'nonce' ), 'cropSettingsNonce' ) ) {
```
eherman-godaddy, line 126: `sanitize_text_field` here as well

includes/admin/class-coblocks-crop-settings.php
mtashjianjr-godaddy: That's odd, considering it's not stored in a database or rendered on the website. We can add it regardless.

eherman-godaddy: Yeah, I agree. It throws a warning in our plugin scanner tool and in the plugin check plugin.

